PR #1, opened 2015-12-06
ENH: Add bin script to start application

davidastephens:
Attempt at adding a command line script to run the application.  

For some reason it starts fine, but I get an error when I make a request from the server.  I have no idea why, because it works fine when I use the same commands from a python shell.  I don't know flask very well, maybe you have an idea why this isn't working.

This is the error message I get:
[I 151206 13:05:45 server:271] Serving on http://localhost:5000
[I 151206 13:05:45 handlers:58] Start watching changes
[I 151206 13:05:45 handlers:60] Start detecting changes
[E 151206 13:05:52 wsgi:355] 500 GET /balance_sheet/ (127.0.0.1) 5141.36ms

### DIFF
--- a/scripts/beancount-web
+++ b/scripts/beancount-web
@@ -1,0 +1,24 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+
+import os
+import sys
+import argparse
+
+from beancount_web import application
+
+"""
+Run script for beancount-web.
+
+"""
+
+
+def main(arguments):
+    parser = argparse.ArgumentParser(description=__doc__,
+            formatter_class=argparse.RawDescriptionHelpFormatter)
+    parser.add_argument('filename', help = "Beancount input file", type = str)
+    args = parser.parse_args(arguments)
+    application.run(args.filename)
+
+if __name__ == '__main__':
+    sys.exit(main(sys.argv[1:]))
--- a/setup.py
+++ b/setup.py
@@ -8,6 +8,7 @@ setup(name='beancount-web',
       author_email='dominik@aumayr.name',
       license='MIT',
       packages=['beancount_web'],
+      scripts=['scripts/beancount-web'],
       install_requires=[
             'beancount',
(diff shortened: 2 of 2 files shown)


aumayr:
Thanks @davidastephens! Works, and was much needed :-)

davidastephens:
So it works for you? I have no idea what the issue is on my computer..  I'll try it in a new virtualenv.

aumayr:
Yes, works fine! In 87b1b5ef1db9bcc0cb255944dcfc7f559104d9cc I added some more arguments.

You could try the following: Grab the latest version (including 87b1b5ef1db9bcc0cb255944dcfc7f559104d9cc), and try running the script with `--debug`. This enables the built-in Flask webserver (instead of the one comming with `livereload`), which has a built-in web-enabled debugger where you can view stacktraces, open a debugging console  (in the browser!), etc. Maybe this gives you a hint where the exception (leading to the `HTTP 500`) comes from.

davidastephens:
Thanks I saw that and pulled down the new one.  Here's what I get:
```
File "/home/david/.virtualenvs/beancount/lib/python3.4/site-packages/Flask-0.10.1-py3.4.egg/flask/templating.py", line 64, in get_source
            try:
                return loader.get_source(environment, local_name)
            except TemplateNotFound:
                pass
 
        raise TemplateNotFound(template)
 
    def _iter_loaders(self, template):
        loader = self.app.jinja_loader
        if loader is not None:
            yield loader, template
jinja2.exceptions.TemplateNotFound: balance_sheet.html

```

Again, not a Flask expert - is there something you need to do to the application so it knows where to find the templates?

aumayr:
I created a new branch called `fix-template-lookup-exception` to track any changes for fixing this. 

A first attempt: 8a8c5aa0b55b485a5b575a4c0260788378ee0590 => The Flask/Jinja2 template loader does seem to be confused about where to look for templates.

PR #2, opened 2015-12-06
BUG: Fix template finding

davidastephens:
Figured out the problem.  It was just that setup.py wasn't copying the templates when installing.  You were probably doing python setup.py develop, which wouldn't have had this issue.

Solution as per: http://flask.pocoo.org/docs/0.10/patterns/distribute/

It doesn't need the commit you put in the Try to fix branch.


aumayr:
Ahh that was it, thanks for the PR!

PR #8, opened 2015-12-09
separate out argparse etc. into cli.py

yagebu:
Also, remove standalone script and use `entry_points` to generate the
executable script.

The flask app and the command line argument parsing don't really belong in the same module IMHO.


aumayr:
Works like a charm! Thanks!

PR #11, opened 2015-12-10
Api refactorings

yagebu:


aumayr:
Thanks @yagebu for the work! How do these changes affect the performance with your large files?

yagebu:
The `reload` function took ~1s before, so it's gotten a bit faster. I'll try to do some more profiling on the api once I get the time to see where things should (and hopefully can) be improved.

Probably the simplest way to make loading the reports with charts more snappy would be to only load the chart data after the page has loaded. The tree tables seem to be generated quite quickly, even for large datasets, but the splitting by month etc. to generate that chart data is a bit more involved..

aumayr:
Asynchronously loading the charts should be simple, will work on that. 

What do you think about pagination for Journal entries (and monthly Account balances)? Would that help performance?

yagebu:
When I use the journal I do quite appreciate that's it all on one page so I can use my browser to quickly search and jump around. Since it's such a long page, I actually don't mind the load time as much...

So I wouldn't consider pagination the highest priority right now.

PR #13, opened 2015-12-11
reduce redundancy in application.py

yagebu:
The templates for most of the reports already contain all the logic of
what parts of the api they need. So pass the api along to the template
context and compute whatever is needed there.

Also, compute api attributes that are independent of the filters applied as soon as the file is loaded. Less computations on page load...


aumayr:
Awesome, thanks @yagebu!

PR #15, opened 2015-12-12
Performance improvements for monthly reports, profiling support, small bugfix

jbms:


aumayr:
Thank you, @jbms! 

How big are the performance improvements you are seeing?

yagebu:
Superb work! For me, pages that took several seconds to load now load almost instantly. All but the journal now load without much delay (and there it's seems to be mostly the templating that takes a lot of time).

Issue #25, opened 2015-12-14
Change Year filter to allow for date ranges


aumayr:
Currently the filter only shows (and filters for) full years. Change that to enable arbitrary date ranges.

yagebu:
I have implemented this in a branch and can submit a PR.

aumayr:
@yagebu Perfect, thanks! I just pushed some changes to the filters, so be sure to check them out before you create a PR.

aumayr:
Implemented in #27

PR #27, opened 2015-12-14
Dateparser

yagebu:
Allow any date and dateranges to be parsed and filtered to by the api. Adds a `dateparser` module that does the parsing.

Currently there is no nice way to input these dates, as the input form allows only to select the suggested years. Just a text input where any text can be input but that still has the list of suggestions would be nice. @aumayr: Maybe it's quickest if you could do this, as I don't know the frontend stuff very well.

The format of dates that can be parsed (from the docstring of dateparser.parse_date):

    Tries to parse the given string into two date objects marking the
    beginning and the end of the given period.

    Example of supported formats:
     - today, tomorrow, yesterday
     - 2010-03-15, 2010-03, 2010
     - march 2010, mar 2010
     - this month, last year, next year
     - october this year, aug last year

    Ranges of dates can be expressed in the following forms:
     - start - end
     - start to end
    where start and end look like one of the above examples

ref #25: partially implements this (all but the frontend part)

### DIFF
--- a/beancount_web/api.py
+++ b/beancount_web/api.py
@@ -199,9 +199,12 @@ class BeancountReportAPI(object):
         self.active_years = list(getters.get_active_years(self.all_entries))
         self.active_tags = list(getters.get_all_tags(self.all_entries))
         self.active_payees = list(getters.get_all_payees(self.all_entries))
+        self.apply_filters()
 
+    def apply_filters(self):
         if self.filter_year:
             begin_date, end_date = parse_date(self.filter_year)
+            self.entries = self.all_entries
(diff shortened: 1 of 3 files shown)


aumayr:
Thanks, @yagebu! I will have a look at it and see how we can implement a nice search field.

aumayr:
Looks good! I added support in the frontend and added some examples. 

One problem, though: Some human-readable date-ranges produce errors, like `this month`. Can you reproduce the error with `this month` for example?

yagebu:
Wrote the dateparser some time ago for different project and it's never been very well tested. I found the bug and will write a fix later.

aumayr:
(see #29)

PR #29, opened 2015-12-14
Fixes

yagebu:


aumayr:
Works great, thanks! 

One question still remains open: How to deal with daterange-strings that can not be parsed successfully?

PR #44, opened 2015-12-15
Api refactorings

yagebu:
The first two commits are quite trivial.

The third one might warrant some discussion since you implemented it this way.

I think that there should never be gaps in the monthly totals charts. If my balance (or change) of an account in some month is zero, that's still something I want to know. Otherwise the time-axis is also screwed up.

Why I think that also the date range (from the first to the last transaction in the currently filtered entries) shown after this change is better: The time axis will be consistent across all my accounts  and if I want to reduce the range of dates to just the ones of some account, I can simply filter by the account.

### DIFF
--- a/beancount_web/api.py
+++ b/beancount_web/api.py
@@ -478,20 +478,12 @@ class BeancountReportAPI(object):
             income_totals = self._table_totals(realization.get(realized, self.account_types.income))
             expenses_totals = self._table_totals(realization.get(realized, self.account_types.expenses))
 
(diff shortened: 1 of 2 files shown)


aumayr:
Thanks, @yagebu, for the work! I am happy with all of it, and merged it into `master`.

For the gaps in the monthly totals charts: I have implemented this this way because I personally have long timespans where there are no transactions (just copied in my bank balances for the last couple of years), and started adding transactions just some months ago. Therefore I had long gaps with total 0 in these charts, and therefore thought I should cut it out. But your code is perfectly correct, so I will stick to that.

PR #53, opened 2015-12-20
ENH: Add yearly view

davidastephens:
Adds a yearly view to the journal page (same premise as the monthly view).

### DIFF
--- a/beancount_web/templates/charts/_chart_yearly_totals.html
+++ b/beancount_web/templates/charts/_chart_yearly_totals.html
@@ -1,0 +1,29 @@
+<script>
+    window.chartData.push({
+        label: "{{ label }}",
+        type:  "bar",
+        dateFormat: "YYYY", // #TODO: Why doesn't this work?
+        data: {
+            labels: [
+                {% for total in yearly_totals %}
+                new Date({{ total.end_date.year }}, {{ total.end_date.month - 1}}, {{ total.begin_date.day }}),
+                {% endfor %}
+            ],
+            series: [
+                {% for currency in operating_currencies %}
+                {
+                    name: "{{ currency }}",
+                    data: [
+                        {% for total in yearly_totals %}
+                            {
+                                y: {{ total.totals[currency] or 0 }},
+                                meta: "{{ total.totals[currency]|format_currency or 0.00 }} {{ currency }}"
+                            },
+                        {% endfor %}
+                    ],
+                },
+                {% endfor %}
+            ]
+        }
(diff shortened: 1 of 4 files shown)


yagebu:
Could I suggest that this is taken as an opportunity to refactor the `monthly*` functions to be agnostic of the given time periods? The only real difference is that `get_next_month` is replaced by `get_next_year`, the rest of the code seems to be mostly copy-paste, which seems like a bad idea for code maintenance. This would also make it easier to implement a weekly and quarterly views in the future (and whatever else people might find useful).

If that seems like to much effort to you, I would be fine if this is merged and then I would do the refactor at some point.

aumayr:
@yagebu I am on it right now before I merge this ;-) Will be finished in half an hour, and would be happy if you could take a look at it and improve if necessary.

yagebu:
@aumayr: Great! Just ping me on the PR and I'll have a look later today

aumayr:
Thanks @davidastephens for this PR. I refactored the code a litte bit to be more generic (as @yagebu pointed out), and merged it into master just now. 

For the `YYYY` that didn't work: I'll comment directly at that line what went wrong.

beancount_web/templates/charts/_chart_yearly_totals.html
```diff
@@ -0,0 +1,29 @@
+<script>
+    window.chartData.push({
+        label: "{{ label }}",
+        type:  "bar",
+        dateFormat: "YYYY", // #TODO: Why doesn't this work?
```
aumayr: @davidastephens Two reasons why this did not work properly:

1. `dateFormat` has to be nested into a dictionary called `options`, see https://github.com/aumayr/beancount-web/commit/be39f1ccd403ce5c898898975405a1fbe84f352b#diff-113bfa7a88d6e5731c07319321981c0fR5
2. The `YYYY`-format-string has to be implemented manually. I did choose to implement these format-strings myself, because libraries like `moment.js` are quite heavy and I need only very few of those. See https://github.com/aumayr/beancount-web/commit/be39f1ccd403ce5c898898975405a1fbe84f352b#diff-d7f3007e2206954d98f57e59551d9403R71

aumayr:
@yagebu `ping`

I already merged this into master, because I did (do) not know how to append commits to an existing PR from someone else :-/

yagebu:
Looks good!

davidastephens:
Great - yes, it did need a refactor.  Looks great.

PR #55, opened 2015-12-20
api: fix daterange filtering of entries

yagebu:
For the list of all entries, we want to have the summarize entries
generated, otherwise the balances are off.


aumayr:
👍

Issue #54, opened 2015-12-20
Rethink App state vs. Session state


yagebu:
Currently, the set of filters is saved as a global state on the `app` object. So it doesn't survive restarts of the server and makes it impractical for several users to access the same beancount-web instance. I recommend we change this to save it as a session state. I can think of two possible ways to do this:

1. "Clean URLs". Filters are sent as `POST` requests and the current filters are saved as a cookie. Flask allows us to do this easily. Quite similar to the way it is now, although I would remove the `/filter/` endpoint (except for the permalink part), as the redirect makes outputting error messages harder. (as I noticed when trying to port my dateparser exception code to master ;) )

2. Keep everything in the URL query string. Advantage would be that every URL is automatically a permalink. As we don't have to many filter options, I think it would be more transparent to use the query string to save the session state and would remove the need for a separate permalink url route.

@aumayr which option do you prefer? I have slight preference for 2., but both would be an improvement going forward.

aumayr:
I would prefer 1., because there may be more query strings in the future (think table-column-ordering, etc.) and managing every link on every page to always have the correct query strings sounds like a huge task. 

For the "transparency"-argument in 2.: The filter-permalink shows all the selected filters in a way 2. would do, so that kind of transparency would be given in 1. as well. Do you agree?

yagebu:
> I would prefer 1., because there may be more query strings in the future (think table-column-ordering, etc.) and managing every link on every page to always have the correct query strings sounds like a huge task.

It would only be feasible to implement 1. if we can patch `url_for` to automatically append the query string (and `app.url_defaults` looks like it does exactly that). I would reserve query strings for things like filters that massively change the data that is displayed. I think the `URL` should be a precise (as far as possible) identifier for the page that's displayed; and `/income_statement/` and `/income_statement/?time=2015` are two very different things, so they deserve unique URLs IMHO.

> For the "transparency"-argument in 2.: The filter-permalink shows all the selected filters in a way 2. would do, so that kind of transparency would be given in 1. as well. Do you agree?

Well on a website I personally couldn't be bothered with a separate link to set a bookmark. Since `beancount-web` is more of a web app, I don't mind as much; still my browser has a nice button to set a bookmark, so if there is no good reason not to have the URL be indentifying for that page, I think it's an unnecessary nuisance.

aumayr:
>  I think the `URL` should be a precise (as far as possible) identifier for the page that's displayed;

True, and a good reason for 2. 

[...]

PR #60, opened 2015-12-21
Save current set of filters in the query string.

yagebu:
As discussed in #54.

Currently the input fields don't work. Since the "form" doesn't have all the necessary input fields to just send it as a form (which might be an option though now), there needs to be some Javascript to handle it. @aumayr: could you have a look? You seem to be much better with Javscript.

### DIFF
--- a/beancount_web/templates/_journal_table.html
+++ b/beancount_web/templates/_journal_table.html
@@ -20,9 +20,9 @@
 {% endif %}
 
 <script>
-    window.contextURL = "{{ url_for('context', ehash='_REPLACE_ME_') }}";
-    window.accountURL = "{{ url_for('account_with_journal', name='_REPLACE_ME_') }}";
-    window.documentURL = "{{ url_for('document') }}?file_path=_REPLACE_ME_";
+    window.contextURL = "{{ url_for('context', ehash='REPLACEME') }}";
+    window.accountURL = "{{ url_for('account_with_journal', name='REPLACEME') }}";
+    window.documentURL = "{{ url_for('document') }}?file_path=REPLACEME";
     window.journalURL = "{{ url_for('journal') }}";
 
 {% if journal is defined %}
@@ -109,7 +109,7 @@
                         </div>
                         {{#if tags }}
                             <div class="tags">
-                                {{#each tags }}<a href="{% endraw %}{{ url_for('filter_entries') }}?next={{ request.path|urlencode }}{% raw %}&filter_type=tags&filter_value={{ this }}" title="Filter for tag #{{ this }}"><span>#</span>{{ this }}</a>{{/each}}
+                                {{#each tags }}<a href="{% endraw %}{{ url_for_current() }}{% raw %}&tag={{ this }}" title="Filter for tag #{{ this }}"><span>#</span>{{ this }}</a>{{/each}}
                             </div>
                         {{/if}}
                     {{/ifCond}}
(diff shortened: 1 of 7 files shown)


beancount_web/templates/_journal_table.html
```diff
@@ -109,7 +109,7 @@
                         </div>
                         {{#if tags }}
                             <div class="tags">
-                                {{#each tags }}<a href="{% endraw %}{{ url_for('filter_entries') }}?next={{ request.path|urlencode }}{% raw %}&filter_type=tags&filter_value={{ this }}" title="Filter for tag #{{ this }}"><span>#</span>{{ this }}</a>{{/each}}
+                                {{#each tags }}<a href="{% endraw %}{{ url_for_current() }}{% raw %}&tag={{ this }}" title="Filter for tag #{{ this }}"><span>#</span>{{ this }}</a>{{/each}}
```
aumayr: This leads to incorrect URLs when no other filters are set. How can we fix that?

beancount_web/templates/_layout.html
```diff
@@ -56,8 +56,8 @@ <h1 class="site-name">{{ title or 'Beancount Web' }}</h1>
             {% for title_, menuitems in navigation_bar %}
                 {% if title_ %}<h3>{{ title_ }}</h3>{% endif %}
                 <ul class="navigation">
-                {% for href, id, caption in menuitems %}
-                    <li{% if id == active_page %} class="selected"{% endif %}><a href="{{ href|e }}">{{ caption|e }}{% if id == 'errors' and errors|length > 0 %}<span>{{ errors|length }}</span>{% endif %}</a></li>
+                {% for id, caption in menuitems %}
+                    <li{% if id == active_page %} class="selected"{% endif %}><a href="{{ url_for('report', report_name=id) }}">{{ caption|e }}{% if id == 'errors' and errors|length > 0 %}<span>{{ errors|length }}</span>{% endif %}</a></li>
```
aumayr: Why does this generate the correct URL for the menu item `source`? :eyes:

aumayr:
Thanks @yagebu for the great work! I will have a look at the `input`-fields and then merge into `url-defaults`.

In the meantime: I have some questions about the PR, which I directly commented on the corresponding source code lines. Could you have a look, please?

aumayr:
I found some bugs:

- https://github.com/aumayr/beancount-web/pull/60/files#diff-588116796d0224ea12134898ca9a828cR112 => This leads to incorrect URLs when there are no filters (no `?` to start query part)
- When selecting a payee filter, and then selecting an account filter, the payee filter is gone. This does happen for other combinations as well (tags, and then time, etc.)

I will merge anyway, because we are on our own branch `url-defaults` anyway. Can you have a look at those bugs after I merged (second PR)?

beancount_web/templates/_layout.html
yagebu: The url route for report is generic enough that it would match any url of the form `/some_string/`. That's why the more specific url routes come first, but `url_for` doesn't know that and happily generates the url for `source` :D

beancount_web/templates/_journal_table.html
yagebu: I guess another helper function and a `REPLACEME` would work

aumayr:
Thanks for the new PR - works now as expected!

aumayr:
Do we need any further work on this? Or can I merge it into `master`?

yagebu:
I just discovered a small bug on the journal page. Will also push my commit for Exceptions when the dateparsing fails

aumayr:
OK then I'll wait for your PRs.

PR #63, opened 2015-12-22
Fixes, add dateparser exceptions

yagebu:


aumayr:
Looks good! I pushed a change to fix the styling of the filter boxes when there is a warning, and now I think we are ready to merge into `master`. Agree?

yagebu:
Yes. The branch seems good to merge

PR #72, opened 2015-12-24
Allow filtering for empty payee

### LINKED ISSUE #42: Enable filtering for entries with no payee
See https://github.com/aumayr/beancount-web/blob/44f322f433a9a365daf3e4e28946ff0c31a2e401/beancount_web/templates/_entry_filters.html#L116

yagebu:
Also, I've made the the headers of the filter dropdowns clickable so that they remove the value of the corresponding filter

Fixes #42

### DIFF
--- a/beancount_web/api.py
+++ b/beancount_web/api.py
@@ -175,10 +175,10 @@ class BeancountReportAPI(object):
         super(BeancountReportAPI, self).__init__()
         self.beancount_file_path = beancount_file_path
         self.filters = {
-            'time_str': None,
-            'tags': set(),
+            'time': None,
+            'tag': set(),
             'account': None,
-            'payees': set(),
+            'payee': set(),
         }
         self.load_file()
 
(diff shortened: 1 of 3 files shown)


aumayr:
Thanks! Works (almost) as expected:

- Bug: `<No Payee>` shows twice if there is no filter.
- Enhancement: What about changing `<No Payee>` to `<em>No payee</em>`? Looks nicer to the eye ;-)

yagebu:
* Can't reproduce the bug...
* Changed to italics

aumayr:
Yes, see the attached screenshot: 
screenshot_multiple_no_payee

yagebu:
Ah I see it happens if you have `""` as the payee on some transaction. In some sense, the payee can be empty in two ways: either it's None or it's the empty string. Since I didn't have the second case in my file, I added `""` to `api.active_payees`. I've removed this now.

It would be more consistent if both cases behave the same way, but that's something we can work on later (or that could maybe be changed in the beancount core).

aumayr:
Works perfectly now, thanks!

Issue #56, opened 2015-12-20
Change Watcher works only once


corani:
On my system (Ubuntu 14.04), it appears the change watcher works only once. If I make one change and refresh the page, the change shows up. Subsequent changes don't show up after page refresh. 

When I restart the application, they (obviously) do show up again.

aumayr:
@corani I cannot reproduce this behavior. Do you make the changes on the main beancount file or on an included one?

corani:
No changes, only thing that might be different is that I'm running with `--host 0.0.0.0`, not sure if that could make a difference.

aumayr:
On my Mac it works as expected (even with `--host 0.0.0.0`), but I will try to replicate it on an Ubuntu VM.

aumayr:
I tried to replicate it in a fresh Ubuntu VM with Firefox, but it still works for me. @corani  Can you double check if this bug still exists in the current `HEAD `?

Please feel free to reopen this issue if it is still not working.

corani:
I'm still seeing this issue, but I'm okay with calling it a problem on my end, until someone else can reproduce it.

aumayr:
OK, thanks!

aumayr:
(Accidentally put this issue number into the commits above)

corani:
I may have figured it out. I'm using VIM, and while watching the inotify events, I see that when I save, it generates a MOVE and DELETE. No further events after that (because it doesn't automatically starts listening on the newly created file)

aumayr:
Ahhh, now it makes sense. Thanks for investigating!

corani:
Confirmed, if I used a different editor (nano) it works as expected.

corani:
The easiest fix seems to be to watch the entire folder where beancount_file is located, though that might generate unnecessary reloads

PR #80, opened 2015-12-29
Don't use pyinotify (#56)

corani:
Some text editors will save files as "move-original, create-new, delete-original" instead of "modify-original". This causes Inotify to loose track of the source file and stop reloading. Use a watcher that runs a background task to monitor the source file instead.

### DIFF
--- a/beancount_web/cli.py
+++ b/beancount_web/cli.py
@@ -4,7 +4,7 @@ import os
 import sys
 import configparser
 
-from livereload import Server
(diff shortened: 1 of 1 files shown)


aumayr:
Thanks, @corani!

corani:
Sorry, seems I've broken it by a mistake during testing. Standby...

corani:
Fix in https://github.com/aumayr/beancount-web/pull/81

PR #81, opened 2015-12-29
Don't use background-task watcher (#56)

corani:
Sorry I messed the previous pull request ;-) Not only did I not import Watcher correctly, as it turns out the (background-task) Watcher in livereload is broken.

This is a little less neat, but it works.


aumayr:
No problem! Thanks for the contribution!

Issue #82, opened 2015-12-29
Add "Year to date" time filter [feature-request]


corani:
I generate my mortgage payment schedule automatically, so I have data for the next decade in my source file. This makes no sense in day-to-day reports, so I restrict the time range to "2015 - today".

It would be nice to have a predefined "Year to date" option for this in the time-filter dropdown.

aumayr:
Should be fairly easy to add this to the suggestions in the time-filter dropdown. Are you going to try to implement this, @corani?

corani:
Sure, I can take a stab at it. Where do you think it would fit better? In the `active_years` list, or in the `time_search_strings` search suggestions?

aumayr:
`time_search_strings` would be better, I think.

PR #83, opened 2015-12-29
Add "Year to Date" time filter

corani:
- In the API porting layer, translate "year to date" to "<current year> - today"
  (This ensures you can use it anywhere)
- Add "Year to Date" search suggestion in time filter
- Get the search suggestions from the code, instead of having them hardcoded in
  the template

Issue #82

### DIFF
--- a/beancount_web/api.py
+++ b/beancount_web/api.py
@@ -210,8 +210,12 @@ class BeancountReportAPI(object):
         self.entries = self.all_entries
 
         if self.filters['time']:
+            time_filter = self.filters['time'].lower()
+            if time_filter == 'year to date':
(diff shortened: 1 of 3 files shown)


yagebu:
I would suggest adding this as a new format to the `parse_date` function and not in the api, then all the date formats stay in one module..

aumayr:
Thanks @corani for the PR!

I changed two things before merging:

- I moved the parsing of `"year to date"` to the `dateparser`-module, as it belongs there (and added `"ytd"` as a valid time-string)
- I updated the template-code a bit to avoid the capitalization that @yagebu mentioned. The suggestions are now an array instead of a dictionary. 

Everything got merged into `master`.

Issue #86, opened 2015-12-30
Commas in axis labels


redstreet:
Hovering over data points on all chart types shows commas, but but the line and bar chart axes don't have commas. Wondering if there is an easy way to turn these on? Thanks.

PR #88, opened 2015-12-30
Format Y-axis of charts (#86)

corani:
- Format the numbers on the Y-axis of various charts to have thousands separators
- Fix bug in net-worth chart where missing data caused the chart to fail to render

Issue #86


aumayr:
Great, thanks @corani!

Issue #92, opened 2016-01-06
Link for custom editor [feature request]


redstreet:
The built-in editor in beancount-web is a nice touch especially for small edits, but it would be great to link to a beancount:// url as shown in [1], to allow for the power of a custom editor to be opened up. I used this feature extensively with bean-web, and miss it in beancount-web. If you would consider adding this in, that would be great! Perhaps as an option in the .conf file?

https://groups.google.com/d/msg/beancount/mLa_Gdbs2xg/f50z-cdNCwAJ

corani:
Would the following be a good solution?

Create an option (boolean): `open-source-files-externally` (this can be set using the `--settings-file` command-line argument)

If the option is set to true, the "Source" page will list the source files as beancount URI links, instead of the current dropdown + editor. Other source-links (e.g. from `errors` or `context`) will use the beancount URI scheme as well.

`beancount://<filename>?lineno=<lineno>`

redstreet:
Sounds good! 

For clarity, I'd name the option ````open-source-files-in-external-editor ```` or simply ````use-external-editor ````.

corani:
`use-external-editor` sounds better indeed :smile: 

I'll see if I can whip up something in a pull request.

aumayr:
Sounds great! I already have the tooling ready to register the URL-scheme on a Mac, so I will add that as well. What is missing are scripts to register the scheme on Windows and Linux.

redstreet:
I can add Windows and possibly Linux scripts as well once I'm able to test.

corani:
@aumayr This issue isn't tagged as `feature` yet

corani:
@aumayr I like the read-only view you added, but the page is getting rather long due to my large history. Is there any way to make it behave more like the internal editor? (make the links load the files asynchronously)

aumayr:
@corani Yes, I can add that.

redstreet:
````use-external-editor```` works great, thank you!

I am ready to add the Windows portion of registering the beancount:// URI.  This consists of two files: a .reg registry file, and a .bat batch file. If @aumayr or @corani could please suggest a good location in the source tree to place them, I'll submit a pull request. Thanks again!

aumayr:
@redstreet sure: 

- First create a `windows` and `linux` directory in `beancount_urlscheme` and place any scripts (except python), binaries or executables that you need in there (as I did in the `mac` directory). 
[...]

PR #94, opened 2016-01-06
Add `use_external_editor` option

corani:
Add `use_external_editor` option that, when set to `True`, will cause
links to source files to open in an external editor.

This requires that:
1. `beancount-web` is running locally
2. The user has access to the source files
3. A content handler for the URI scheme `beancount://` is installed


corani:
Issue #92 

May require more styling :smile:

aumayr:
Thanks, @corani! I will add "3. A content handler for the URI scheme beancount:// is installed" with the next commits.

Issue #97, opened 2016-01-11
Filters shouldn't be shown on every page


corani:
Currently the filters (Time, Tag, Account, Payee) are shown on all pages. However, they don't make sense for a bunch of them, including `Custom Query` and the `Configuration` section pages.

PR #98, opened 2016-01-11
Hide filters on specific pages (issue #97)

corani:
Filters were shown on all pages, even those where they didn't make
sense or do anything. This removes the filters from pages such as
`Custom Query`, `Source`, `Options`, and `Errors`.

Issue #97


aumayr:
Merged - thanks!

Issue #96, opened 2016-01-11
Save Beancount Queries


corani:
One of the distinctive features of Beancount is the ability to write `SQL` like queries to create custom reports. For example, today I needed to get an overview of my expenses per month over the last year:

```sql
SELECT
  MONTH(date) AS month,
  SUM(COST(position)) AS balance
WHERE
   account ~ 'Expenses:' AND
   currency = 'CNY' AND
   YEAR(date) = 2015
GROUP BY 1
ORDER BY 1;
```

Instead of creating these queries from scratch every time (or having to save them elsewhere), it would be useful if it were possible to save them, right from the `Custom Query` page.

UI wise, I imagine a select box, `Save` and `Delete` button next to the existing `Submit Query` button. The select box would contain a list of saved queries and a `Save as new Query` option.

aumayr:
This would definitely save some time! 

What I like about `beancount-web` is that it is "stateless", in the sense that it does not have it's own database. All data is coming from `api.py` and the `.conf`-file. 

I think we should keep it that way, even when implementing a feature like this. So in order to implement this, the following would keep it that way (in my mind, anyway ;-) ):

 - Store the queries (strings) in a plain text file
 - Specify the location of that file in the `.conf`-file

aumayr:
I will create a new branch and take a shot at it.

corani:
Great, plain text file for the queries is what I had in mind as well. Maybe something like:

```ini
[Expenses per month]
SELECT
  MONTH(date) AS month,
  SUM(COST(position)) AS balance
WHERE
   account ~ 'Expenses:' AND
   currency = 'CNY' AND
   YEAR(date) = 2015
GROUP BY 1
ORDER BY 1;

[Balances from 2015]
balances from year = 2015
```

aumayr:
@corani I added a first draft of the code in e0e7319. Can you try and report back if this fits your needs?

corani:
You're almost there! This is working a treat.

A couple of issues I found:
- if the stored-queries file doesn't exist, it should be created
- `~` to indicate the user's home folder doesn't work 
- After saving a query, the select box isn't updated until after the page is reloaded
- Selecting the first "empty" option from the select box doesn't clear the query editor
- After you press `Submit Query`, the select box defaults back to the "empty" option
- If you press `Cancel` in the `Query Title` dialog, the query is saved anyway, but with an empty name
- You can save multiple queries under the same name
[...]

PR #101, opened 2016-01-13
`Delete Query` option in `Custom Query` (#96)

corani:


aumayr:
Closing this unmerged because we implemented it another way. Thanks anyway, @corani!

PR #102, opened 2016-01-13
Prevent metadata keys from linewrapping

corani:
Originally, long keys in metadata would line-wrap, causing the following keys to be indented. This caused an incomprehensible mess when you've got multiple pieces of metadata on one item.

Remove the `width` directives, and add some separation between keys and values with content (`:  `)

### DIFF
--- a/beancount_web/static/sass/styles.scss
+++ b/beancount_web/static/sass/styles.scss
@@ -628,11 +628,14 @@ article {
                         }
 
                         dt {
-                            width: 0;
                             min-width: 65px;
                             font-weight: normal;
                             font-style: italic;
                             color: #85AACF;
+                        }
+
+                        dt::after {
+                            content: ":\00a0\00a0";
                         }
 
                         dd a {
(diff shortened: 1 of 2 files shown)


corani:
By the way, how do I get the conversion from scss -> css working? It's getting a bit old editing both files by hand :smile: 

Or is it a bug that this doesn't happen automatically during install?

yagebu:
To compile the `scss` files, [install compass](http://compass-style.org/install/) and run `compass watch` in the `sass` directory.

aumayr:
@corani I don't really get what the original problem is and can't reproduce it accordingly. Can you post a screenshot of the problem you are describing (prior to the fix)?

corani:
Sure:

screenshot from 2016-01-14 16 20 01

aumayr:
Ahh, now I understand what you mean. I found a cleaner solution to fix that in c0673a9, so I will close this PR unmerged.

Thanks for the hints and effort, @corani!

corani:
No problem

PR #103, opened 2016-01-13
Api refactor

yagebu:
As api.py has gotten quite big and mess, I've split some code out into new submodules.

The `api.helpers` module contains complex code that is closer to the beancount core, whereas `api.serialization` contains functions that take beancount objects and turn them into something JSON-serializable.

The only non-refactoring change I've made is to remove `entry['change']` from balance entries in the journal. If a balance statement doesn't fail, there is no "change".

There's still some more to be done, but I didn't want this PR to get too big and I'll try to some more code cleanup in a follow-up PR.


aumayr:
Thanks @yagebu, this was long overdue :innocent:

PR #104, opened 2016-01-15
Display beancount `Query` directives in `Custom Query` (#96)

corani:
This implements @yagebu 's suggestion (in #96) to use the `Query` directive that's currently an experimental feature in beancount itself. This allows storing "canned" queries within the beancount source-file.

This pulls the "canned" queries from the source-file and displays them in a dropdown box in `Custom Query`. To add queries, add the following to the source file (I put them in a separate included file):

```beancount
option "experiment_query_directive" "TRUE"

2015-01-01 query "Expenses per Month" "
SELECT
  MONTH(date) AS month,
  SUM(COST(position)) AS balance
WHERE
   account ~ 'Expenses:' AND
   currency = 'CNY' AND
   YEAR(date) = 2015
GROUP BY 1
ORDER BY 1;
"
```

### DIFF
--- a/beancount_web/api/__init__.py
+++ b/beancount_web/api/__init__.py
@@ -29,7 +29,7 @@ from beancount.core.interpolate import compute_entries_balance
 from beancount.core.account import has_component
 from beancount.core.account_types import get_account_sign
 from beancount.core.data import get_entry, posting_sortkey, Open, Close, Note,\
-                                Document, Balance, Transaction, Pad, Event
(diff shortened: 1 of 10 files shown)


aumayr:
Thanks, @corani!

I changed the URLs, etc. to use the `hash` of the entry as the identifier, as the title may not be URL-safe (see ac78c6fb9bdb3447d907c41ae37c8bc56ee4b818).

corani:
You're right, I should have thought of that :smile:

PR #105, opened 2016-01-15
Retain stored query after submitting

corani:
When pressing `Submit` after selecting a stored query, the select-box
is cleared in the result page.

Same issue as in 25401b96


aumayr:
Thanks!

PR #107, opened 2016-01-18
Further refactorings

yagebu:
The only user-facing change this brings is that now the 'Monthly changes' bar chart has the opposite sign (as it's now just Income+Expenses). This make it slightly more consistent, since we don't invert the sign anywhere else, but I don't really have a strong opinion on the matter (the way I changed the implementation, this just happened), so I can also change it back.


aumayr:
Thanks, @yagebu! 

About the sign: I personally don't mind if it is inverted, so as long as no one complains this shall stay as it is after your changes :wink:

PR #111, opened 2016-01-20
journal: show total running balance; display posting attributes more accurately

yagebu:
I guess this should now make you happier than the last PR following the discussion in 4fca095bde527bb9e8e66f83daf6ca41761b1bfb. Except for the following ;)

> but I would like to include the Decimal and currency in the return from API as well

I've not included this at this point, as it's not used anywhere, and would pollute the API output unnecessarily. If we at some point want to use Decimal and currency, it's a trivial change to add them.


aumayr:
👍
Thanks!

PR #112, opened 2016-01-29
Ytd and collapse legs

corani:
- `Year to date` didn't include `today`
- After showing legs in journal, they couldn't be collapsed again (regression from 7f16559a)

### DIFF
--- a/beancount_web/util/dateparser.py
+++ b/beancount_web/util/dateparser.py
@@ -80,7 +80,7 @@ def parse_date(string):
     where start and end look like one of the above examples
     """
     if string.strip().lower() in ['year to date', 'ytd']:
-        return parse_date(str(datetime.date.today().year) + ' - today')
+        return parse_date(str(datetime.date.today().year) + ' - tomorrow')
 
     is_range = is_range_re.match(string)
     if is_range:
--- a/beancount_web/util/dateparser.py
+++ b/beancount_web/util/dateparser.py
@@ -40,8 +40,9 @@ def daterange(year=None, month=None, day=None, timedelta=datetime.timedelta()):
     """
     year, month, day = map(lambda x: int(x) if x else None, (year, month, day))
     if not (year or month or day):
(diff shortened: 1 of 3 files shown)


beancount_web/util/dateparser.py
```diff
@@ -80,7 +80,7 @@ def parse_date(string):
     where start and end look like one of the above examples
     """
     if string.strip().lower() in ['year to date', 'ytd']:
-        return parse_date(str(datetime.date.today().year) + ' - today')
+        return parse_date(str(datetime.date.today().year) + ' - tomorrow')
```
yagebu: This is the wrong place to "fix" this problem. For our purpose, exclusive dateranges should be returned for all parsed dates, so this needs fixing in the `daterange` function.

corani: You're right, after a bit more testing, even a time filter of `today` doesn't actually include `today`...

beancount_web/util/dateparser.py
```diff
@@ -40,8 +40,9 @@ def daterange(year=None, month=None, day=None, timedelta=datetime.timedelta()):
     """
     year, month, day = map(lambda x: int(x) if x else None, (year, month, day))
     if not (year or month or day):
-        date = datetime.date.today() + timedelta
-        return date, date
+        start = datetime.date.today() + timedelta
+        end = start + datetime.timedelta(days = 1)
```
corani: Adding 23:59:59 would have been more correct, but apparently only `date` is considered, `time` is ignored.

yagebu:
:+1:

aumayr:
Thanks! Works so far, but the Journal Leg toggling still has a bug: When showing all legs ("Show legs") and then clicking on a transaction, the legs will be shown twice.

corani:
Good catch, I missed that one.

It was pretty much the same issue (missing `parent-hash`)

aumayr:
Thanks, @corani, works now perfectly!

PR #119, opened 2016-02-09
Add Yearly and Monthly totals to account balance pages

xentac:
This is hardly the most efficient way to calculate these values, but they do
use the same realization.get interface.

I'm open to suggestions on better methods to call, but I barely understand the beancount interface at this point.


aumayr:
Works for me :+1: Thanks, @xentac!

Issue #85, opened 2015-12-29
Renaming `beancount-web`


aumayr:
Martin Blais (on the Mailing list):

> A suggestion: If you gave beancount-web a more distinct name, it might alleviate some of the confusion that will occur to newcomers to the project. "beancount-web" sounds pretty close to "bean-web" and I'm pretty sure not everyone will make the distinction right away, we don't have the most favorable combo of names. Of course that's your call, but I think a better name might also give the project a stronger sense of identity and thus some extra elan.
> 
> Some totally wild ideas off the top of my head (feel free to ignore me):
> 
> - "frijoles" (beans in spanish)
> - "refritos" (https://en.wikipedia.org/wiki/Refried_beans)
> - "congri" (https://en.wikipedia.org/wiki/Moros_y_Cristianos_(food) - a cuban rice & beans mix)
> (No idea why all these spanish names come to mind.)
> - "feijoada" (bean stew, https://en.wikipedia.org/wiki/Feijoada, if you prefer pt_br)
> - "cassoulet" (https://en.wikipedia.org/wiki/Cassoulet, a french version)
> - "fasolada" (greek version of the same theme)
> - "bean-stew" (while I'm at it with the food themes)
> - "legumes" (which are beans...)
> - "fava" (cool large white beans I love to eat)
> - "corn" or "squash" (bean, corn & squash are related: https://en.wikipedia.org/wiki/Three_Sisters_(agriculture))
> - "pod" or "bean-pod" (sac that contains growing beans)
> - "gassy" (one step further from eating beans, okay I'm going a bit nuts here)
> Alright, resetting off the food theme:
> - bean-sight (after "beancount" + "Insight")
> - beanster (startup-like)
> - beanjurious (adjective)
> - bean-frontend (boring)
> - bean-bag (it's comfy)
> - nugget (a small bean of gold)
> - some other name related to accounting (unfortunately many of those are already in use for other projects)

What do you think about renaming the project, @corani, @yagebu, @blais, @jbms?

yagebu:
Giving it a

yagebu:
I definitely think that renaming it is a good idea (it's to close to the beancount commands otherwise). My additional suggestion for a new name would be `erbse` (as in the German word Erbsenzähler, the translation of beancounter). It's quite short and the namespace is not polluted by all the other `bean-*` commands.

[...]

PR #120, opened 2016-02-09
adapt to beancount's API changes

### LINKED ISSUE #117: beancount API changes
Martin on the Mailing-list:

I made some important changes to the following classes:

```
    beancount.core.position.Position
    beancount.core.inventory.Inventory
    beancount.core.data.Posting
```

"Position" objects have been restructured to reflect the structure implied
by the syntax. Instead of a .number (Decimal) and .lot (Lot) objects, a
Position now has a .units (Amount) and .cost (Cost) objects.

The "Lot" class has been removed; it has been replaced by a "Cost" class
with .number (Decimal), .currency (string), .date (datetime.date), .label
(string) attributes.

The "Inventory" class is a  list of "Position" objects, and as such it has
received corresponding adjustments and minor changes.

"Posting" objects do not contain a .position attribute anymore. Instead, the
data structure has been flattened to contain .units and .cost objects
directly. As a result, a Posting acts a bit like a subclass of a Position
rather than using composition to contain one.

During the parsing phase, a "Posting" object's .cost attribute does not
point to a "Cost" instance, but rather to a "CostSpec" instance, which
reflects the unresolved and partially incomplete lot specification from the
input syntax that needs to get resolved to a particular lot during the new
booking process.

Here's a pictorial representation of changes. Before we had:

```
 Assets:Checking    -40    QQQ       {66.32 USD, 2015-12-14} @ 70.32 USD
 '---account---'  'number' '--------------lot--------------' '--price--'
                           'currency''---cost--' '--date--'

    posting                     (Posting)
        .account                (str)
        .position               (Position)
            .number             (Decimal)
            .lot                (Lot)
                .currency       (str)
                .cost           (Amount)
                    .number     (Decimal)
                    .currency   (str)
                .lot_date       (datetime.date)
        .price                  (Amount)
            .number             (Decimal)
            .currency           (str)
```

After the changes we have something that looks more like the input:

```
 Assets:Checking  -40 QQQ   {66.32 USD,   2015-12-14, "0d6f0"} @ 70.32 USD
 '---account---' '-units-' '------------cost-----------------' '--price--'
                            number currency date       label

    posting             (Posting)
        .account        (str)
        .units          (Amount)
            .number     (Decimal)
            .currency   (str)
        .cost           (Cost, CostSpec, or None)
            .number     (Decimal)
            .currency   (str)
            .date       (datetime.date, or None)
            .label      (str, or None)
        .price          (Amount, or None)
            .number     (Decimal)
            .currency   (str)

```

And field-by-field conversions:

```
    posting.account                    -> posting.account
    posting.position.number            -> posting.units.number
    posting.position.lot.currency      -> posting.units.currency
    posting.position.lot.cost.number   -> posting.cost.number
    posting.position.lot.cost.currency -> posting.cost.currency
    posting.position.lot.lot_date      -> posting.cost.date
    (none)                             -> posting.cost.label
    posting.price.number               -> posting.price.number
    posting.price.currency             -> posting.price.currency

```

Notes:

- These changes are mostly driven by the need for the parser to output
  incomplete cost specifications. The booking of lots against existing
  inventory is isolated in a beancount.parser.booking and before booking we
  need to be able to represent missing values from the input syntax, since
  the booking process is able to interpolate and fill in these missing
  values. It was required to make important changes to the data structures
  for this reason. The parser outputs 'posting.cost' with an instance of a
  new "CostSpec" class, not "Cost". After booking, all the postings's costs
  are replaced by instances of "Cost" and "CostSpec" is gone.

- This representation is much closer to the syntax. It's also flatter and
  simpler. The original representation's design was driven by an old idea
  that the lots were to be used as the key in Inventory mappings, but as it
  turns out, operations on lots are just as well implemented with an
  association list and performance isn't a problem. Moreover, lots with a
  cost are never merged using such a mapping, only reduced against, so this
  was an unnecessary distraction.

- If you were creating Position objects in plugins or other code, please
  look at the new interface and make necessary changes to your code. You can
  create new Position objects from a pair of Amount and Cost objects.

- `beancount.core.position.lot_currency_pair()` has been converted to
  `beancount.core.position.Position.currency_pair()`.

- Renamed `beancount.core.position.Position.cost()` to
  `beancount.core.position.Position.at_cost()` due to the conflict resulting
  from the Position class changes.

yagebu:
Missing the fix for rendering of postings in the journal (but nothing else AFAIK).

### DIFF
--- a/beancount_web/api/helpers.py
+++ b/beancount_web/api/helpers.py
@@ -1,11 +1,10 @@
 from datetime import timedelta
 import re
-import collections
 
 from beancount.core import flags
 from beancount.ops.holdings import Holding
-from beancount.core.number import ZERO
 from beancount.core.data import Transaction
+from beancount.core.inventory import Inventory
 from beancount.parser import options
 from beancount.ops import prices
 from beancount.utils import bisect_key
@@ -31,24 +30,21 @@ def entries_in_inclusive_range(entries, begin_date=None, end_date=None):
 
 
 # This really belongs in beancount:src/python/beancount/ops/holdings.py
-def get_holding_from_position(lot, number, account=None, price_map=None,
-                              date=None):
+def get_holding_from_position(position, price_map=None, date=None):
     """Compute a Holding corresponding to the specified position 'pos'.
 
-    :param lot: A Lot object.
-    :param number: The number of units of 'lot' in the position.
-    :param account: A str, the name of the account, or None if not needed.
+    :param position: A Position object.
     :param price_map: A dict of prices, as built by prices.build_price_map().
     :param date: A datetime.date instance, the date at which to price the
         holdings.  If left unspecified, we use the latest price information.
 
     :return: A Holding object.
     """
-    if lot.cost is not None:
+    if position.cost is not None:
         # Get price information if we have a price_map.
         market_value = None
         if price_map is not None:
-            base_quote = (lot.currency, lot.cost.currency)
+            base_quote = (position.units.currency, position.cost.currency)
             price_date, price_number = prices.get_price(price_map,
                                                         base_quote, date)
             if price_number is not None:
@@ -56,23 +52,23 @@ def get_holding_from_position(lot, number, account=None, price_map=None,
         else:
             price_date, price_number = None, None
 
-        return Holding(account,
-                       number,
-                       lot.currency,
-                       lot.cost.number,
-                       lot.cost.currency,
-                       number * lot.cost.number,
+        return Holding(None,
+                       position.units.number,
+                       position.units.currency,
+                       position.cost.number,
+                       position.cost.currency,
+                       position.units.number * position.cost.number,
                        market_value,
                        price_number,
                        price_date)
     else:
-        return Holding(account,
-                       number,
-                       lot.currency,
+        return Holding(None,
+                       position.units.number,
+                       position.units.currency,
                        None,
-                       lot.currency,
-                       number,
-                       number,
+                       position.units.currency,
+                       position.units.number,
+                       position.units.number,
                        None,
                        None)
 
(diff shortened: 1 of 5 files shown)


aumayr:
Thanks @yagebu for the PR!

One question before I merge: What about current installations of `beancount-web` that update their code? They will break if they still use an old (generally: currently unnumbered) version of `beancount`.

I already took the steps to switch to versioned releases of `beancount-web` (with a PyPI-entry), but that is (a) very fresh and not finished and (b) I still am thinking about the renaming of the project (see #85) before making any further steps into that direction.

Anyhow, in `setup.py` and `requirements.txt` we have to somehow make the user aware of the fact that he/she has to update `beancount` as well. Is there any way to do that? Or do you have any other thoughts about this issue/renaming/versioning?

\cc @blais

aumayr:
(As you can see when you look at this PR in a browser I also experimented with Travis CI, but still (a) have not finished it and (b) have to find out what use it would have ;-) )

blais:
FWIW I'm not planning to make such changes to the API often anymore. This change has actually been bubbling up for about 6 months and I delayed it for a long while in order to avoid further such changes. I finally decided to go for it now and more this forward because (1) maintaining three separate alternative "booking" branches in the dependent libraries was starting to not make sense, and (2) it was pretty stable, I had tested it for a while and I'm confident this will stay, that this is better.

yagebu:
I guess that as long as `beancount` doesn't have releases, we can't do much about this problem. Maybe put a note in the README suggesting to upgrade both if things break.

PS: I'm still all for renaming `beancount-web` to something sexier ;)

aumayr:
@blais Thanks for clarifying! So if the API shouldn't change often anymore, would you consider making a numbered release of the current (new API) state?

@yagebu Renaming it is, then: See the new comment in #85

Issue #113, opened 2016-01-29
Yearly and Monthly account balances don't seem to make sense


xentac:
If I click on any account and look at the yearly or monthly balances, it shows me numbers that seem to represent the change in the account.  What I'd really like is the absolute balance of the account in yearly or monthly blocks (probably on the last day of the block).  Then I can look at a particular account and be happy or sad that its absolute value is trending upwards or downwards.

aumayr:
What the yearly and monthly balances do is calculate the balances for transactions in the corresponding year/month, and not represent a "running" balance and just display the intermediate balances at the end of the year/month.

aumayr:
@xentac But if you want to change the behaviour it should be fairly trivial. Enabling this behaviour via a config-setting could be a good way to keep both versions. PRs welcome :-)

xentac:
In the case where you're trading commodities it makes even less sense, it looks like I gained or lost USD when I'm really trading a commodity.

I'll see about writing some patches, but I'll have to wait till I get approval to release those patches from my company...

corani:
I kind of agree with @xentac, the term `balance` seems to imply a running balance, not a change. I'm wondering if it would make more sense to rename these tabs to something like `Yearly/Monthly changes`  and add `Yearly/Monthly balances` tabs that show a running balance.

Both views are useful to me, I'd rather not hide either behind a config-setting.

aumayr:
OK I see the point now, and like @corani's suggestion (adding both variants as separate "tabs"). I can provide the UI-adaptions (the tabs have to be visually updated if there are more than 2-3) if you want to implement this in the API, @xentac or @corani.

aumayr:
The mentioned UI-adaptions (the Journal filters now wrap correctly): a296ed8

xentac:
Pull request sent: https://github.com/aumayr/beancount-web/pull/119

aumayr:
Merged - thanks all!

PR #121, opened 2016-02-10
don't compute interval_balances by hand

yagebu:
This removes more convoluted basic arithmetic code in the API that `beancount` can handle for us.

I've also renamed the two new tabs on the account page introduced by @xentac in #119. To me, both "balance" and "totals" shouldn't refer to the "monthly balance", so now it's `balance` for the running totals, and `changes` for the monthly changes. This is also more in line with what our charts are called. (This was actually suggested by @corani in #113)

PS: I've removed the totals line at the bottom somewhere in the process. Could put it back of course, but since we already have it in the top line, I don't really think we need it. What do you think?

### DIFF
--- a/beancount_web/api/__init__.py
+++ b/beancount_web/api/__init__.py
@@ -233,7 +233,7 @@ class BeancountReportAPI(object):
             'totals': self._balances_totals(names, begin_date if not accumulate else date_first, end_date),
         } for begin_date, end_date in interval_tuples]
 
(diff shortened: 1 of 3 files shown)


corani:
I kind of like the totals in the footer, since that's what I usually first look at, it's good they're called-out instead of somewhat hidden.

aumayr:
Thanks @yagebu for another very good PR!

PR #122, opened 2016-02-11
Refactor config

yagebu:


beancount_web/application.py
```diff
@@ -24,6 +24,15 @@
 app.config['ASSETS_MANIFEST'] = None
 assets.init_app(app)
 
+app.api = BeancountReportAPI()
+
+app.config.raw = configparser.ConfigParser()
+user_config_defaults_file = os.path.join(os.path.dirname(os.path.realpath(__file__)), 'default-settings.conf')
+app.config.raw.readfp(open(user_config_defaults_file))
```
corani: Does this close the file automatically?

aumayr:
> It should be possible to use the `app` object without the cli, so
initialize the configuration and api in application.py and save it on
the global config object which is always available in the templates.

Just curious: What use-case do you have in mind for using `beancount-web` without the cli?

yagebu:
> Just curious: What use-case do you have in mind for using beancount-web without the cli?

Running the `WSGI` app on a "normal" webserver. For deployment on some webhost or just any server basically, it's more comfortable if the app object is mostly self-contained. (not that I actually have plans to do that anytime soon myself, but we should allow it)

beancount_web/application.py
yagebu: Good point, only copied the code and didn't really check it. Will change it to use `with`

yagebu:
I added some more commits... (that are unrelated to the original topic of this PR ;))

PR #123, opened 2016-02-12
Minor source editor fixes

corani:
- Return to the same file after saving
- Reset caret to line 1 when changing file
- Consistent order of source files

### DIFF
--- a/fava/api/__init__.py
+++ b/fava/api/__init__.py
@@ -451,9 +451,9 @@ class BeancountReportAPI(object):
         } for journal_entry in journal if 'balance' in journal_entry.keys()]
 
     def source_files(self):
-        return list(set(
-                        [self.beancount_file_path]
-                        + [os.path.join(os.path.dirname(self.beancount_file_path), filename) for filename in self.options['include']]
+        # Make sure the included source files are sorted, behind the main source file
+        return [self.beancount_file_path] + sorted(filter(lambda x: x != self.beancount_file_path,
+                    [os.path.join(os.path.dirname(self.beancount_file_path), filename) for filename in self.options['include']]
                 ))
 
     def source(self, file_path=None):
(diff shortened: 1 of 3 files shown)


fava/templates/source.html
```diff
@@ -15,7 +15,7 @@
                 </tr>
             </thead>
             <tbody>
-                {% for source_file in api.source_files() %}
+                {% for source_file in api.source_files()|sort %}
```
aumayr: @corani `api.source_files()` returns the source files in the order that beancount loads them - so your main file first. Are you sure you want them to be sorted by name instead?

corani: What I've been noticing is that the order in which they're displayed (for me) is pretty random between restarts. Definitely not the order in which they're specified in the main source file. (I've got my files split by year and load them in order, yet in the UI they're random)

aumayr: Ah OK I see. Should be sorted here, then:

https://github.com/aumayr/fava/blob/master/fava/api/__init__.py#L455

corani: I think it's more of a representation thing, not really worth putting it in the API. I'm happy to change it though, if you still think it's better.

aumayr: The API should be consistent in what it returns, nonetheless, so I think it's worth changing.

fava/api/__init__.py
```diff
@@ -451,9 +451,9 @@ def linechart_data(self, account_name):
         } for journal_entry in journal if 'balance' in journal_entry.keys()]
 
     def source_files(self):
-        return list(set(
```
corani: A `set` is unordered, using it to remove the duplicate main source file causes the ordering to get lost.

aumayr:
Perfect, thanks!

PR #124, opened 2016-02-12
simplify api.all_accounts; generate less json by hand

yagebu:


aumayr:
There was a small bug (https://github.com/aumayr/fava/blob/master/fava/application.py#L215) but I fixed it. Thanks!

yagebu:
Good catch. Thanks :)

Issue #50, opened 2015-12-19
500-Error in wsgi:355 when running without --debug


corani:
After pulling today, I am no longer able to run `beancount-web` without `--debug`. When I do, I get an Internal Server Error (500) at `wsgi:355`.

According to `git-bisect`, this is since 838afa4a8536c1d234d6f9c3de802c361d8740c0. I suspect this may be related to #38.

When I make sure `beancount_web/static` is writable for me, the problem goes away, and I see that a `packed-general.js` file is created in `beancount_web/static/stylesheets`.

aumayr:
This is definitely related, thanks for reporting. 

I don't know how to recreate this on my Mac - which OS are you using, @corani ?

corani:
I'm on Linux (Ubuntu 14.04).

beancount-web is installed as `root`, so the `static/stylesheets` folder is owned as such, however, I run it as a normal user, which therefor doesn't have write access at that location.

aumayr:
I don't know how to fix this properly, so I removed the "packaging" of JS-files altogether, serving the individual files again. This shouldn't be a performance problem, as the browser caches them anyway and they are almost certainly served locally, where speed/latency is also a non-issue.

PR #125, opened 2016-02-12
Manage frontend deps

yagebu:
We're lacking a solution to manage the (increasing number of) javascript dependencies. So far everything was included file by file so there was no information of the required dependencies on a per-file basis. Keeping all of them in the repository is not correct and will makes upgrades harder.

To install Javascript packages, there seems to be no way around `npm` in the JS world, and `Flask-Assets` doesn't provide a feature set that's worth using IMHO. The two competing solutions at the moment to bundle them etc. seem to be `webpack` and `browserify`. After testing both for a bit I decided to go with webpack.

I've removed all Javascript dependencies from the repo and added the needed configuration to use a combination of `npm` and `webpack` to install, minify and bundle them (oh and the `sass` too ;)).

This means there is an additional step involved to get started with the installation from source, using the Makefile or for development:

    cd fava/static; npm install; npm run dev

 but we should generate them when building a release. This way we can have the benefit of having minified files while avoiding problems like in #50. For normal users, installing from PyPI using `pip` should be the recommended route. And during development the (minimal) additional effort is definitely worth it, as it greatly simplifies the handling of dependencies (and takes care of both JS and SASS at the same time).

NOTE: this is not completely ready yet:

- [x] zoom is broken on bar charts (this might be due to the fact `chartist-plugin-zoom` is not packaged very well)
- [x] for line charts, the filter given by zoom has "+" instead of spaces
- [x] document these changes in the README
- [x] include generated files in build

I'll continue working on this, but wanted to give you a heads-up ;)


aumayr:
@yagebu Thank you very much for this effort, it looks very good! 

Please let me know if I can help somewhere!

corani:
I like to see "+145 −44,859" :+1:

aumayr:
When I look at the build process and how things get slightly more complex I think it is time to start a proper documentation of the project.

What do you think?

yagebu:
> When I look at the build process and how things get slightly more complex I think it is time to start a proper documentation of the project.

I wouldn't bother much just for the dev instructions. I think those are fine in the README. It would be nice though to have some documentation for the users, like an explanation of the options and more screenshots. But for this project something like Sphinx seems overkill and just a small website should be enough (kindof like a longer and better formatted readme with a TOC).

aumayr:
> It would be nice though to have some documentation for the users, like an explanation of the options and more screenshots. 

It could also be built right into `fava`, like the help pages in desktop applications.

yagebu:
> It could also be built right into fava, like the help pages in desktop applications.

Good idea!

yagebu:
> Please let me know if I can help somewhere!

Could you try to see why zoom isn't working for bar charts? I can't really figure it out, and since it's working for line charts, it seems to be loaded correctly...

yagebu:
Filed an issue there and added the old version of the `chartist-plugin-zoom`.

Now everything JS seems to work on my end :)

aumayr:
I think I know where the problem with the bar charts is: I changed the `chartist-plugin-zoom` code. 

A diff against the version from it's project page with the version included here shows what I have changed:

```
37,39c36,38
<         if (!(chart instanceof Chartist.Line)) {
<           return;
<         }
---
>         // if (!(chart instanceof Chartist.Line)) {
>         //   return;
>         // }
```
and 

```
197,201c196,204
<               chart.options.axisX.highLow = { low: project(x1, axisX), high: project(x2, axisX) };
<               chart.options.axisY.highLow = { low: project(y1, axisY), high: project(y2, axisY) };
< 
<               chart.update(chart.data, chart.options);
<               onZoom && onZoom(chart, reset);
---
>               if (chart instanceof Chartist.Line) {
>                 chart.options.axisX.highLow = { low: project(x1, axisX), high: project(x2, axisX) };
>                 chart.options.axisY.highLow = { low: project(y1, axisY), high: project(y2, axisY) };
> 
>                 chart.update(chart.data, chart.options);
>                 onZoom && onZoom(chart, reset);
>               } else if (chart instanceof Chartist.Bar) {
>                 onZoom && onZoom(chart, reset, x1, x2, y1, y2);
>               }
```

aumayr:
@yagebu Apart from the `chartist-plugin-zoom`, can I test and merge?

yagebu:
> Apart from the chartist-plugin-zoom, can I test and merge?

:+1: and I guess it would be good to tag a release quickly after and publish it to PyPI, so that people can just `pip install` it.

> I think I know where the problem with the bar charts is: I changed the chartist-plugin-zoom code.

Would it make sense to ask to have your code included in the main `chartist-plugin-zoom` repository?

aumayr:
Merged, and version 0.2.1 released on Github and PyPI. 

> Would it make sense to ask to have your code included in the main `chartist-plugin-zoom` repository?

There is no official support for these types of charts, and I just hacked it together quickly, but I think that how I implemented it would be the way to go (as it leaves it to the user of `chartist-plugin-zoom` how to handle the zoom-event).

corani:
I found that a relatively recent version of both NPM and NodeJS is required for things to work. The versions on my system were rather old and this caused heaps of errors that didn't point to outdated npm and nodejs being the culprit. Because of this I spent an evening trying to get it to work...

Probably a good idea to highlight this in the dev-build instructions.

aumayr:
Oh, @corani I didn't think of that because my versions were recent enough, thanks for pointing it out! Do you know which versions of npm and nodejs did not work for you?

corani:
Mine were truly ancient. Node one of the last 0.x, npm a 1.x

aumayr:
@corani fdc7cbca690d4401313bbba8da31d370fab3576d

corani:
:+1:

PR #128, opened 2016-02-14
Do the journal templating in Python again

yagebu:
I've never been really happy with doing the templating of the journal in Javascript as we loose all the beancount Python objects that we can easily and correctly deal with in Jinja. Also it didn't really feel much faster for me as it introduced the additional delay of loading the page first, parsing the Javascript and only then running the template. Since the beancount API changes, the journal was broken, so this seemed like a good time to work on this and also add some missing directives to the journal.

I've reimplemented the journal in Python. After making some performance improvement (mostly just caching `url_for` and some attributes), load time for me is now about half of what the Javascript took to complete.

The speed of the toggle checkboxes has also massively been improved as the journal doesn't have to be re-rendered every time. There are now separate toggle boxes for different transaction types like paddings (a padding is not a different kind of directive, but rather a specific transaction). This also means that there are some more configuration options.

One other thing that might improve the responsiveness of the page would be to use a list instead of a table (table redraws are more time-expensive, as a wide column at the bottom might change the whole layout), which could enable even quicker page-loads (or at least they would feel quicker as the rendering could start instantly). Maybe better left for a different PR though.

### DIFF
--- a/fava/api/serialization.py
+++ b/fava/api/serialization.py
@@ -28,7 +28,7 @@ class BeanJSONEncoder(JSONEncoder):
 
 transaction_types = {
     '*': 'cleared',
-    '!': 'uncleared',
+    '!': 'pending',
     'P': 'padding',
     'S': 'summarize',
     'T': 'transfer',
@@ -56,12 +56,14 @@ def serialize_entry(entry):
             new_entry['balance'] = entry.diff_amount + entry.amount
 
     if isinstance(entry, Transaction):
-        new_entry['transaction_type'] = transaction_types[entry.flag]
+        if entry.flag in transaction_types:
+            new_entry['transaction_type'] = transaction_types[entry.flag]
+        else:
+            new_entry['transaction_type'] = 'other'
 
-        new_entry.pop('entrys', None)
         new_entry['tags'] = entry.tags or []
         new_entry['links'] = entry.links or []
-        new_entry['legs'] = [serialize_posting(p) for p in entry.postings]
+        new_entry['postings'] = [serialize_posting(p) for p in entry.postings]
 
     return new_entry
 
--- a/fava/api/serialization.py
+++ b/fava/api/serialization.py
@@ -1,8 +1,8 @@
 import decimal
 from datetime import date, datetime
 
-from beancount.core import interpolate, compare
-from beancount.core.data import Balance, Transaction, Price
+from beancount.core import compare
+from beancount.core.data import Balance, Transaction
 from beancount.core.amount import Amount
 from beancount.core.position import Position
 from beancount.core.number import ZERO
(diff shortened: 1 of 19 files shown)


yagebu:
I fixed one small bug and would consider this complete.

aumayr:
Thanks @yagebu for the huge effort! I'm also happier with the Journal being rendered without Handlebars. 

There is a bug with metadata, but I already have a fix and will merge soon.

aumayr:
@yagebu I pushed your changes to a new branch: `jinja-journal`. 

There are still some bugs lurking:

- Metadata did not show up (iterator was missing, and the proper CSS-class, but I fixed that already)

- The serialization of the transaction type is not clear: The [Beancount Syntax Documentation](https://docs.google.com/document/d/1wAMVrKIA2qtRGmoVDSUBJGmYZSygUaR0uOMW1GV3YE0/edit?pref=2&pli=1#) is:
     > A flag is used to indicate the status of a transaction, and the particular meaning of the flag is yours to define
 
   So I am not sure if the fixed values in https://github.com/aumayr/fava/blob/jinja-journal/fava/api/serialization.py#L29-L35 are a good solution - I for one use a `?`-flag to mark transactions that need some more work (like if I do not know where this transaction comes from, etc.). Having a fixed set (and therefore an opinionated) solution is good, but we should also have a fallback mechanism for other flags.

- "Hide/Show metadata" does not work for me :-/

yagebu:
> Metadata did not show up 

Ups, I don't have any metadata in my journal, so I forgot to implement it.

> Having a fixed set (and therefore an opinionated) solution is good, but we should also have a fallback mechanism for other flags.

The ones that are generated by `beancount` should be fixed with that name in `fava` too. How about just assigning the type "other" for anything that `beancount` doesn't know and then add a checkbox "Other" to filter it? Oh and `!` should be renamed to "pending", then it's consistent with h/ledger (they only now `!` and `*` and call them pending and cleared). I guess under these semantics, `!` has the same meaning as your `?` ;)

yagebu:
> "Hide/Show metadata" does not work for me :-/

I'm not sure what other changes you needed to get the metadata working, but does something like the last commit I just pushed fix it?

aumayr:
Your fix is working - but I was not able to pull it into the branch, so I just copied the changes over O:-) (90bd305dc1df5cfe9e5d157a83a4532fa3d620c7)
In addition to that I had to change the stylesheet, but now metadata is working for me, toggling and all. 

The solution with "other" for flags not in the predefined list sounds good - can you implement that? I will then merge everything into master.

aumayr:
> One other thing that might improve the responsiveness of the page would be to use a list instead of a table (table redraws are more time-expensive, as a wide column at the bottom might change the whole layout), which could enable even quicker page-loads (or at least they would feel quicker as the rendering could start instantly). Maybe better left for a different PR though.

I can work on that after this PR is closed.

yagebu:
> The solution with "other" for flags not in the predefined list sounds good - can you implement that?

Done. I've merged your changes, so this PR should be good to go if nothing got screwed up in the process.

> I can work on that after this PR is closed.

:+1:

aumayr:
Perfect, merged it into master!

And #132 for the change from `<table>` to `<ol>`

Issue #127, opened 2016-02-14
Installing is broken


aumayr:
Things about the installation to fix:

- [ ] PyPI does not list `beancount-fava` in it's search results, so `pip3 install beancount-fava` does not work
- [x] `beancount` does miss `python-dateutil` and `ply` from it's requirements, so these two have to be installed manually
- [x] `gen/style.css` and the `gen/*.js`-files are missing from the release .zip-s, so even when installed from source it does not work

aumayr:
@yagebu What is the best workfow for the last point (`gen/..`)?

yagebu:
> PyPI does not list beancount-fava in it's search results, so pip3 install beancount-fava does not work

From looking at PyPI's bug tracker, it seems that search is somewhat broken, or the search index doesn't get updated often. Let's just hope it get's fixed quickly...

> beancount does miss python-dateutil and ply from it's requirements, so these two have to be installed manually

@blais Is there anything blocking [beancount Issue 72](https://bitbucket.org/blais/beancount/issues/72/consider-installing-with-setuptools-again)? This would greatly simplify installation of `beancount` and `fava`. @aumayr as a workaround, we could just include them in our dependencies as long as `beancount` doesn't...

> gen/style.css and the gen/*.js-files are missing from the release .zip-s, so even when installed from source it does not work

If I create a source distribution (with `python3 setup.py sdist [--format=zip]`), the generated files are included. How did you build the release?

aumayr:
I fixed some of the things mentioned above: 

- correct files are included/excluded
- description and keywords get added to PyPI
- version number is parsed from `fava/__init__.py` instead of hardcoding it twice
- `python-dateutil` and `ply` added to requirements to fix [beancount Issue 72](https://bitbucket.org/blais/beancount/issues/72/consider-installing-with-setuptools-again)

I testet it on the PyPI test-server and it worked as expected. I will push a new version to PyPI (0.2.2) to test if it will show up in the search results now.

aumayr:
What is still missing is the following:

- `beancount` does not install well, because the PyPI entry is too old (2014-09-07) and installing it via the URL to the bitbucket-repo does fail (and has to be invoked as `pip install beancount-fava --process-dependency-links' to enable the URL-fetching-feature at all).

- `beancount-fava` does not show up in the search results on PyPI. 

[...]

PR #130, opened 2016-02-15
Update setup

yagebu:
This in combination with the fixes in beancount should make the installation through `pip` work.

Note that it might also be nice to create a wheel and upload it to PyPI (`python3 setup.py bdist_wheel`), as it's the easiest way to distribute a platform-independent binary distribution.

ref #127


aumayr:
Thanks, @yagebu! I bumped the version to 0.2.3, and submitted everything to PyPI with `python3 setup.py sdist bdist_wheel upload`, so a wheel is included as well.

aumayr:
And thanks @blais for fixing the setuptools-issue!

PR #133, opened 2016-02-17
JS refactorings

yagebu:
- I've modularized the JS code some more and removed the now unnecessary `clipboard` entrypoint. It's now decided in `main.js` what modules to run on page load. Only the editor is still separate.
- Caching some selectors and using ID selectors in a few more places to improve performance.
- I've removed the JS code that sets the sidebar height and implemented a CSS solution. This means no more "flicker" on page load and the sidebar height will adjust automatically to layout changes (like toggling in the journal)

### DIFF
--- a/fava/static/javascript/charts.js
+++ b/fava/static/javascript/charts.js
@@ -7,39 +7,40 @@ require('jquery-treemap/src/jquery-treemap');
 
 require('./helpers');
 
-$(document).ready(function() {
-    $('.charts-container').html('');
+var defaultOptions = {
+    height: 240,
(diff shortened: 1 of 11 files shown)


aumayr:
Works - thanks!

> I've removed the JS code that sets the sidebar height and implemented a CSS solution. This means no more "flicker" on page load and the sidebar height will adjust automatically to layout changes (like toggling in the journal)

Arrrr I wanted to fix that for a long time, and took several shots at it, but never got it right. This solution works perfectly and is very elegant (pure CSS and no additional elements)!

Issue #131, opened 2016-02-17
Show query directives (titles) as submenus to "Query" view in sidebar


aumayr:

corani:
I can take a shot at this. Do you have something like this in mind:

- Custom Queries
  - Query 1
  - Query 2
  - Query 3

(Maybe limit the number of queries to default 5, with a configuration option for a different number?)

aumayr:
@corani Yes, exactly that! 5 as a limit sounds reasonable, with the date as the basis for sorting.

PR #135, opened 2016-02-18
Show custom queries in sidebar (#131)

corani:
Add `sidebar-show-queries` option to specify the number of custom
queries to show in the sidebar. When set to `0`, the section will
be hidden.

### DIFF
--- a/fava/application.py
+++ b/fava/application.py
@@ -214,10 +214,20 @@ def template_context():
 
         return account_name in collapse_accounts
 
+    def sidebar_queries():
+        show = app.config.user.getint('sidebar-show-queries')
+        max_length = 20
+        if show > 0:
+            truncate = lambda x: x if len(x) <= max_length else x[:(max_length - 1)].rsplit(' ', 1)[0] + '\u2026'
+            return [(url_for('get_stored_query', stored_query_hash=query['hash']), truncate(query['name'])) for query in app.api.queries()[:show]]
+        else:
+            return None
+
     return dict(url_for_current=url_for_current,
                 url_for_source=url_for_source,
                 uptodate_eligible=uptodate_eligible,
                 should_collapse_account=should_collapse_account,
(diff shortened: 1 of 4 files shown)


fava/application.py
```diff
@@ -214,10 +214,20 @@ def should_collapse_account(account_name):
 
         return account_name in collapse_accounts
 
+    def sidebar_queries():
+        show = app.config.user.getint('sidebar-show-queries')
+        max_length = 20
+        if show > 0:
+            truncate = lambda x: x if len(x) <= max_length else x[:(max_length - 1)].rsplit(' ', 1)[0] + '\u2026'
```
corani: There is a way to do this in CSS, with `text-overflow: ellipsis`, but I couldn't get that to work...

aumayr: There is a nice way to do it directly in the template with a filter: http://jinja.pocoo.org/docs/dev/templates/#truncate

I will update the code accordingly.

fava/application.py
```diff
@@ -214,10 +214,20 @@ def should_collapse_account(account_name):
 
         return account_name in collapse_accounts
 
+    def sidebar_queries():
```
yagebu: This function shouldn't exist and the whole thing should be done in the template IMHO.

aumayr:
Thanks, @corani! 

I adapted the styling slightly and moved the code to the template ( fd5712ffa4fe094a7ecbd76a0cbca27ef04848c9)

fava/application.py
corani: If truncating can be done in Jinja (as @aumayr mentioned), then it's probably better indeed to move it to the template. Currently it became an unreadable mess...

aumayr: All set and done and merged: fd5712ffa4fe094a7ecbd76a0cbca27ef04848c9 :-)

yagebu: Nice :+1:

Issue #126, opened 2016-02-13
Basic tests and Travis CI integration


aumayr:
- [x] Write basic test cases
- [x] Add Travis CI integration

PR #142, opened 2016-02-18
Tests

### LINKED ISSUE #36: Add script to "scrape-test" the frontend
@blais (on the Mailing list):

> One kind of test that is easily implemented and which has been really beneficial in avoiding total undetected breakage in bean-web is a generic "scrape everything and check for errors" automated test, like I do it:
> https://bitbucket.org/blais/beancount/src/c5252218c11b9f96c57e1423fb3f38314916149b/src/python/beancount/web/scrape.py?at=default&fileviewer=file-view-default
> 
> This just runs as part of my suite of unit tests and gives me a sense that changes to the core don't break the front-end. You may want to consider adding something like this to beancount-web.

yagebu:
Add a basic scrape test. Recursively goes through all pages and records all uses of `url_for`. For any endpoint and possible combination of filters, it loads one url. (takes ~30 seconds on my computer)

fix #36


aumayr:
Ref #126

aumayr:
Thanks @yagebu!

Interestingly the test fails for me:

```python
../fava/templates/account.html:4: in top-level template code
    {% import '_account_macros.html' as account_macros with context %}
../fava/templates/_layout.html:105: in top-level template code
    {% block content %}{% endblock %}
../fava/templates/account.html:15: in block "content"
    {% with linechart_data = api.linechart_data(account_name) %}
../fava/api/__init__.py:433: in linechart_data
    journal = self.journal(account_name, with_change_and_balance=True)
../fava/api/__init__.py:312: in journal
    postings = realization.get_postings(real_account)

################################

    def get_postings(real_account):
        """Return a sorted list a RealAccount's postings and children.

        Args:
          real_account: An instance of RealAccount.
        Returns:
          A list of Posting or directories.
        """
        # We accumulate all the postings at once here instead of incrementally
        # because we need to return them sorted.
        accumulator = []
        for real_child in iter_children(real_account):
>           accumulator.extend(real_child.txn_postings)
E           AttributeError: 'NoneType' object has no attribute 'txn_postings'

../../../../.virtualenvs/bw/lib/python3.5/site-packages/beancount-2.0b6-py3.5-macosx-10.11-x86_64.egg/beancount/core/realization.py:352: AttributeError
```

aumayr:
Aside: Maybe we should introduce `requirements.txt` in form of `dev_requirements.txt` again, as eg. `pytest` is needed to run the tests (and I didn't have it in my virtualenv)?

aumayr:
@yagebu Ping me (via a comment here) if you are finished (I see that you still push changes to this branch/PR).

yagebu:
Done. The failure was due to a bug in the api. Builds on Travis for me for 3.4 and 3.5 (on Python 3.3 beancount fails to install...)

aumayr:
Works perfect! Thanks @yagebu!

PR #144, opened 2016-02-19
Show basename for document names

corani:
Originally the whole path (which can be very long, and often includes the full account name) was shown for documents. Show only the basename.

Unfortunately there doesn't seem to be a `Jinja` basename filter, so I added a field to the document entries in the API.

### DIFF
--- a/fava/api/__init__.py
+++ b/fava/api/__init__.py
@@ -166,6 +166,9 @@ class BeancountReportAPI(object):
 
             entry = serialize_entry(posting)
 
+            if isinstance(posting, Document):
+                entry['basename'] = os.path.basename(entry['filename'])
+
             if with_change_and_balance:
                 if isinstance(posting, Balance):
                     entry['change'] = {}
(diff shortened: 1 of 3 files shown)


fava/api/__init__.py
```diff
@@ -166,6 +166,9 @@ def _journal(self, postings, include_types=None, with_change_and_balance=False):
 
             entry = serialize_entry(posting)
```
yagebu: This doesn't belong in `_journal`...

It should go in `serialize_entry` or even better, just [add a filter](http://flask.pocoo.org/docs/0.10/templating/#registering-filters) and then do it in the template.

corani: I'll move it to `serialize_entry`. Adding a filter for this one use seems a bit over-kill to me...

aumayr:
Thanks @corani - I like the idea.

But to me this definitely is a "view" thing, which does belong into the templates, not the API. So I will move this into a filter.

PR #146, opened 2016-02-19
Net worth related refactorings.

### LINKED ISSUE #141: Net worth disagrees with bean-report
When I run bean-report networth I get different numbers in my two currencies than from fava.  The bean-report numbers are higher and closer to what I was seeing about a month ago with fava.  My guess is something changed.

yagebu:
Possibly this addresses #141 (which I can't reproduce, so no idea actually ;)). Anyway these are sensible changes.

Edit: On second thought, I'm pretty sure this fixes #141. Before, only the units were passed to `get_holding_from_position`, so the `if position.cost is not None` code-path would never be reached. This means that positions that are held at cost, their units would be used instead, and the conversion in the computation of the net worth would done from the currency to the target currency and not from the cost currency, which is what beancount does.

### DIFF
--- a/fava/api/__init__.py
+++ b/fava/api/__init__.py
@@ -29,8 +29,9 @@ from beancount.core.realization import RealAccount
 from beancount.core.interpolate import compute_entries_balance
 from beancount.core.account import has_component
 from beancount.core.account_types import get_account_sign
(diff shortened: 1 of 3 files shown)


aumayr:
Thanks @yagebu!

Issue #35, opened 2015-12-15
Data up-to-date indication


aumayr:
redstreet0@gmail.com (via Mailing list):

> When displaying the balance sheet, for Assets and Liabilities, I typically always compare my current balance to what the bank/credit-card/financial institution currently shows. Once they compare correctly, I put in a 'balance' directive for that account.
> 
> It would be immensely helpful if account names were optionally color coded in the balance sheet display to help with this process, like so:
> 
> Show account in:
> - green:  if the latest posting is a balance check that passed (i.e., known-good)
> - red:    if the latest posting is a balance check that failed (i.e., known-bad)
> - yellow: if the latest posting is not a balance check (i.e., unknown)

@blais  (via Mailing list):

> That's a cool idea, and I would add maybe a "gray" if the account hasn't been updated in a while (as compared to the last available date in the file).

redstreet:
The "Update Activity per Account" on the statistics page you have seems to be a natural page to add this feature to, at least to begin with.

aumayr:
@redstreet I implemented a first draft of this in the `up-to-date-indicator`-branch in f7972aa9235a44df7616779ef6e741b5373e4f24

Can you test if this looks (and works) like you expect it to (on the Statistics page in the "Update Activity per Account"-column)? 

(Note: The `grey`-suggestion from @blais is not implemented yet.)

aumayr:
(This is really helpful! Tested it while coding it, and this definitely has to make it into `master` somehow, even if not directly on the Balance sheet but on the Statistics page).

redstreet:
It's awesome, and it passes all my basic tests! It's barely gotten out of the gate, and it's already useful to my actual data right away. Great to see this feature in action that I've always wanted so badly! Glad you find it useful too. Here are some comments:

- show account in red if _any_ balance check failed for an account (not just the last)

- the table should be sortable by the up-to-date indicator - i.e, by the color coding, so one can quickly see all accounts that are in red, etc. Perhaps it doesn't need to be sortable, but group-able by the "up-to-date-ness"

- Related to the previous point: typically, though not always, only Assets and Liability accounts correspond to physically held accounts, and thus need to be updated from the institution's statements. It would be nice if we could mark an account, using its metadata, perhaps, to participate or not participate (black list would be better) in the up-to-date indication. For eg, I would remove my entire Expenses and Equity account from the indication, so I have much less "noise" to deal with in the table. I see this as less critical though, and largely for later refinement.

[...]

PR #152, opened 2016-02-22
fix bug with up-to-date indicator

### LINKED ISSUE #143: "Red Circle" for valid balance checks
I'm seeing red circle for all balance checks in the "statistics" page. If I click the account and look at the entries, the balance transaction appears in green. 

It was working very well with an early version of beancount-web, I just updated to the latest pip installation yesterday :)

yagebu:
that should fix #143 

(still only uses the filtered entries, but I'm too busy right now to fix it)

### DIFF
--- a/fava/api/__init__.py
+++ b/fava/api/__init__.py
@@ -567,7 +567,7 @@ class BeancountReportAPI(object):
         last_entry = journal[-1]
 
         if last_entry['meta']['type'] == 'balance':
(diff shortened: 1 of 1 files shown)


aumayr:
> that should fix #143

Thanks!

> (still only uses the filtered entries, but I'm too busy right now to fix it)

No worries. The rewrite of this feature (according to #35) is still open, so as long as this is documented this way it is OK.

PR #154, opened 2016-02-23
Fix decimal import

wzyboy:
Some Linux distros do not ship cdecimal by default, and users have to explicitly install cdecimal module. Beancount detects this and will use `cdecimal.Decimal` instead of `decimal.Decimal`, this causes `isinstance(o, decimal.Decimal)` in `BeanJSONEncoder` to return `False`, thus raises `TypeError: Decimal('12.34') is not JSON serializable`.

Instead of import decimal or cdecimal directly, import from beancount.core.amount solves this issue. This will import the proper decimal implemention detected by Beancount.

### DIFF
--- a/fava/api/serialization.py
+++ b/fava/api/serialization.py
@@ -1,9 +1,8 @@
-import decimal
 from datetime import date, datetime
 
 from beancount.core import compare
 from beancount.core.data import Transaction
-from beancount.core.amount import Amount
+from beancount.core.amount import Amount, decimal
 from beancount.core.position import Position
 from beancount.core.number import ZERO
 from flask.json import JSONEncoder


wzyboy:
Here are the original comments in `beancount/core/number.py`:

```
"""The module contains the basic Decimal type import.

About Decimal usage:

- Do not import Decimal from the 'decimal' or 'cdecimal' modules; always import
  your Decimal class from beancount.core.amount.

- Prefer to use D() to create new instances of Decimal objects, which
  handles more syntax, e.g., handles None, and numbers with commas.

"""
```

corani:
Good catch!

aumayr:
Thanks @wzyboy!

Issue #160, opened 2016-02-25
API: is_leaf needs to be refactored


aumayr:
returns either `True` or `[]`

https://github.com/aumayr/fava/blob/83e30e2bed5f414ce67f364c42d783fc5f09f711/fava/api/__init__.py#L162

PR #162, opened 2016-02-25
filter: make list more compact, clean up template

yagebu:
I've removed the suggestions from the filter lists, they're nice to have on first use maybe but just take up space afterwards. And now we have the help page to get new users started.

I've also decreased the padding between the filter suggestions, before the list was just endless with a lot of payees or accounts.

### DIFF
--- a/fava/api/__init__.py
+++ b/fava/api/__init__.py
@@ -21,7 +21,7 @@ optimize for performance at all.
 
 import operator
 import os
-from datetime import date, timedelta
+from datetime import date
 
 from beancount import loader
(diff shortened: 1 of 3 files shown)


yagebu:
Added a (completely unrelated) fix for #160

aumayr:
Looks good, thanks! 

I added a placeholder in the time filter input field and added a help page entry in ed97ab2a7853fdb66471d433048a01c7f7d32793. Placeholders so that this is not completely lost to users that do not look at the help pages.

yagebu:
> I added a placeholder in the time filter input field

Good idea!

Issue #132, opened 2016-02-17
Change Journal rendering from <table> to <ol>


aumayr:
@yagebu:

> One other thing that might improve the responsiveness of the page would be to use a list instead of a table (table redraws are more time-expensive, as a wide column at the bottom might change the whole layout), which could enable even quicker page-loads (or at least they would feel quicker as the rendering could start instantly). Maybe better left for a different PR though.

yagebu:
Technically [`table-layout:fixed`](https://www.w3.org/TR/CSS2/tables.html#propdef-table-layout) should have the same performance implications, but using a list seems semantically more correct. Additionally, it will allow to express the postings as sublists of the transaction entry, which lets us scrap some of the `hidden-*` CSS classes and will simplify the toggling code.

aumayr:
@yagebu This is hard.

I started working on it in a branch locally (and can push the branch if you are interested), but there seems no way around a `flexbox`-based approach to make the "table" look good (full width, but a growing description-column, etc.). Do you have expierience with `flexbox` or a good idea how to solve this elegantly in SCSS?

yagebu:
It might be possible to use an approach with negative margins and such to get exactly one column to grow and the others to be of fixed width. But 2016 seems like the year to start using `display: flex`. Maybe doesn't give us any performance boost, but it would still simplify the page.

I'll have a go at it in the next few days.

aumayr:
> I'll have a go at it in the next few days.

Thanks! Should I push the changes I already have into a separate branch?

yagebu:
> Should I push the changes I already have into a separate branch?

Yes, will be good to have something to start with.

aumayr:
OK I just looked over the changes I already made in a local branch, and they are in an embarrassing state and would be of no real value to you :-/

aumayr:
Small rendering error in the Cost column: 

![screenshot](https://cloud.githubusercontent.com/assets/5135450/13333290/9f07d906-dc06-11e5-83c5-6d75c137b378.png)

Fixed in d463cc8

aumayr:
I think this is fixed for now. Thanks again, @yagebu!

PR #164, opened 2016-02-25
Journal

yagebu:
Use an ordered list to render the journal page (using `display: flex` to get the column widths).

I'm not sure if this increases performance. The initial page load time seems to be about the same, but the toggling seems to be work quicker (and the toggling logic got simpler).

implements #132


aumayr:
Works perfect! Thanks @yagebu!

PR #169, opened 2016-02-26
Sass cleanup

yagebu:
I've tried to clean up the CSS a bit and fix some small bugs. I've made the columns with numbers in the tree tables fixed size so that they line up.

Also, the gray dots should only be shown in combination with the colored dots, right?

### DIFF
--- a/fava/static/javascript/main.js
+++ b/fava/static/javascript/main.js
@@ -135,7 +135,7 @@ $(document).ready(function() {
     Mousetrap.bind({
         '?': function() {
             $('#overlay-wrapper').show();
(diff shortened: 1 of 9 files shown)


yagebu:
Ah, looks like you fixed some of the same bugs ;)

Will rebase

yagebu:
Rebased.

But 3cd553eea7947fc7e46dffa123e889e791f10ecc, while fixing the possibly overflowing tables on the acocunt page, breaks the tables in the balance sheet, income statement and trial balance...

aumayr:
Looks good, thanks! 👍

Made just a small adjustment: 91cf82dc903f2e962dd728eafda9e58cda5be1af

PR #170, opened 2016-02-26
up-to-date indicator: don't use filtered entries

yagebu:
Don't use the filtered entries to get account activity indicators.


aumayr:
Works, thanks!

PR #171, opened 2016-02-27
Downloads: Correct sheet order

corani:
Problem:
In the downloads for XLS, XLSX and ODS the "Query" is the first
sheet, "Results" the second. This is confusing.

Root cause:
When constructing the `pyexcel.Book` a `dict` is provided with
the required sheets. A dict is unordered, as as result, the
Sheets are added in alphabetical order to the Book.

Solution:
Use an OrderedDict, so the correct ordering is maintained.

### DIFF
--- a/fava/application.py
+++ b/fava/application.py
@@ -2,6 +2,7 @@
 import configparser
 import os
 from datetime import datetime
+from collections import OrderedDict
 import io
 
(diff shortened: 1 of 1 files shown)


aumayr:
Ahh, I missed that one. Works great, thanks!

PR #173, opened 2016-02-27
Ignore filters in Custom Query

corani:
Problem:
When a filter (e.g. Time Filter) is in use, this affects Custom
Queries as well.

Root cause:
When executing a Query, `entries` is used, which contains the
entries after filters have been applied.

Solution:
Use `all_entries` instead.


aumayr:
Good catch! Thanks!

PR #174, opened 2016-02-27
Sidebar: highlight selected Custom Query

corani:
Problem:
When selecting a specific `Custom Query`, the whole block is
highlighted.

Root cause:
Highlighting is done based on the current page. All specific
queries are part of the same `Custom Query` page.

Solution:
Add another level of highlighting, based on the query-hash.


aumayr:
Looks good, thanks!

Issue #163, opened 2016-02-25
Editor: Strip trailing white-space on save


corani:
Apparently [this functionality](https://github.com/ajaxorg/ace/blob/v1.2.0/lib/ace/ext/whitespace.js) exists in Ace, it would be helpful to keep source-files clean. 

Since I keep them in Git, I like to avoid unnecessary changes due to white-space.

PR #178, opened 2016-03-02
Trim whitespace (#163)

corani:


aumayr:
@corani Thanks for the PR!

I didn't merge, but copy over most of your changes, because I did not want to expose the whole config to JavaScript via an AJAX-call, but just assigned this one value (like the others before it) explicitly: 

8335f78ffc88d9091393ab29a6e2c696e273bb40
db36ab45df06d3225f428e1a88a2e15ff6cf6dba

And also thanks for the other two small changes, which I also copied over:

b46005e9f9a08973ffbdf1c95eb293f87f364b3e
80621cbde42043c47b979a8d79e30223c6142f5d

corani:
Sure, that works just as well :-)

PR #183, opened 2016-03-06
fix date column headers on account page

### LINKED ISSUE #182: Yearly and Monthly Totals and Changes exhibiting off by one error
Charges that occur in March 2016 are showing up in the April 2016 column, similarly with yearly changes and totals.  My guess is it's a nice off by one error.  I'll try to look into it, but I'm not in town for the next 3 days, so if someone else wants to take a look, I wouldn't complain.

yagebu:
fix #182


aumayr:
Thanks, @yagebu!

Issue #184, opened 2016-03-07
Display metadata that attached to postings


wzyboy:
I just found that metadata could be attached not only to transactions, but also to postings: [Beancount - Language Syntax §Metadata](https://docs.google.com/document/d/1wAMVrKIA2qtRGmoVDSUBJGmYZSygUaR0uOMW1GV3YE0/edit#heading=h.dplfcfjm39a1). (Not sure if Martin added this syntax recently or there is a memory gap in my head.)

However in Fava, only metadata that attached to transactions are shown. It would be nice if metadata attached to postings could be shown as well.

corani:
Good catch!

aumayr:
Fixed in #185

PR #185, opened 2016-03-07
Add metadata to postings (#184)

corani:
Metadata can not only be attached to transactions, but also to
individual postings in transactions.

Add metadata in serialization of postings, and display it in the
journal view.


aumayr:
Thanks @corani, I did miss that you could also add metadata to legs. 

I changed the code slightly: Remove the `__automatic__` key when having legs where the value is not stated explicitly and move it to a macro to avoid code duplication.

Issue #181, opened 2016-03-05
Custom query `bql` parameter mixed with `time` parameter


wzyboy:
Doing a custom query after a time range is selected leads to URL parameter error:

`http://localhost:5000/query/?time=2016?bql=balances%20from%20year%20%3D%202015%3B`

The second `?` should be `&`.

corani:
Fix in #186, the `?` was hardcoded in the URL.

wzyboy:
Thanks @corani !

corani:
Please wait for @aumayr to merge the PR before closing :wink:

aumayr:
@wzyboy @corani Merged, thanks!

PR #186, opened 2016-03-07
URL parameter error in Custom Query (#181)

corani:
Doing a custom query after a time range is selected leads to URL parameter error:
```
http://localhost:5000/query/?time=2016?bql=balances%20from%20year%20%3D%202015%3B
```
The second ? should be &.

Root cause:
The code to construct the redirect URL hardcodes the `?`.

Solution:
Check for the existance of a `?` in the URL. If it exists, use `&` instead.


aumayr:
Thanks, @corani!

PR #188, opened 2016-03-07
use display context

### LINKED ISSUE #187: Decimal display improvements
Some currencies (like JPY, KRW), or some non-currency commodities (like credit card rewards "points"), do not use fractional parts in the real world. But Fava seems to display fractional parts anyway.

If all the transactions of a certain commodity (e.g. JPY) use only integers, `bean-query` and `bean-web` displays the amounts as `100 JPY`, but in Fava `100.00 JPY`. Could Fava hide the fractional parts if all the transactions use integers?

yagebu:
fix #187

### DIFF
--- a/fava/api/__init__.py
+++ b/fava/api/__init__.py
@@ -81,6 +81,9 @@ class BeancountReportAPI(object):
         self.account_types = options.get_account_types(self.options)
 
         self.title = self.options['title']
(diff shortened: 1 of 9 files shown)


aumayr:
Should have done this from the beginning... Thanks, @yagebu

Issue #168, opened 2016-02-26
Make serializing generic query results better


aumayr:
Exporting to CSV/XLS/XLSX means converting the results into a 2-dimensional array. It would be nice to convert columns of type `Cost` (details of which type a column is are already included in the result from the API) into two columns for number and currency. 

Origin: #166 by @alexjj

> Now that you've added exporting queries to csv/xls etc. I realised it would be more useful to separate number and currency into two columns. The number column is filled red and the type error appears. It doesn't really impact the functionality of the query, although when exporting to xls/ods it's not formatted as a number and it saves it as '12.34, which obviously means formula don't recognise it as a number.
> 
> For reference the query is:
> 
> ```sql
> SELECT 
>     month, account, sum(number), currency
> FROM
>     year = 2016
> WHERE 
>     account ~ 'Expenses'
> GROUP BY month, account, currency
> ORDER by month, currency, account
> FLATTEN
> ```

alexjj:
I hope my original issue didn't confuse the two issues, and I'll reply to #166 shortly. The Query-error type not recognized is solely in fava web view. It may or may not impact how the spreadsheet columns are generated. I've used xlsx writer (I know you're not though) a _tiny_ amount in the past and remember specifying the [format](https://xlsxwriter.readthedocs.org/format.html#set_num_format) for a cell. This was the type of thing I had in mind as a possible solution. The [pyexcel docs](https://pyexcel.readthedocs.org/en/latest/tutorial03.html) weren't entirely detailed on formatters for me to be sure.

At home I use libreoffice, and opened the ODS. At work I have Excel 2010 and opened the .xls and .xlsx in that.

I use English-US (work) and English-UK (home).

The .ods I opened in libreoffice, the number column had them listed as '123.43 and to make it a number I'd have to remove the '.

For .xls and .xlsx, Excel sees the numbers as strings. It'll count the number of cells but won't do any calculations. It doesn't treat it as a number. The ' isn't there but might as well be. If I edit/activate the cell and hit enter it registers  as a number.

I hope that's a bit clearer

aumayr:
Thanks @alexjj for clarifying, I understand the problem now and will have a look at the formatters in pyexcel. 

Could you attach the ODS and XLS-files to this issue (if they do not contain any personal info)? This way I could test it on exactly those files. 

(Sidenote: This feature will probably end up in beancount, because these reports should be generated there. But if we fix issues like this now, the results will be better when we port this over to beancount.)

[...]

PR #189, opened 2016-03-08
Make serializing generic query results better (#168)

corani:
During export to Excel/ODS:

- Don't stringify integers and decimals
- Split Position into (number, currency)
- Split Inventory into (number, currency) for each Position within the Inventory

### DIFF
--- a/fava/application.py
+++ b/fava/application.py
@@ -2,8 +2,6 @@
 import configparser
 import os
 from datetime import datetime
-from collections import OrderedDict
-import io
 
 import markdown2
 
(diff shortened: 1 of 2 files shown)


yagebu:
I had a quick look at the code yesterday and the only thing I was going to suggest was to put in a separate module ;) You already have done that this so I have no further comments. Doesn't matter if it's not the most elegant code if it works (and the performance of it is quite irrelevant)

aumayr:
Looks good, thanks @corani!

PR #200, opened 2016-03-11
Lint

yagebu:
- Make `flake8` happy
- use tox to run tests (this allows to also use multiple environments locally, not just on travis)
- make travis run lint


aumayr:
Thanks @yagebu!

Issue #197, opened 2016-03-09
Source Editor: Position is lost on save


corani:
When saving a source file the current position is lost. The editor reverts back to line-1 of the main source file.

aumayr:
@corani This works for me - the cursor stays in the same line (but column 1) when I click "Save" in the Source editor.

corani:
I just noticed that things work fine in debug mode (`--debug`), but not in normal mode. No idea why that should make a difference though...

yagebu:
The difference could be that `livereload` is not enabled in debug mode currently

corani:
Good point! So probably `livereload` triggers a page reload before the `save` action has had the chance to redirect the user to the correct page (including filename and lineno)

aumayr:
Ahh of course, good catch @corani and @yagebu. 

Because on a "Save" the file is changed on the disk, `livereload` reloads the page (as it should when a source file on disk changes.)

PR #201, opened 2016-03-11
Shutdown LiveReload before saving source files (#197)

corani:
Problem:
When LiveReload is active, saving a source file caused LiveReload
to reload the page. As a result, the cursor position got lost.

Solution:
Stop LiveReload before saving. To do this, one needs to dispatch
as `LiveReloadShutDown` event on the document. For some reason I
couldn't get this to work using jQuery's `trigger` or `triggerHandler`
functions, so I lifted the code used by LiveReload itself.

### DIFF
--- a/fava/static/javascript/editor.js
+++ b/fava/static/javascript/editor.js
@@ -9,39 +9,6 @@ require('ace-builds/src-min/mode-sql');
 require('ace-builds/src-min/theme-chrome');
 require('ace-builds/src-min/ext-whitespace')
 
-// Lifted from LiveReload
-// (https://github.com/livereload/livereload-js/blob/master/dist/livereload.js#L178-L207)
-var CustomEvents = {
-    bind: function(element, eventName, handler) {
-      if (element.addEventListener) {
-        return element.addEventListener(eventName, handler, false);
-      } else if (element.attachEvent) {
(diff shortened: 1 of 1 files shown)


fava/static/javascript/editor.js
```diff
@@ -179,6 +212,8 @@ $(document).ready(function() {
             event.preventDefault();
             event.stopImmediatePropagation();
 
+            CustomEvents.fire(document, "LiveReloadShutDown");
```
aumayr: @corani A simple `$(document).trigger('LiveReloadShutDown');` does the trick for me as well - no `CustomEvents`-code needed.

Can you try if this works for you as well?

corani: That's what I tried first, but it didn't work for me. (Chromium 48.0.2564.82 Ubuntu 14.04 (64-bit))

I'll do a reset and reinstall of my local version to make sure it's not conflicting with some local changes and let you know later.

aumayr: Perfect, thanks!

corani:
You're right, after a clean build, using jQuery `$(document).trigger('...')` works just fine.

yagebu:
I've squashed your two commits and merged the change.

Thanks!

Issue #193, opened 2016-03-09
Query Editor: Add `source` link for saved queries


corani:
Rationale:
When you want to permanently change a saved query, depending on how your source files are organized, it can be cumbersome to locate the query.

Proposed solution:
Below the query editor (e.g. left of the `Submit Query` button), add a link to the source file at the location where the specific query is located.

aumayr:
Sounds good, and should be easy to implement.

corani:
Adding a link to a context page is easy enough, as is adding a link for static query pages (those loaded with a `query_hash` parameter).

However, for dynamically loaded queries (changing the query from the select box) it's a bit more tricky, as we only have the `query_hash`, but not the source file and line number.

Any ideas how to solve this?

aumayr:
We could store more than just the `query_hash` in the `value="..."`-attribute of the `<select>`, like `value="[query_hash];[source_file];[lineno]"`, and then split that value in the part of the JavaScript-code that submits the form and use it accordingly.

PR #202, opened 2016-03-11
Add `Source` link to Query Editor (#193)

corani:


aumayr:
ref #193

aumayr:
Works great, thanks @corani!

I re-styled it slightly: The link is now next to the `select`. I felt that this is the more natural way for it to be, as it is closely associated with the selected option.

aumayr:
And I like the idea with the `data-` attribute way more than my suggestion with the ugly hack in the `value`-attribute!

Issue #208, opened 2016-03-14
Editor: Support code folding


corani:
Add the possibility to fold regions of source files. In the beancount documentation, Martin gives the example of org-mode:

```
* Banking
** Bank of America

2003-01-05 open Assets:US:BofA:Checking
2003-01-05 open Assets:US:BofA:Savings

;; Transactions follow …

** TD Bank

2006-03-15 open Assets:US:TD:Cash

;; More transactions follow …
```

Alternatively, [C-style folding](https://github.com/ajaxorg/ace/wiki/Non-Standard-Code-Folding) could be implemented.

This is useful if you organize your source files not purely chronologically, but grouped by institution.

aumayr:
Fixed in #209 (01a6582b32dc5e6a5ac84db6c42009622f52a312)

PR #209, opened 2016-03-15
Enable code folding (#208)

corani:
Support Org-mode style sections for code folding:

```
* Banking
** Bank of America

...

** TD Bank

...
```


aumayr:
Works great, thanks @corani!

I slightly changed the styling: The border is now the same color as the text.

PR #216, opened 2016-03-18
Click

### LINKED ISSUE #176: When starting fava, check if beancount file exists
Right now there is no good way to tell the user that the beancount file could not be found.

yagebu:
- Use click for the CLI. Much nicer to use than `argparse`. Fix #176.
- Parse all settings explicitly when loading the settings file. Means we can just use `config[settings_key]` to access them later on. Puts all the settings-parsing logic in one place.
- If there's a user settings file, show both the default and the user file on the source page.
- Move more application logic to `application.py`

### DIFF
--- a/fava/application.py
+++ b/fava/application.py
@@ -44,8 +44,19 @@ def load_settings():
 
     for option in config.bool_options:
         app.config[option] = app.config.raw.getboolean('fava', option)
-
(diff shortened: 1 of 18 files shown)


aumayr:
`click` is awesome! I didn't know it, but it's API really looks very nice! Thanks!

Issue #228, opened 2016-03-21
Sort the Time/Year drop down menu at the top in descending order


redstreet:
The year drop down menu seems to list all available years. Users with a long transaction history have to use the scroll bar to scroll to the end to select the current year, which is the last entry.

Even worse, per Martin's suggestion with beancount, some of my account related to funds have open directives with years the funds were created, which means more irrelevant years to scroll through.

The more general common case for most folks would be the current year, followed by recent years. So it would make more sense to sort this in descending order IMHO.

![image](https://cloud.githubusercontent.com/assets/691755/13908698/7834cdda-eec5-11e5-8799-bb2ee828f100.png)

corani:
Off the top of my head, this is in templates/_layout, it's probably as
simple as adding a `|reverse` to the api call that gets the years. Feel
free to give it a try!

redstreet:
Awesome, thanks for the pointer, @corani . Done, and PR submitted.

aumayr:
@redstreet Thanks for the PR - so obvious, so helpful!

redstreet:
Reopening this feature request: can the new "Time" dropdown feature be sorted in descending order? Thanks!

aumayr:
@redstreet Fixed in eeeb4760c8ee05545cd50898934b69a8427b6f77

PR #229, opened 2016-03-21
underscore to hyphen in collapse-accounts; reverse order of years in time filter.

redstreet:
1) Minor bug caused by hyphen instead of underscore.

2) Reversed order of  years in time filter. See #228


redstreet:
Sorry, going to submit separate pull requests

PR #231, opened 2016-03-21
minor fix: underscore to hyphen in collapse-accounts

redstreet:
This was causing a minor bug that prevented the collapse-accounts feature from working if enabled.


yagebu:
Thank you! Merged (6f5dd3130a5141ffe1417e43fa0d38561b58f6aa)

PR #234, opened 2016-03-21
Documentation update: relevant account display and custom editor

redstreet:
Adding these because these are not easy to "discover" simply by looking at the web
interface.


aumayr:
Thanks!

PR #241, opened 2016-03-27
Various UI refactorings

### LINKED ISSUE #218: Improve Journal filter toggle boxes
From #210 

@yagebu:

> The toggle boxes could be improved, e.g. shift-click or ctrl-click to just show/hide one type of directive.

yagebu:
- I've put the page title in the header as there is plenty of space there.
- I've reworked the journal toggles into one row of CSS-styled toggle buttons. Fix #218 
- Also some removed some of the txn type filters (now there's just `cleared`, `pending` and `other`)
- Lots of other SASS cleanup and refactorings. We have a lot of duplicate and redundant code. I've tried to enforce sensible global styles over special cases on all the single pages.


aumayr:
@yagebu These changes are awesome!

- I like that the `<h1>`'s are in the header - makes the page look sharper and saves a ton of space. Maybe I change the styling a little bit - make the page title in the header bold or in another background-color, so it is more distinguishable, but still saves the space as it does now.
- The Journal-toggles look very good! I'll add a `cursor: pointer` to them, so it is clearer that these are clickable toggles. Also, titles for the toggles (for when you hover) with a explanation what the toggle does.
- The SCSS grew heavily towards the beginning of the project, and honestly I did forget about refactoring it. 
- "General Journal" -> "Journal" is a good decision overall. Maybe renaming the "Journal" subheader on the Account view to "Account Journal" to make it clearer what the distinction is? What do you think?

Again, thanks for the good work!

yagebu:
> Also, titles for the toggles (for when you hover) with a explanation what the toggle does.

:+1: 

> Maybe renaming the "Journal" subheader on the Account view to "Account Journal" to make it clearer what the distinction is? What do you think?

Sounds good.

aumayr:
Will add these changes when you merged.

PR #246, opened 2016-03-28
Account tree

yagebu:
I've refactored the account "tree-tables" to also use ordered lists and flexbox like the journal.

- This means we can better preserve the hierarchical structure of the account tree, which greatly simplifies the Javascript part.
- The account "table cells" are not all of the same size anymore but also indented according to the hierarchy, which looks nicer IMHO.
- The whole account "cell" is now clickable to collapse the account.

I'll clean up the SCSS some more in the coming days and then merge this.


aumayr:
@yagebu This looks awesome, not just code-wise.

I quite like account "table cells" that are intended as well. The SCSS needs some polish - things like line-heights - but shouldn't be too much.

Issue #198, opened 2016-03-10
Check out graphics from ledgible


aumayr:
https://github.com/lipidity/ledgible

corani:
The [sunburst chart](https://lipidity.com/cgi/ledgible/app.py/sunburst?a=^Expenses&a=^Income) is kind of fancy (though [this presentation](http://bl.ocks.org/kerryrodden/7090426) is a bit better)

aumayr:
This looks very nice :heart_eyes:

Will see how I can build this into fava. One problem I instantly see: This takes up a lot of space, so it probably would need it's own page. And the question is for which "data" to show this - the trial balance? For single Account pages?

What also would be nice, now that new graphics are on the table: https://github.com/DKirwan/calendar-heatmap would be a nice way to give an overview of accounts that have daily transactions, like when you are a business and you have many small invoices.

corani:
Hi like the calendar-heatmap!

Maybe we can use it to show amounts as well on the Income Statement,
instead of the current bar-chart, when the interval is "daily".

On Tue, Mar 15, 2016 at 2:52 PM, aumayr <notifications@github.com> wrote:

> This looks very nice [image: :heart_eyes:]
>
> Will see how I can build this into fava. One problem I instantly see: This
> takes up a lot of space, so it probably would need it's own page. And the
> question is for which "data" to show this - the trial balance? For single
> Account pages?
>
> What also would be nice, now that new graphics are on the table:
> https://github.com/DKirwan/calendar-heatmap would be a nice way to give
> an overview of accounts that have daily transactions, like when you are a
> business and you have many small invoices.
>
> —
> You are receiving this because you commented.
> Reply to this email directly or view it on GitHub:
> https://github.com/aumayr/fava/issues/198#issuecomment-196689473
>



-- 
Best regards,
Daniël Bos


Your government is reading your email. Slow them down with encryption.

My public key: http://goo.gl/gms497 (4096 bit RSA, id EF2D5D91)
Fingerprint  : D8D0 9FBE F075 F709 7B52  2F73 326C 2123 EF2D 5D91

aumayr:
For the sunburst chart:

Right now the charts are about 250px high, and this one would have to be much higher. Would it be OK if this chart is integrated where the others are right now, but will become much higher when clicked? Or would be a separate page/subpage more appropriate?

yagebu:
> Would it be OK if this chart is integrated where the others are right now, but will become much higher when clicked?

[...]

Issue #10, opened 2015-12-10
Treemap: Option to display negative numbers


aumayr:
redstreet0 on negative numbers:

> Yes, this is one drawback of Treemaps. However, I think that for what it would mainly be used for in personal finance (getting a quick view of the big picture of each of Expense/Income/Asset/Liabilities), it is mostly safe to simply ignore them in the treemap, especially for small numbers.

> However, one possible exception and an approach: show the magnitude of negative subaccounts, but with a special marker. Eg: show only negative accounts in red (or stripes or whatever). Allow for this to be an option, and by default, turn it off - i.e., ignore negative subaccounts like you're doing. The place where turning it on would help is to find problems. For instance, a significant negative number under an Expense or an Asset account (or a flipped Income or Liability account) might be an indication of a problem.

PR #252, opened 2016-04-07
Treemap

yagebu:
Reimplement the treemap using d3 and make it an actual treemap respecting the hierarchy.

- The chart is higher in most cases and has fixed ratio.
- Automatic resizing on window resize.
- You can zoom in by clicking the area.
- The tooltips now follow the mouse and don't flicker anymore.
- The colors are different for now, as it's simpler to use one of d3's color scales.
- Currently, this ignores the option to display negative balances. IMHO displaying negative values in a treemap is just not well-defined behaviour so I'd leave it like it is. Ref #10.

Ref #198

### DIFF
--- a/fava/static/javascript/charts.js
+++ b/fava/static/javascript/charts.js
@@ -44,9 +44,10 @@ function addInternalNodesAsLeaves(node) {
 };
 
 function treeMap() {
-    var width, height;
+    var width, height, kx, ky;
     var x = d3.scale.linear(), y = d3.scale.linear();
     var treemap = d3.layout.treemap();
+    var transitionDelay = 200;
     var div, svg, root, current_node, cells, leaves, tooltipText;
 
     function chart(div) {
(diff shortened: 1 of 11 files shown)


aumayr:
This looks gorgeous! I do like the colors as they are, as the old ones were a bit "heavy".

Will be a good starting point for me to implement the "sunburst"-charts (#198)

fava/static/javascript/charts.js
```diff
@@ -29,144 +27,234 @@ var defaultOptions = {
...
+        var copy = $.extend({}, node)
+        copy.children = null;
+        node.children.push(copy);
+        node.balance = null
+    }
+};
+
+function treeMap(div, data, currency) {
+    var width = container.width();
+    var height = width / 2.5;
```
aumayr: This is a bit much on my 24" (1920px width), maybe have a max-value of 400px?

PR #256, opened 2016-04-09
Sunburst

aumayr:
This adds a sunburst chart type (#198) on top of PR #252:

screenshot

The code is taken from this example: http://bl.ocks.org/kerryrodden/7090426
I removed the breadcrumb trail and the legend, and adapted the label in the middle to fit fava's needs.

- [x] Make the Account name in the label click-able
- [x] Maybe display two circles next to each other, as there is space available
- [x] There is a bug with the label when switching charts (especially via the `c` keyboard shortcut)
- [x] The name for this chart in the links to the different charts is not yet decided yet. It get's really crowded there, so maybe another mechanism/UI for showing which charts are available is necessary. 
- [x] Add chart to more pages

Open:

- The colors the algorithm chooses are not ideal. Ideally when the inner circle-segment is green, all it's children are shades of green as well.

### DIFF
--- a/fava/api/__init__.py
+++ b/fava/api/__init__.py
@@ -434,7 +434,6 @@ class BeancountReportAPI(object):
 
     def treemap_data(self, account_name, begin_date=None, end_date=None):
         return {
-            'label': account_name,
             'balances': self.balances(account_name, begin_date, end_date),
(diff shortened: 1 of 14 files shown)


yagebu:
Looks great!

Issue #165, opened 2016-02-25
Display budgets


aumayr:
Beancount itself is a double-book accounting tool, and does not know of the concept of "budgets". 

But the more I use fava I wish there would be a way to display information if a certain account is over budget. I do practice setting budgets for various "accounts" myself to enforce certain spending behavior, and it would be nice to integrate this into the system. 

In the "[budgeting](https://github.com/aumayr/fava/tree/budgeting)"-branch I drafted a first version of how this might work and look:

- There is a new file which contains the budget-rules. It is a simple plain-text-file and has content like this:

  ```
  Expenses:Food:Groceries
      2013-W52       100.00 USD
      2016-W01       120.00 USD
      2016-W02       110.00 USD
      2016-W06       150.00 USD
  
  Expenses:Food:Alcohol
      2016-W01        20.00 USD
      2016-W02        10.00 USD
      2016-W06         5.00 USD
  
  ...
  ```

  The format is very simple: First, the account name (`Expenses:Food:Groceries`). Then, lines with the week of the year where this budget applies to (and starts) (`2016-W01`), and finally the amount and currency (`20.00 USD`). I will call these lines "Budget entries".

- These budget entries represent the budget for the specified week of the year, and all weeks thereafter, until a new rule is set for a week after that. So you can set a budget of `10.00 USD` at the very first week of 2015 (`2015-W01`), and this budget applies to all weeks until a new entry starting in week 5 of that same year (`2015-W05`) appears and sets the new budget to `20.00 USD`. 

- A budget for a specific account, currency, start-date and end-date is then calculated by stepping through each day in that period, and adding the budget for the week of that day (divided by 7) to the total sum, until the end-date is reached. 

I included a test-file in `fava/util/test-budget.budget`, which contains the rules above for Groceries and Alcohol (accounts in the file generated by `bean-example`). 

You can test it by starting fava with a `bean-example`-generated file, and navigating to http://localhost:5000/account/Expenses:Food/monthly_changes/ (for example). It should look like this:

![screenshot](https://cloud.githubusercontent.com/assets/5135450/13332840/376008d4-dc04-11e5-9ae6-a756cb5d58af.png)

Things to do:
- Discussing if this is a good idea or not
- Discussing if it is to be included in fava or not
- Discussing if the core functionality of this should be factored out into it's own package or should stay within fava
- Writing tests for the "custom language parser"
[...]

PR #257, opened 2016-04-09
Budgeting

aumayr:
This is the current state of #165: 

- Reads the budget info from metadata on the `open` directives:

  ```
  2015-10-01 open Expenses:Books
    budget-2014-Q1: "100.00 EUR"   
    budget-2015-01: "40.00 EUR"   
  ```

- Renders the budget info if there is enough information for the specified date range
- Code for reading budget info from a separate file is there as well, but not used

screenshot3

What does work and what not:

- The parsing from metadata works pretty well, but this screams for test cases (we have Travis CI)
- Performance has to be tested as well, as the code is not written with performance in mind
- The formatting of the code is not what it should be
- It does not display the budget info in the "Monthly Balances"

What is also missing is integration into existing charts and new charts.

@cgrinds You can create PRs against the `budgeting2` branch and I will merge them, so they will be visible here. 

TODO:

- [ ] Display the budget info in the "Monthly Balances" view
- [x] If budgets are active, and present for the current period, show the balance/change even if it is Zero
- [x] Format budget "(0.00)" as yellow
- [ ] Integrate budget info into new chart: "(Monthly/Weekly...) Budget"
- [x] Testcases for the budget calculation
- [x] Sample input file(s) (we have none yet)
- [x] Documentation: Help-System and Website (Features)


cgrinds:
Thanks @aumayr.

In terms of performance - I wonder if it would make sense to add benchmarks to fava, not just for budgeting, but for fava proper. Something simple like a seeded bean-example with 20 years of data, and some benchmarks, possibly just wrapping some of the existing tests. 

Just an idea, I'm not sure how amenable the fava codebase is to that style of performance benchmarking.     Benchmarking just the Python code would also miss any client side performance issues, but is less brittle than integration style benchmarking with Selenium or phantom.js, etc. 

I did a quick search and only saw one perf issue against fava that was fixed around Dec 2015, so maybe perf isn't an issue?

Any thoughts on whether it's worth the effort?

cgrinds:
@aumayr this branch is missing the install_requires for 'modgrammar>=0.10' in setup.py

aumayr:
> Any thoughts on whether it's worth the effort?

I don't know how to do these kinds of tests (of the Python code), and if it is worth the trouble. What do you think, @yagebu?

cgrinds:
@aumayr is there documentation on running fava with the most recent version of beancount. The normal `pip install --editable .` doesn't pull the latest - presumably because it's not on pypi yet?

aumayr:
@cgrinds You have to install beancount from source: http://furius.ca/beancount/doc/install

yagebu:
Can you lint your code and remove the unused modules (`budgets_separate_file` and `budgets_metadata`)?

aumayr:
@yagebu Done (67145b6 and d61baa0)

aumayr:
But before merging, some TODOs are left (see top).

yagebu:
Rebasing the backend code is not a lot of effort, but the front-end code probably has to be rewritten from scratch. I have no interest in this feature, so that's too much work for me. I've rebased the code minus the template changes and pushed to the `budgeting` branch.

Some comments:

- Wrongly specified budget shouldn't crash but add an error message to the error list
- The budget function should be accept and return an inventory, no hacky hand-constructed "inventories" please.
- Also it should be used in `api` not in `api.serialization`

aumayr:
Thanks @yagebu! 

I have trouble integrating the `budgets`-call into the api (and also regarding these points):

> - Wrongly specified budget shouldn't crash but add an error message to the error list
> - The budget function should be accept and return an inventory, no hacky hand-constructed "inventories" please.
> - Also it should be used in `api` not in `api.serialization`

I don't know where the call should go, and how to integrate it into the existing data structures that the api returns to the UI. The UI-part is not hard, but the api-part is for me. 

Can you give more hints as to how and where to integrate the `budgets`-call into the api or provide this crucial part?

fava/default-settings.conf
```diff
@@ -28,6 +28,8 @@ journal-show-type-note = True
 journal-show-type-document = True
 journal-show-type-pad = False
 journal-show-type-query = False
+journal-show-type-custom = True
+journal-show-type-budget = True
```
yagebu: I don't see a need to distinguish these two. Two toggle buttons for these seem kind of overkill

fava/api/__init__.py
```diff
@@ -281,7 +285,7 @@ def _real_account(self, account_name, entries, begin_date=None,
     def balances(self, account_name, begin_date=None, end_date=None):
         real_account = self._real_account(account_name, self.entries,
                                           begin_date, end_date)
-        return [serialize_real_account(real_account)]
+        return [serialize_real_account(real_account, self.budgets.budget)]
```
yagebu: instead of passing the budget function, it should be called here (and the inventory it should return can be passed to `serialize_real_account`)

fava/templates/macros/_account_macros.html
```diff
@@ -69,3 +69,20 @@ <h1 class="droptarget" data-account-name="{{ account_name }}">
...
 </h1>
 {% endmacro %}
+
+{% macro balance_with_budget(currency, number, budget=None, show_currency=False, css_class="balance") %}
+<span class="{{ css_class }}">
+    {% if number %}
+        {% if budget %}
+            {% set budget = (budget - number)|round(10) %}
+            {% if budget == 0.0 %}
+                {% set budget = (budget - budget) %}  {# we have to round budget and use the weird (budget-budget) trick to make it really zero, not near zero #}
```
yagebu: this seems like quite a hack. Why is it necessary?

setup.py
```diff
@@ -34,13 +34,15 @@ def read_md(fname):
...
+        'beancount-pygments-lexer>=0.1.2',
         'beancount>=2.0b6',
         'click',
-        'pygments>=2.1.1',
-        'beancount-pygments-lexer>=0.1.2',
-        'markdown2>=2.3.0',
         'Flask>=0.10.1',
         'livereload>=2.4.1',
+        'markdown2>=2.3.0',
+        'modgrammar>=0.10',
```
yagebu: remove

yagebu:
I've added some comments to the code. I guess some of the code is outdated anyway so they might not apply but I hope they're helpful. I don't think if I'll have the time to contribute a lot to this feature in the coming weeks...

There should be some tests for the `api.budgets` part before we merge this IMHO.

fava/default-settings.conf
aumayr: If custom directives "become a thing" and people start using them, along with SQL, I think it would make sense to have a custom toggle buttom for them. 

Same goes for budgets: If this feature becomes more mature and maybe there is an official "budget" directive, it's own button makes sense as well, IMHO.

fava/templates/macros/_account_macros.html
aumayr: `budget` is a `Decimal`, and `budget = (budget - number)` does not return an exact `0.0`. It passes the `if budget == 0.0`, but then it is not really zero, and so the template-`{% if budget > 0.0 %}` always is greater than 0.0.

I could not find another, more elegant way to make this work.

aumayr:
Thanks, @yagebu, these pointers are enough for me to figure out how to integrate it.

For the tests: I will add some. 

Thanks again!

aumayr:
Further UI-changes (styling, etc.) only make sense after another rebase to the current master, as I plan to integrate budgets into different charts/create new charts, etc.

For the tests, we have to wait until the custom directives make it into an official PyPI-release, otherwise pytest does not work.

fava/templates/macros/_account_macros.html
yagebu: Hmm ok, but this should happen in the api then, were we could test for it.

aumayr:
@yagebu I integrated the budgets-function into the api, added some tests and a help-page.

I think it is ready to merge into master. There is still work to be done - graphs and other niceties, but generally it works very well for me, the tests pass (although more could be added), and as it is a big changeset it diverges more and more from `master`, which will make it harder and harder to integrate/merge.

What do you think?

fava/api/budgets.py
```diff
@@ -0,0 +1,120 @@
...
+                )
+
+                self.currencies.add(entry.values[2].value.currency)
+
+        for name, account in self.datelines.items():
+            self.datelines[name] = sorted(self.datelines[name],
+                                          key=lambda
+                                          dateline: dateline.date_start)
+
+    def _daterange(self, start_date, end_date):
```
yagebu: should be in `util.date`

fava/api/budgets.py
```diff
@@ -0,0 +1,120 @@
...
+
+        if account_name not in self.datelines.keys():
+            return currency_dict
+
+        for single_day in self._daterange(date_from, date_to):
+            dateline = self._matching_dateline(
+                self.datelines[account_name], single_day)
+
+            if dateline:
+                if dateline.currency not in currency_dict:
```
yagebu: use a defaultdict

fava/api/budgets.py
```diff
@@ -0,0 +1,120 @@
...
+        super(Dateline, self).__init__()
+        self.date_start = date_start
+        self.period = period.lower()
+        self.number = number
+        self.currency = currency
+
+    def value(self, date_):
+        return self.number / self.days(date_)
+
+    def days(self, date_):
```
yagebu: should be in `util.date`

fava/default-settings.conf
yagebu: > If this feature becomes more mature and maybe there is an official "budget" directive

Sure, but there's no budget directive yet.

yagebu:
> Format budget "(0.00)" as yellow

Is this not quite unlikely to happen, and shouldn't it be green in any case? Hitting the budget exactly is still in the budget.

fava/api/__init__.py
aumayr: When doing this the child-accounts won't have any budget information. By passing the function here it can be called for each of the children in `serialize_real_account`.

fava/default-settings.conf
aumayr: OK I see the point of not wasting the space for a toggle button that might not be used at all by the user.

So what do you think about only showing it when budget directives are used?

fava/api/budgets.py
```diff
@@ -0,0 +1,120 @@
...
+
+            if dateline:
+                if dateline.currency not in currency_dict:
+                    currency_dict[dateline.currency] = Decimal(0.0)
+                currency_dict[dateline.currency] += dateline.value(single_day)
+
+        return currency_dict
+
+
+class Dateline(object):
```
yagebu: That this is an object kind of confused me, since half of its functionality seems to be implemented in `Budgets`. I think it would be clearer to have all the date-related only functions in `util.date`, then we would only have `value` left as a method (and it's so trivial, we could just do that computation in the one place where we use it). So maybe just have `Dateline` namedtuple.

yagebu: Why is it called "dateline"? Isn't it simply a "budget"

aumayr:
> > Format budget "(0.00)" as yellow
> 
> Is this not quite unlikely to happen, and shouldn't it be green in any case? Hitting the budget exactly is still in the budget.

Making them yellow (giving 0.0 it's own color) does signal the "you have used your budget". As for green: That 0.0 is in the budget is debatable… You have nothing left to spend, so why green then? You did also not overspend, so red would be wrong either! The yellow does signal exactly this third state, namely "budget used, but not overspent".

fava/api/budgets.py
aumayr: > Why is it called "dateline"? Isn't it simply a "budget"

It represents "one line of dated info about a budget", as there might be multiple lines for different currencies, etc. But I agree with all you wrote above, and will fold this into the Budget class itself, as it does not really have any methods, etc.

yagebu:
> Making them yellow (giving 0.0 it's own color) does signal the "you have used your budget". As for green: That 0.0 is in the budget is debatable… You have nothing left to spend, so why green then? You did also not overspend, so red would be wrong either! The yellow does signal exactly this third state, namely "budget used, but not overspent".

The point I was trying to make was maybe rather that this simply won't happen (how do you hit your budget exactly??) and we have some weird rounding code which we wouldn't need otherwise AFAIK.

aumayr:
> The point I was trying to make was maybe rather that this simply won't happen (how do you hit your budget exactly??)

This is quite important to me. For example, my phone bill is pretty predictable, having a base contract with a certain allowance of GBs and minutes. The budget therefore should be hit every month exactly. But if I'm abroad and use my phone there or if I'm over my allowance of GBs, the budget will be overspent, lighting up as red.

aumayr:
@yagebu The last four commits (67e26b3 through 7753481) address your comments in the changeset.

yagebu:
I don't really think much of the history in this branch is worth keeping, so it can be squashed into three (`date.util`, `api.budgets`, the frontend stuff) or just one commit (might be easier).

aumayr:
Can you help with that? Or just post the `git` stuff/workflow I need to do that?

yagebu:
> Or just post the git stuff/workflow I need to do that?

`git rebase -i` should be your friend. Maybe read up on what it does if you've not used it before. It's really a command one should know how to use. I use it for nearly all my changes/PRs to squash commits and rebase etc.

aumayr:
Will close this as I will create a new PR

Issue #213, opened 2016-03-18
Switch between multiple root source files


corani:
Imagine you are responsible for multiple finances, e.g. your _personal_, your _side-business_ and some volunteer _organization_. Obviously you wouldn't want to mix those into one source file, so you keep separate source files, each with their own account hierarchy.

Currently you'd have to restart fava with different command-line arguments, to be able to switch between them. I'd like to be able to switch between them more naturally, e.g. from a dropdown menu in the topbar. I imagine this could replace the current `Beancount` title.

aumayr:
Ahh this is a very good idea! 

I personally have multiple ones (my own, my Mom's side business, my Dad's car restoration project), and so this would be much welcomed! 

A few thoughts:

- You could start it by using `fava [file1] [file2] ...`
- Switcher, as you suggested, at the top left (the title of the beancount file as a (nicely styled) dropdown/`<select>`)
- As the application is practically "stateless", meaning every view can be recreated by a permalink with the correct `GET`-params, the currently selected beancount file would have to be tracked via a `GET`-param as well (like `?b=1` for the first file in the "list of root-files" (`file1` in the sample above), etc.)
- #214 would allow settings per beancount file (currently there is just a "global" settings file, which would apply to all "instances"). Settings per beancount file would be very nice!

yagebu:
Good suggestion!

> As the application is practically "stateless", meaning every view can be recreated by a permalink with the correct GET-params, the currently selected beancount file would have to be tracked via a GET-param as well (like ?b=1 for the first file in the "list of root-files" (file1 in the sample above), etc.)

I would prefer keeping the file in the main part of the URL. I think `/title/report` where title is simply the title of the beancount file makes the most sense conceptually.

aumayr:
> I would prefer keeping the file in the main part of the URL. I think `/title/report` where title is simply the title of the beancount file makes the most sense conceptually.

Agreed, putting it in the Path-part of the URL is better.

corani:
One thing that immediately tripped me up: make sure you have a

```
option "title" "Something"
```

in each of your root source files, otherwise you'll see only one of them.

aumayr:
@corani Good point! We should have a check for that on startup, throwing an error when missing.

corani:
One more thing: the foreground color in the select box is white, so non-selected options are shown as white-on-white.

aumayr:
Which browser are you using? Can you upload a screenshot?

corani:
[...]

PR #259, opened 2016-04-09
Enable switching between multiple root source files

aumayr:
ref #213 

- `beancount-fava` now accepts multiple filenames
- The title (top left) is now a dropdown where one of the specified
  files can be selected
- The URL for all views is prefixed by a slug generated from the
  filename
- Also update the `test_scrape` function

TODO:

- [ ] Add documentation for this feature, maybe with a screenshot

### DIFF
--- a/fava/application.py
+++ b/fava/application.py
@@ -12,7 +12,7 @@ from werkzeug import secure_filename
 from fava import config
 from fava.api import BeancountReportAPI, FilterException
 from fava.api.serialization import BeanJSONEncoder
-from fava.util import uniquify
+from fava.util import slugify, uniquify
 from fava.util.excel import to_csv, to_excel, HAVE_EXCEL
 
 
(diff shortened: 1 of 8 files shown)


aumayr:
@yagebu Please feel free to optimize the code when you see something odd.

yagebu:
Cool, will have a look later today or tomorrow.

> The URL for all views is prefixed by a slug generated from the filename

I suggest using the title (as per beancount option) instead to generate the slug.

aumayr:
Thanks! 

Problem with the title is that you would have to parse all the files for extracting this...

yagebu:
> Problem with the title is that you would have to parse all the files for extracting this...

But that's something we do at startup anyway, right?

aumayr:
But only the currently active file. With this we would have to load all files.

yagebu:
> But only the currently active file. With this we would have to load all files.

I think we should load all files right at startup.
More files -> slower startup = expected
More files -> slow page-load when switching = annoying

The application should maintain a dict of APIs, one for each opened file.

aumayr:
This works really well for me. I think we can merge this. 

What do you think, @yagebu ?

PR #260, opened 2016-04-09
Sort treemap left to right, largest to smallest

cgrinds:


aumayr:
This breaks the treemaps for me, as it only displays the root-node ("Expenses", for example) :-/

cgrinds:
Sort function needs to be set before the nodes are added. Let me know if this works for you.

aumayr:
Works now 👍 Thanks!

Sidenote: There is something strange going on with these treemaps - they display seemingly random data to me, mixed up currencies, etc.

PR #265, opened 2016-04-13
Update README about optional dependencies

yegle:
Update README to include the optional excel feature.


aumayr:
Thanks!

We should add this to the website as well…

aumayr:
4e6a0d4

Issue #266, opened 2016-04-14
Remove "Equity" charts


corani:
Since the `Equity` treemap and sunburst charts show only "Opening Balances", they provide no useful information. Removing them would clean up the chart picker somewhat (from 15 to 11 for me)

yagebu:
Ref the discussion in #267

cgrinds:
I don't understand what a closing_balance is, but after this change the treemap and sunburst show no data. Treemap shows "Chart is empty" when I go to page `http://127.0.0.1:5000/income_statement/?interval=month` and click on `Expenses (USD)`

I did a
```
make clean
make build-js
pip install --editable .
```

I tried with my leger file (which showed charts before commit: d38334a2607addef6b70aef0ae1779ae33ec0af5) and a bean-example created one. Same empty data in both cases.

yagebu:
@cgrinds: Thank you for pointing out this mistake! "Closing" generates conversion entries such that all accounts sum up to zero. Additionally, it moves all income and expenses to an equity account. I forgot about the second, which emptied the income and expenses charts.

I've reverted the commit. I guess they should really only be used on the balance sheet.

PR #267, opened 2016-04-14
Remove `Equity` charts (#266)

corani:
Since the `Equity` treemap and sunburst charts show only "Opening Balances",
they provide no useful information. Removing them would clean up the chart
picker somewhat (from 15 to 11 for me)

### DIFF
--- a/fava/templates/balance_sheet.html
+++ b/fava/templates/balance_sheet.html
@@ -11,11 +11,9 @@
 
     {{ charts.treemap(api.options['name_assets']) }}
     {{ charts.treemap(api.options['name_liabilities']) }}
-    {{ charts.treemap(api.options['name_equity']) }}
 
     {{ charts.sunburst(api.options['name_assets']) }}
     {{ charts.sunburst(api.options['name_liabilities']) }}
-    {{ charts.sunburst(api.options['name_equity']) }}
 
     <div class="halfleft">
         {% with table_title="Assets", real_accounts=api.closing_balances(api.options['name_assets']), totals=True %}


yagebu:
I'm against this. It makes more sense to me to show the same accounts the balance sheet shows. And if you filter by time, these charts do actually show more information.

corani:
Charts are supposed to help you understand data by being able to visually
compare data points. I don't see the use when there is only one data point.

For example, the treemap shows just one massive block the size of about 30%
of my screen. Since there's nothing to compare it with, I might as well
just read it from the table just below it...

Besides that, "Opening Balances" is not a metric of interest for about 99%
of the users. I've certainly never looked at it, and I'm a fairly heavy
user. I don't think such a prominent position is necessary.

On Thu, Apr 14, 2016 at 4:36 PM, Jakob Schnitzer <notifications@github.com>
wrote:

> I'm against this. It makes more sense to me to show the same accounts the
> balance sheet shows. And if you filter by time, these charts do actually
> show more information.
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/aumayr/fava/pull/267#issuecomment-209828947>
>



-- 
Best regards,
Daniël Bos


Your government is reading your email. Slow them down with encryption.

My public key: http://goo.gl/gms497 (4096 bit RSA, id EF2D5D91)
Fingerprint  : D8D0 9FBE F075 F709 7B52  2F73 326C 2123 EF2D 5D91

yagebu:
> I don't see the use when there is only one data point.

Sure if there's just the opening balances, but `beancount` can generate transactions in 5 more subaccounts of Equity.

I agree that it's not as useful as the other charts, but that doesn't make it useless. By the same argument we could remove it from the balance sheet, and as long as it's there the chart should also be there for consistency.

With more than one operating currency, the chart picker surely contains too many items. IMHO, there should only be one item in the list for each type of data that's shown in the chart (and then it would only contain 4 items on the balance sheet). That's something I'm planning to implement, I don't know when though.

Issue #214, opened 2016-03-18
Move fava settings to source file options


corani:
Tracking: [Allow unknown Options for other programs](https://bitbucket.org/blais/beancount/issues/111/allow-unknown-options-for-other-programs#)

@corani in #210
> By the way, I've been thinking, wouldn't it be nice if we would be able to put all the settings in the source file, instead of in the now separate settings file? I don't know how best to do that currently, though I've suggested @blais to open up the option directive to allow arbitrary options instead of only a predefined list.

@blais  in #210 
> It's not a bad idea to open them up, though I'd want to warn the user about unused options so we don't get into the unfortunate scenario an option is defined but has no effect. In order to do this, I'd have to build a mechanism to register for options to be at least recognized.
>
> Another problem is that the currently option values have only a limited number of types, and they're not really in the language, their semantics is parsed in the grammar. I would probably change that. In any case, you'd get at least a string.
>
> Filed a ticket: https://bitbucket.org/blais/beancount/issues/111/allow-unknown-options-for-other-programs
> 
> Can you enumerate some use cases?

corani:
@corani in [beancount#111](https://bitbucket.org/blais/beancount/issues/111/allow-unknown-options-for-other-programs#)

Some examples of how we could use this to store settings for fava:

```
option "fava.theme" "alternative"
option "fava.editor.print-margin-column" "66"
option "fava.editor.strip-trailing-whitespace" "True"
option "fava.sidebar.show-queries" "5"
option "fava.sidebar.permalinks.Pending Transactions" "<URL>"
option "fava.sidebar.permalinks.Cleared Transactions" "<URL>"
```

Basically I propose using a "<application>.<option>" scheme. Let applications register their preamble, if necessary, they can validate the remainder themselves. Optionally, for consistency, we could move all of beancount's options to "beancount.<original-option>" and deprecate the original ones.

Some values I'd like to be able to store are: Strings, Decimals, Integers, Booleans. Lists are already possible (by using the same option key multiple times). Dictionaries could be constructed by splitting on the dots.

Conceptually, I'd like to be able to do something like this (with the options above):
```python
>>> api.getOptions("fava")
{ "theme": "alternative",
  "editor": {
    "print-margin-column": 66,
    "strip-trailing-whitespace": True
  },
  "sidebar: {
    "show-queries": 5, 
    "permalinks": {
      "Pending Transactions": "<URL>",
      "Cleared Transactions": "<URL>"
    }
  }
}
```

[...]

PR #268, opened 2016-04-15
Parse Journal filters from get params

### LINKED ISSUE #239: Store Journal filters in GET-params
Split from #217, ref #210.

aumayr:
fix #239

### DIFF
--- a/fava/static/package.json
+++ b/fava/static/package.json
@@ -28,8 +28,8 @@
     "d3": "^3.5.16",
     "jquery": "^2.2.0",
     "jquery-dragster": "github:catmanjan/jquery-dragster",
-    "jquery-query-object": "^2.2.3",
     "jquery-stupid-table": "github:medsabir/Stupid-Table-Plugin#feat-npm-publish",
-    "mousetrap": "^1.5.3"
+    "mousetrap": "^1.5.3",
+    "urijs": "^1.17.1"
   }
 }
--- a/fava/static/javascript/editor.js
+++ b/fava/static/javascript/editor.js
@@ -1,4 +1,4 @@
-require('jquery-query-object');
+const URI = require('urijs');
 
 require('ace-builds/src-min/ace');
 require('ace-builds/src-min/ext-searchbox');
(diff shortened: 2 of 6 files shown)


aumayr:
@yagebu Please have a look at the code, and give your 👍 if I can merge.

fava/static/package.json
```diff
@@ -30,6 +30,7 @@
     "jquery-dragster": "github:catmanjan/jquery-dragster",
     "jquery-query-object": "^2.2.3",
     "jquery-stupid-table": "github:medsabir/Stupid-Table-Plugin#feat-npm-publish",
-    "mousetrap": "^1.5.3"
+    "mousetrap": "^1.5.3",
+    "query-string": "^4.1.0"
```
yagebu: This adds a second JS dependency to deal with the query string. Do we really need both?

yagebu:
I would propose the following format to store which types to show: Have one query parameter (simply called `show` maybe), that takes a list of types. Something like `[...]?show=open&show=closed[...]`. Makes for shorter and more "human-readable" URLs.

The list should also be complete and not depend on the config. Currently it doesn't give you real "permalinks" as only values that are different from the settings are stored.

fava/static/package.json
aumayr: Which one do you mean?

aumayr:
> I would propose the following format to store which types to show: Have one query parameter (simply called `show` maybe), that takes a list of types. Something like `[...]?show=open&show=closed[...]`. Makes for shorter and more "human-readable" URLs.

I like this idea, and should be easy to implement. 

> The list should also be complete and not depend on the config. Currently it doesn't give you real "permalinks" as only values that are different from the settings are stored.

In general this is true and a good idea as well. I hesitate to implement this for just this feature - it would be generally better to implement #214 (Move fava settings to source file options) and have all the settings in the beancount-file, as otherwise `fava`-specific settings are spread over two places, when they should be in just one.

What do you think

fava/static/package.json
yagebu: `jquery-query-object`

aumayr: Ahh, got it. Didn't see that, will fix it/remove one of the two. Do you have a preference for which to remove?

yagebu: Even the author of `jquery-query-object` [doesn't recommend](https://github.com/alrusdi/jquery-plugin-query-object#disclaimer--) using it, so I'm for keeping `query-string` instead (or considering [URI.js](https://github.com/medialize/URI.js) if we need something more powerful)

aumayr:
@yagebu Implemented in 6a003ae

aumayr:
> The list should also be complete and not depend on the config. Currently it doesn't give you real "permalinks" as only values that are different from the settings are stored.

Any use case for the full list?

yagebu:
> Any use case for the full list?

This feature exists to give you the ability to save a permalink, right? And currently it just doesn't do that. If you change your settings, the behaviour of this link will change, so it's not a permalink.

Also, if I save a link, it would be nice to be able to tell just from the link what it will do. With `?show=open` I would expect to see *just* open directives and not whatever else I've enabled in my settings. There's also no need for `hide=...` then, as everything that's not shown is hidden obviously.

aumayr:
OK, implemented it this way in dac91f8.

Issue #210, opened 2016-03-15
Permalink certain journal filtered views in the sidebar


aumayr:
@corani over at #196:

> I like the idea of being able to permalink certain journal filtered views in the sidebar. That would be a very useful addition.

I think this is worth having a discussion. There could be a new Sidebar-section with these "custom views". And we could read them from the configuration-file, where you could put an array of custom links (permalinks) to be displayed in the sidebar.

What do you think, @corani?

/cc @yagebu

yegle:
Might be related but I wish I can add a saved custom query to the sidebar

corani:
You can already add custom queries. Just save them in your source file and add the following to your settings file:

```
sidebar-show-queries = <number>
```

This still show the first <number> of queries in the sidebar (ordered by date)

corani:
@aumayr permalinks in the settings file is okay with me, but we need a title for each of them. Is there any way to put a dictionary in the settings file? Or should we create an additional section for them (so not under `[fava]`, but as key->values under e.g. `permalinks`?

By the way, I've been thinking, wouldn't it be nice if we would be able to put all the settings in the source file, instead of in the now separate settings file? I don't know how best to do that currently, though I've suggested @blais to open up the `option` directive to allow arbitrary options instead of only a predefined list.

blais:
It's not a bad idea to open them up, though I'd want to warn the user about
unused options so we don't get into the unfortunate scenario an option is
defined but has no effect. In order to do this, I'd have to build a
mechanism to register for options to be at least recognized.

Another problem is that the currently option values have only a limited
number of types, and they're not really in the language, their semantics is
parsed in the grammar. I would probably change that. In any case, you'd get
at least a string.

Filed a ticket:
https://bitbucket.org/blais/beancount/issues/111/allow-unknown-options-for-other-programs

Can you enumerate some use cases?

yagebu:
>> I like the idea of being able to permalink certain journal filtered views in the sidebar. That would be a very useful addition.

>I think this is worth having a discussion. There could be a new Sidebar-section with these "custom views". And we could read them from the configuration-file, where you could put an array of custom links (permalinks) to be displayed in the sidebar.

I think we should support saving entry filters instead/first (and support arbitrary `from`-expressions as filters). I personally don't really see a big improvement in saving the chosen view of just the journal. The toggle boxes could be improved, e.g. shift-click or ctrl-click to just show/hide one type of directive. If we have that, I can't really see a use case where a saved view would help much.

[...]

PR #269, opened 2016-04-15
Store links in sidebar

aumayr:
ref #210 

Adds a new setting to display custom links in the sidebar:

```ini
; Show custom links on top of the sidebar
sidebar-custom-links =
    Journal 2015: /journal/?time=2015
    Income Statement 2015: /income_statement/?time=2015
```

The format for the links is `[label]: [link]`. Note that `": "` can not be part of the label. 

Will be rendered like this:

screenshot

TODO:

- [ ] Add Documentation-entries.


yagebu:
This seems to be something file-specific to me, so why not use `custom` directives for this?

aumayr:
> This seems to be something file-specific to me, so why not use `custom` directives for this?

I hesitate to implement this for just this feature - it would be generally better to implement #214 (Move fava settings to source file options) and have all the settings in the beancount-file, as otherwise `fava`-specific settings are spread over two places, when they should be in just one.

What do you think about #214? Good idea? What format should we choose to store the settings in the beancount file?

yagebu:
> I hesitate to implement this for just this feature - it would be generally better to implement #214 

I view this feature as being mostly unrelated to #214. This is not a "setting" for me, but rather something like a more flexible `query` directive. So it doesn't really make sense without the context of the beancount file (just like budgets).

> What do you think about #214? Good idea? What format should we choose to store the settings in the beancount file?

Probably a good idea, although not a pressing issue IMHO. I guess the `custom` directives are the best choice right now. Before we tackle this, we should evaluate if we really need all the settings we currently have.

aumayr:
> > I hesitate to implement this for just this feature - it would be generally better to implement #214 
> 
I view this feature as being mostly unrelated to #214. This is not a "setting" for me, but rather something like a more flexible `query` directive. So it > doesn't really make sense without the context of the beancount file (just like budgets).

Implemented in 9f7cfb5

> > What do you think about #214? Good idea? What format should we choose to store the settings in the beancount file?
> 
Probably a good idea, although not a pressing issue IMHO. I guess the `custom` directives are the best choice right now. Before we tackle this, we should > evaluate if we really need all the settings we currently have. 

Let's discuss this in #214. Can you post your thoughts on the various settings there?

aumayr:
This works for me. The custom links can be specified with

```
2016-04-01 custom "fava-sidebar-link" "Income 2014" "/income_statement?time=2014"
```

Only a help entry is missing.

yagebu:
Does this still work after #259? If so, can you rebase?

I guess we'll have to wait for a new beancount release that contains the custom directive before we can merge this.

fava/api/__init__.py
```diff
@@ -626,3 +626,10 @@ def account_open_metadata(self, account_name):
             if isinstance(posting, Open):
                 return posting.meta
         return {}
+
+    def sidebar_links(self):
+        # 2016-04-01 custom "fava-sidebar-link" "Income 2014" "/income_statement?time=2014"  # noqa
+        return [(entry.values[0].value, entry.values[1].value)
```
yagebu: This runs through all entries once on each page load which is horribly inefficient for such a small feature.

aumayr: Good point! Fixed in 3c61688

aumayr:
> Does this still work after #259? If so, can you rebase?

It works, and I rebased.

> I guess we'll have to wait for a new beancount release that contains the custom directive before we can merge this.

Agreed. But I think @blais would not hesitate to release a new version with the custom directives if requested.

aumayr:
Tests keep failing because the `custom` directive is not part of the last official PyPI release of beancount. 

@blais Would you consider releasing a new version to PyPI with the `custom` directive included?

yagebu:
> It works, and I rebased.

I think something went wrong with the rebase, as it shows 40 commits here. But the diff is not big, so it's probably easiest to just create one new commit with all the changes:

    git merge master       # add new commits since the rebase
    git reset master
    # then git add and git commit

blais:
Done.
https://pypi.python.org/pypi/beancount/
2.0b10

aumayr:
@blais The files on PyPI seem to be missing, so 2.0b10 cannot be found:

https://pypi.python.org/pypi/beancount/json

blais:
PyPI is pissing me off.
Tested manually, now I can't upload to the same filename.
I baked another version b11.
Try onw.

aumayr:
It works now, thanks @blais!

> PyPI is pissing me off.

Sorry for causing trouble!

aumayr:
@yagebu I, again, messed something up with git. I will add the missing documentation here, and create a new PR once I am finished.

aumayr:
Because the git history was messed up, this PR will be closed, and the successor can be found here:

#279

PR #271, opened 2016-04-24
wip: charts: rewrite bar and line charts using d3

### LINKED ISSUE #253: Bar chart: Increase clickable area for small bars
Currently there's no minimal length of the bars in bar charts, which may result in a chart like below, where I can barely see the amount of the last bar.

screen shot 2016-04-07 at 5 58 42 pm

yagebu:
This removes the "zoom" functionality, which would apply a time filter. It never worked well for me as it was quite hard to hit the right months. I'm planning to implement pan+zoom instead.

Some improvements:
- Rescale on resize and adjust size of bars
- Number of labels to show depends on the width now
- Properly group multiple currencies together, increase mouseover area.
- Use voronoi tessellation to show tooltip for nearest point in line charts.

Fix #253


corani:
Couple of issues:

- Some of the numbers on the Y axis get cut off
- No legend in line chart (Balance sheet)
- Bar charts tooltips show 0, 1 or 2 decimals and no thousands separators
- Line chart tooltips show 9-11 decimals and no thousands separators

yagebu:
> Some of the numbers on the Y axis get cut off

This happens for numbers of which magnitude?

The other issues should be fixed in aefe25dd64e12c7334db53e6b9b97d8f09d80f9c

corani:
See these two examples. In the first, part of the number gets cut off. In the second, part of the minus.

screenshot from 2016-05-01 19 31 59
screenshot from 2016-05-01 19 32 23

corani:
I noticed another, likely related, case. When the X-axis ends with a tick, the number gets cut off as well:

screenshot from 2016-05-01 19 36 58

aumayr:
Thanks @yagebu for the great work on this!

A couple of ideas:
- I don't like the colors. I will try to come up with new suggestions and post a screenshot here.
- In some cases I wish for vertical lines (where the ticks are). Would that be hard to implement?
- In the Chartist-version the currencies in the legend were clickable, so one could hide different currencies. How hard is it to implement this in d3.js as well?

redstreet:
Nice looking charts!
Related ticket: #277. A couple more suggestions (very subjective, please feel free to not take any of these):
- Light gray gridlines usually look smoother and nicer than dotted/broken lines
- the plot looks nicer without the vertical axis line (which is unnecessary)

image


- The legend shows up at an inconvenient spot in the net worth graph. Does the library have a way to auto-place the legend? If not, could we perhaps use intelligent default locations for each plot?

image

PR #279, opened 2016-05-04
Display custom links in the sidebar

aumayr:
#210

This is the successor to #269, where git was messed up by merges/rebases/etc. A lot of discussion can be found in #269.

### DIFF
--- a/fava/application.py
+++ b/fava/application.py
@@ -55,7 +55,7 @@ def load_settings():
     for option in config.list_options:
         if app.config.raw.has_option('fava', option):
             app.config[option] = \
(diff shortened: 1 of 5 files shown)


fava/application.py
```diff
@@ -55,7 +55,7 @@ def load_settings():
     for option in config.list_options:
         if app.config.raw.has_option('fava', option):
             app.config[option] = \
-                app.config.get('fava', option).strip().split("\n")
+                app.config.raw.get('fava', option).strip().split("\n")
```
aumayr: This was a bug in `master`.

Issue #212, opened 2016-03-18
Number should be clickable in Monthly Change view


yegle:
In /account/$ACCOUNT/monthly_changes/ view, the number listed there should be link to a view that list all transactions in that month that affect the balance of this account.

Or better, I'd like to be able to hover on the number and list the transactions in a popup menu.

corani:
This shouldn't be hard, for each cell we have the `account` and the `time`, so we can easily construct a link to `/account/$ACCOUNT/?time=$TIME`, though I suggest to make the entire cell clickable, not just the number.

aumayr:
I don't like popups, but the link sounds good!

aumayr:
@yegle @corani I implemented this in PR #283 

Can you test if this works for you?

corani:
Works for me

Issue #276, opened 2016-05-02
Selected month is off by one


redstreet:
If I type "April 2016" into the time search, the address changes correctly to "?time=April+2016". However, Monthly Expenses and Monthly Income show March 2016.

yagebu:
I can't reproduce this. Is this on `master` or the last release?

redstreet:
I'm running from 567166.

cgrinds:
I'm on master and see the same - when I'm on the Income statement page. `income_statement/?time=May+2016` I see the following
![image](https://cloud.githubusercontent.com/assets/242252/15054833/68c085d4-12d7-11e6-83ad-d64c3c66ddbb.png)

yagebu:
That's quite odd, can you check the source and look for the corresponding chart data, it should look like the following:

    window.chartData.push({"date_format": "MMM YYYY", "interval_totals": [{"begin_date": "2016-05-01", "totals": {"EUR": 35.20}}], "label": "Monthly Net Profit", "type": "bar"})

If that shows the wrong date for you, I don't really understand why I can't reproduce this. Can you give more details how you installed it?

cgrinds:
@yagebu I installed fava from source. Just now, I cloned into a new directory, created a new bean-example file (seed 1), and recreated the problem.

And yes, I see the chart data, but the chart says April 2016, just like the screenshot above.
```
Request made: http://127.0.0.1:5000/example-beancount-file/income_statement/?time=May+2016

    <script>window.chartData.push({"date_format": "MMM YYYY", "interval_totals": [{"begin_date": "2016-05-01", "totals": {"IRAUSD": 1200.00, "USD": -2870.86, "VACHR": -5}}], "label": "Monthly Net Profit", "type": "bar"})</script>


    <script>window.chartData.push({"date_format": "MMM YYYY", "interval_totals": [{"begin_date": "2016-05-01", "totals": {"USD": -5239.70, "VACHR": -5}}], "label": "Monthly Income", "type": "bar"})</script>


    <script>window.chartData.push({"date_format": "MMM YYYY", "interval_totals": [{"begin_date": "2016-05-01", "totals": {"IRAUSD": 1200.00, "USD": 2368.84}}], "label": "Monthly Expenses", "type": "bar"})</script>

```

yagebu:
[...]

PR #283, opened 2016-05-06
Make numbers clickable in Account Monthly Change view

aumayr:
ref #212


yegle:
It looks like there's an off-by-one error:

The old version shows the number correctly for May
screen shot 2016-05-09 at 8 12 05 am

But in the new version same number goes to April
screen shot 2016-05-09 at 8 12 24 am

yagebu:
@yegle #276

Issue #159, opened 2016-02-23
fava on cygwin


redstreet:
This is an installation question. I've been running the erstwhile beancount-web on cygwin fine. The  last version I ran was c95dc (2016-01-09). Unfortunately, I'm unable to install the latest version on Cygwin.

If I install it by running "pip3 install beancount-fava", it fails when trying to compile libxml:
    error: Cygwin gcc cannot be used with --compiler=mingw32

That's a common error, and each solution is fraught with issues. I tried compiling with gcc (that my python interpreter is compiled with) and that failed too. I can't compile with VC++ for many reasons. Note that I have libxml2 installed, and therefore, a compile here should not be needed. Is this because libxml (and not "libxml2") is mentioned as the dependency? I haven't looked.

If I install fava by hand (cloning the repo and running "python3 setup.py install"), it seems to install fine. First issue: both methods of installation should yield the same result. Second issue: when I actually run fava, it shows me a broken barebones page with the correct data but one that seems to have its css missing. I haven't debugged further.

So, summarizing the questions:
- is fava supported on cygwin? Are people running it?
- do I really need python 3.5? I'm running python 3.4 currently
- why does the installation from the repo differ from a PyPI install?
- the issue with fava showing a page without css

redstreet:
Update: tried to install from the repo onto a Linux system, and though the installation went smoothly, I still get a page without the css. The css part is my uneducated guess. Here is a screenshot:

![image](https://cloud.githubusercontent.com/assets/691755/13273033/7055a758-da55-11e5-9ccf-2fdb00b2bd1e.png)

blais:
If the problem is downstream with the Beancount install, please let me know.
You can try installing Beancount using pip3 now, with

  pip3 install beancount

redstreet:
@blais thanks, but I think Beancount looks fine. I installed the latest via pip3 after seeing the latest CHANGES, and everything seems to work fine with it so far.

aumayr:
> is fava supported on cygwin? Are people running it?

I have not tried it yet, but will in a VM. 

> do I really need python 3.5? I'm running python 3.4 currently

Python 3.4 should be fine (the Travis CI tests run for 3.4 and 3.5, and both work)

> why does the installation from the repo differ from a PyPI install?
> the issue with fava showing a page without css

[...]

PR #286, opened 2016-05-08
Documentation: added info on building python egg.

redstreet:
Reference: #159


yagebu:
Wheels are better than eggs for this, so we should explain how to build them instead. I'll  add a sentence about it to the website.

Issue #288, opened 2016-05-08
Please make clicking on the filters labels not clear them


xentac:
50% of the time I click accidentally just to open the menu and instead I clear my filters.  In addition, this makes it harder to use on an android phone (you have to hold till a menu comes up then click out of the menu).

I would really prefer it if there was a button at the top of the list to clear the filter and clicking only opened the menu (in addition to the mouseover so that we support both mouse and touch).

aumayr:
I can see the problem, and an obvious solution (as you suggested):

See PR #290 for a draft and a screenshot.

PR #290, opened 2016-05-09
Add explicit "Clear filter" button to clear filters

aumayr:
ref #288


aumayr:
What do you think about this implementation @xentac?

screenshot

xentac:
I think that's great, though I might put it under the search box so that
search shows up first. My brain didn't like

Link
Text input
Links

aumayr:
screenshot

@xentac now it is below

PR #296, opened 2016-05-25
Replace filter dropdowns with a standard form

yagebu:
Having dropdowns for the filters means that  the inputs aren't keyboard-accesible at all currently. A standard form with inputs provides better usability and is more intuitive IMHO.

- Empty inputs are styled to have a blue background so that they don't look too distracting in the header.
- When they contain some value, they are restyled to look like normal input fields. Makes it easy to spot active filters.
- They can be reached by tabbing and normal form keyboard shortcuts are available.
- The tag and payee filter can take multiple values as a comma separated list.
- `awesomplete` provides completion for the inputs. This looks nicer and is more powerful than the current solution. (actually, since this uses `<datalist>` some basic completion functionality is available even without JS)
- This greatly simplifies the filters frontend code (all filters are just strings now) and all the parsing happens in the backend, where we have tests.

One thing that's worse than before: To filter for transactions without payee, one has to give "," as the filter, otherwise we can't tell if there's no filter or the filter is for "No Payee". One thing we could do is have some string like "No Payee" actually filter for empty payee. I guess the inconsistency is acceptable (I don't think anyone has "No Payee" as the payee actually..). Any suggestions how to better handle this?


aumayr:
I like the idea, and the implementation as well! Works for me! 

What needs polish is the styling. Are you adapting the styling? Or should I have a go at it?

yagebu:
> Or should I have a go at it?

👍

aumayr:
A first go at styling the new filter controls is in the `yagebu-filters`-branch: 

https://github.com/aumayr/fava/tree/yagebu-filters

screenshots

yagebu:
> A first go at styling the new filter controls is in the yagebu-filters-branch: bd39fa6

Mhh, some aspects of this I like very much (the square styling for example or the resizing inputs), for some I very much prefer my earlier version.

It seems to be somewhat widely used, but it doesn't [seem](https://www.quora.com/What-icons-best-represent-filter-refine-in-an-iPhone-app) to be intuitive for myself and others, and I don't really see the point of the funnel icon to represent a filter. It doesn't make any sense as a symbol for it. As the point of this PR was to turn the filters into a more standard UI element, this seems to be a step in the wrong direction. We already have the "Filter" submit button anyway.

Unless there's a much better way than the weird funnel that screams "I'm a text input", I think we should really keep style input fields more like input fields. Sure, it's nice to have a header bar that's one colour, but this is an integral part of the interface and shouldn't be hidden.

yagebu:
@aumayr: The inputs only have to states now: empty (light blue background) and non-empty or active (white background). They still don't look as much as inputs as they could, but anything now would be minor changes that can easily be done separately. I've tried some things, like having a dotted underline on the input but that doesn't work well with our slim header-bar.

We should redesign the header-bar somewhat (no need to look exactly like some other websites...). I think I'll make it slightly taller and a slightly different shade of blue.

corani:
The white background for active filters make them look too much like tabs (since there's no separation with the main body)

yagebu:
> The white background for active filters make them look too much like tabs (since there's no separation with the main body)

#303

aumayr:
I really like how this turned out! 🙂

PR #300, opened 2016-05-27
bundle and use better fonts

yagebu:
- I've upped the base font size to `14px`. We want to display a lot of information, but `12px` is just tiny.
- Fira Sans is a very well-designed sans-serif font for interfaces, so use that by default and complement it with Fira Mono for the numbers.
- Use Source Code Pro in the editor.
- Use medium weight instead of bold weight in a lot of cases. The amount of bold text we had was just over the top.

Overall, this should lead to a more consistent look. We should definitely ship fonts, if anyone is unhappy with the choices I've made, then please make a suggestion for what font to use. I've tried some other fonts in the layout, including Open Sans and Source Sans Pro, but Fira just looked best IMHO (and is a bit more distinctive than Open Sans for example)

PS: a great resource regarding typography: http://practicaltypography.com/

### DIFF
--- a/fava/docs/_index.md
+++ b/fava/docs/_index.md
@@ -2,22 +2,21 @@
 title: Index
 ---
 
-Welcome to the help pages for `fava`, a web interface for the
-double-entry bookkeeping software [Beancount](http://furius.ca/beancount/) with
-a focus on features and usability.
-The box on the right contains a list of the available help pages.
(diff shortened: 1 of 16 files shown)


aumayr:
At first I was not a fan of this change, but after using it for some days now, I really like it!

PR #310, opened 2016-06-01
Add /jump handler to change url query string

yegle:


fava/application.py
```diff
@@ -7,6 +7,7 @@
                    redirect, send_from_directory, g, send_file)
 from flask_babel import Babel
 import markdown2
+import werkzeug
```
corani: Better only import what you need (i.e. werkzeug.urls), instead of everything.

yegle: Fixed, thank you.

aumayr:
@xentac Looks good, thanks! Will merge tomorrow (as I don't have time right now). 

I want to add a test for this, to make sure the function does what it should. Do you know how to add tests? Or should I have a go tomorrow?

yegle:
I don't see a test file for application.py, but I can add one. Should be easy. Will do later today.

yegle:
Hmm I rebased my code to the latest one in master, which seems to confused github's UI.

Anyway I think it's better to squash my commits into one before merging it. I'm not familiar with Github's PR feature, do I need to do the squash, or you can do it on the web UI?

aumayr:
Your rebasing looks good in Github's UI.

Thanks again!

Issue #308, opened 2016-05-30
Unable to scroll sidebar


adamgibbins:
Running latest master branch I'm unable to see the bottom of the sidebar unless I zoom out to 90% in Chrome. There's simply no way to scroll it, it doesn't move with the rest of the page.

aumayr:
In `master` the sidebar (and header) have been changed to be fixed, not scrolling with the content. 

How high is your browser window? Can you post a screenshot?

adamgibbins:
I'm running a Macbook Retina 13" at 1440x900:
<img width="1440" alt="2016-05-31_zhyyakf8fv" src="https://cloud.githubusercontent.com/assets/46830/15668432/1b1293c2-2713-11e6-9a2c-09a88c73c43e.png">

If I zoom out to 90% (personally I actually find this a better fit, doesn't bother me running like this at all, I find the spacing better etc):
<img width="1440" alt="2016-05-31_i9t2nktwh4" src="https://cloud.githubusercontent.com/assets/46830/15668465/4bc45640-2713-11e6-8f9e-988fdfb92a74.png">

aumayr:
Actually I just realized I have the same issue on my MacBook Air 13", where (with just one custom link in the sidebar) the sidebar is too high for the window.

I'll fix this in CSS.

aumayr:
This is tricky. I can make the sidebar scrollable if there is too much content, but you have to explicitly move the mouse over the sidebar and then scroll. The browser does (at least for me) not show a scrollbar, but it is there if you scroll. 

The main problem is that "Errors" are at the very bottom of the sidebar, so if you make a mistake in your Source file, this is not obvious until you scroll the sidebar, which is annoying.

So I have the following suggestions:

1. Move the error-idicator to a bubble next to the "Source"-menu-item or to it's own item below "Source" (as it was before, but hide it if there are no errors).
2. Remove "Documents" and "Notes" from the sidebar. 

What do you think @adamgibbins? /cc @yagebu

corani:
I don't think removing sections from the sidebar is a good idea: If you
then add a few more custom links, you'd run into the same problem again.

In, for example, gmail, they've solved this problem in the same way as your
first suggestion: the sidebar is scrollable, but you have to explicitly
move the mouse over it (at which time it'll show a scrollbar).

As an aside, I believe it should be possible to always show the scrollbar
on a div, though apparently some Mac versions need specific hacks for this.

[...]

PR #311, opened 2016-06-02
Make sidebar shorter and scrollable

aumayr:
This is a response to #308 with the following changes:

- "Documents" and "Notes" removed: These can be added with `custom-sidebar-links` if needed by the user
- Sidebar is scrollable
- Errors now show up beneath Source if there are any


corani:
This still keeps the `documents` and `notes` templates around. That doesn't really seem necessary to me, since these views can be rendered with the `journal` template.

Additionally, I suggest adding an example to the custom link help text, specifically showing how to bring back these views (i.e. `/<slug>/journal/?show=documents`)

aumayr:
> This still keeps the `documents` and `notes` templates around. That doesn't really seem necessary to me, since these views can be rendered with the `journal` template.

Whups, thanks for the pointer! Will remove them.

> Additionally, I suggest adding an example to the custom link help text, specifically showing how to bring back these views (i.e. `/<slug>/journal/?show=documents`)

Good idea! Will add these.

PR #312, opened 2016-06-02
Replace quarter calculation with simpler range check.

yegle:
Also updated the test to be more "table-driven".


aumayr:
Thanks 👌🏼

PR #318, opened 2016-06-05
Bug fix for /jump handler.

yegle:
A previous change introduced a bug when /jump handler tries to remove a
parameter from query string that doesn't necessary exist. This change
fixed the bug with proper exception handling and a test case to ensure
that.

### DIFF
--- a/fava/application.py
+++ b/fava/application.py
@@ -429,7 +429,10 @@ def jump():
     qs_dict = url.decode_query()
     for key, values in request.args.lists():
         if len(values) == 1 and values[0] == "":
-            del qs_dict[key]
+            try:
+                del qs_dict[key]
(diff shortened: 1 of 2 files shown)


fava/application.py
```diff
@@ -429,7 +429,10 @@ def jump():
     qs_dict = url.decode_query()
     for key, values in request.args.lists():
         if len(values) == 1 and values[0] == "":
```
corani: `and key in qs_dict` might be better than relying on exceptions, since this is not an _exceptional_ case :smile:

yegle: I'd insist on catching exception here. This is the right way of deleting an item from a python dict, even if it's a single threaded program.

PR #321, opened 2016-06-06
Add support for resolutions as low as 560px

xentac:
I couldn't get much lower than that because multiple currency income/balance tables (where you have 2+ operating currencies) get too squishy to read.

The original layout should look about the same at the original min-width but it should look way better on all pages at smaller resolutions.

This is my first pass after an afternoon of fiddling.  I'm sure I'll come up with better improvements in the future.

### DIFF
--- a/fava/application.py
+++ b/fava/application.py
@@ -375,7 +375,6 @@ def template_context():
                 url_for_source=url_for_source,
                 api=g.api,
                 operating_currencies=g.api.options['operating_currency'],
-                num_currencies=len(g.api.options['operating_currency']),
                 today=datetime.now().strftime('%Y-%m-%d'),
                 interval=request.args.get('interval',
                                           app.config['interval']),)
--- a/fava/application.py
+++ b/fava/application.py
@@ -375,6 +375,7 @@ def template_context():
                 url_for_source=url_for_source,
                 api=g.api,
                 operating_currencies=g.api.options['operating_currency'],
+                num_currencies=len(g.api.options['operating_currency']),
                 today=datetime.now().strftime('%Y-%m-%d'),
                 interval=request.args.get('interval',
                                           app.config['interval']),)
--- a/fava/static/sass/_base.scss
+++ b/fava/static/sass/_base.scss
@@ -138,7 +138,7 @@ $header_height: 50px;
 $aside_width: 160px;
 
 body {
-    padding: $header_height 0 0 $aside_width;
+    padding: $header_height 0 0 0;
 }
 
 header {
@@ -218,12 +218,36 @@ header {
     }
 }
 
+#aside_button {
+    position: fixed;
+    top: $header_height;
+    left: 0px;
+    height: 32px;
+    width: 32px;
+    background-color: $color_sidebar_background;
+    border-style: solid;
+    border-color: darken($color_sidebar_background, 16);
+    border-width: 0 1px 1px 0;
+    z-index: 1000;
+    display: block;
+}
+
+#aside_button.active {
+    left: $aside_width;
+}
+
+#aside_button svg {
+    height: 100%;
+    width: 100%;
+}
+
 aside {
     position: fixed;
-    top: 50px;
+    top: $header_height;
     bottom: 0;
     left: 0;
     padding-top: 1em;
+    margin-left: -$aside_width;
     flex-shrink: 0;
     width: $aside_width;
     color: $color_sidebar_text;
@@ -231,6 +255,7 @@ aside {
     background-color: $color_sidebar_background;
     overflow-y: auto;
     @include scrollbar(.5em, darken($color_sidebar_background, 30), transparent);
+    z-index: 1000;
 
     h3 {
         text-transform: uppercase;
@@ -298,6 +323,9 @@ aside {
             }
         }
     }
+}
+aside.active {
+    margin-left: 0;
 }
 
 
@@ -594,3 +622,18 @@ table.queryresults {
         margin-left: 3px;
     }
 }
+
+/* Modifications for ipad */
+@media (min-width: 768px) {
+    body {
+        padding: $header_height 0 0 $aside_width;
+    }
+
+    aside {
+        margin-left: 0;
+    }
+    #aside_button {
+        display: none;
+    }
+}
+
--- a/fava/static/sass/_base.scss
+++ b/fava/static/sass/_base.scss
@@ -167,6 +167,7 @@ header {
         margin: 0;
         font-size: 16px;
         font-weight: normal;
+        overflow: hidden;
 
         strong {
             margin-left: 10px;
--- a/fava/static/sass/_base.scss
+++ b/fava/static/sass/_base.scss
@@ -47,6 +47,16 @@
   }
 }
 
+$transitions: all 0.2s ease-out;
+
+@mixin smooth-transitions {
+    -webkit-transition: $transitions;
+    -moz-transition: $transitions;
+    -ms-transition: $transitions;
+    -o-transition: $transitions;
+    transition: $transitions;
+}
+
 // Globals
 
 * {
@@ -142,6 +152,7 @@ $safe_two_column: $min_width * 2 + $aside_width;
 
 body {
     padding: $header_height 0 0 0;
+    @include smooth-transitions;
 }
 
 header {
@@ -234,6 +245,7 @@ header {
     border-width: 0 1px 1px 0;
     z-index: 1000;
     display: block;
+    @include smooth-transitions;
 }
 
 #aside_button.active {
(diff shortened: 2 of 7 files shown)


corani:
Looking good! Some observations:

- Some numbers overflow the column width (the same issue in master), while the "Other" column has plenty of width left (looks to be almost twice the width of currency columns)
- Filters in the top bar take too much space on small screens (Title is clipped). Perhaps make those collapsible as well?

xentac:
It's hard to know exactly how many characters you'll need for a particular column at a particular time.  If you're using one default currency then I kept the column widths exactly the same as they were before.  It's only if you have more than one currency that columns and fonts are scaled.

I believe the default was chosen for around 10 characters in each currency column and an extra three characters for the additional currencies in the other column.

Can you post a screenshot or at the very least tell me what resolution you're looking at?

I didn't mind the title being clipped.  I rarely look at it ;)  I'd rather have filters available.  I'll think about double sizing the header (with collapse) in the case where the resolution is too small.

xentac:
I made a couple quick changes in the last few minutes.  Make sure you sync with my branch so you can see the most recent cool menu transitions ;)

corani:
See the image below for both.

screenshot_20160606-134244-01
This is on my phone (QHD)

corani:
> cool menu transitions
Nice! This caused one more issue though, the dropdown master-file picker now falls behind the sidebar.

xentac:
Oh, I see how it is.  Have you tried not having so much income? :P I'll have to test this with more large currencies.  I might be able to get away with reducing the Other column or I might have to add a large_numbers css class that will shrink all num input once it gets higher than a certain number of digits.

I think the dropdown picker was probably always hidden.  I set z-index for those menus to 1000.  I'll adjust them.

corani:
>  Have you tried not having so much income?

LOL. House prices are out of control here...

xentac:
I have fixed the file selector problem, but I had to make it click activated instead of hover activated.

I'll revisit the large numbers problem another day, in a different PR.

corani:
Works fine now. Click activated anyway is better than hover, since it works properly on touch screens.

The large number issue isn't very pressing for me (it was there before this change), if it annoys me too much, I can always reduce the number of operating currencies, since over 90% of transactions is in my primary currency anyway.

xentac:
I've thought about it a bit and the problem I run into is with more than 2 currencies, you really have limited column space.

To see what I mean, open the Chrome inspector and click the "Toggle device toolbar" button in the top left.  Choose "Responsive" as the type and enter a width of 560px.  That's the minimum screen size right now.  Even with hundreds of thousands all my numbers stick out too much.

Inspect one of the regular cells, highlight the "<span class="num">" tag, and change the font-size style until your number fits within that space.  With 4 currencies (and the other column) and millions worth of digits, I have to drop font-size down to 0.4rem to fit within the table.  This makes all values unreadable.

I'm open to suggestions.

fava/templates/_layout.html
```diff
@@ -36,23 +37,16 @@
...
-        </div>{% endif %}</h1>
+        <h1 class="site-name">{{ api.title or 'fava' }}{{ ' ▾' if config['APIS']|length > 1 else '' }}<strong>{{ page_title }}</strong></h1>
         {% if show_filters %}
         <nav>
             {% include "_entry_filters.html" %}
         </nav>
         {% endif %}
     </header>
+    <a href="#" id="aside_button">
+        <svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" enable-background="new 0 0 24 24" id="Layer_1" version="1.0" viewBox="0 0 24 24" xml:space="preserve"><circle cx="12" cy="12" r="2"/><circle cx="12" cy="5" r="2"/><circle cx="12" cy="19" r="2"/></svg>
```
yagebu: Add some width/height to the svg so that this doesn't turn out to be some freakishly huge dot if the CSS doesn't load.

xentac: Added in ad7405f.

xentac:
@aumayr is there anything else you want from this branch before you merge it?  I can fix the conflicts, but I don't want to have to do it too many times before you take a look.

fava/templates/_layout.html
```diff
@@ -101,6 +95,16 @@ <h1 class="site-name">{{ api.title or 'fava' }}{{ ' ▾' if config['APIS']|lengt
     </article>
     {% include "overlays/_keyboard_shortcuts.html" %}
     {% include "overlays/_documents_upload.html" %}
+    {% if config['APIS']|length > 1 %}
```
yagebu: Why did this code move?

yagebu:
Please make the beancount file chooser show up on hover again. If you want to implement some other solution for touch devices, that's fine, but please don't make it worse on the desktop.

fava/application.py
```diff
@@ -375,6 +375,7 @@ def url_for_source(**kwargs):
                 url_for_source=url_for_source,
                 api=g.api,
                 operating_currencies=g.api.options['operating_currency'],
+                num_currencies=len(g.api.options['operating_currency']),
```
yagebu: this can be done in the template

fava/templates/_layout.html
xentac: Because of the way z-indexing works, it became impossible for that dropdown to display above the aside menu because it was declared in the header tag before the aside tag.

The other option, I suppose, would have been to move the entire aside tag above the header, but that solution seemed weirder than moving this one down.

fava/static/sass/_base.scss
```diff
@@ -595,3 +630,36 @@ table.queryresults {
         margin-left: 3px;
     }
 }
+
+/* Modifications for ipad */
```
yagebu: Maybe explain what's different to phones and why in this comment. Won't this also be active on some of these huge modern phones too?

fava/application.py
xentac: Please tell me how.  Every time I tried to use len() inside the template it got very angry at me.

fava/static/sass/_base.scss
xentac: I used the chrome inspector to emulate the various phone resolutions and the only one that was this large was ipad.  I can add more comments explaining myself though.

fava/application.py
yagebu: `operating_currencies|length`

xentac:
I originally tried to have the menu show up on hover and click but the two actions conflicted.  If you're hovering and you click should it show up or not?  Occasionally the hover event is fired on mobile, but it's very inconsistent, so tapping sometimes makes the menu flash and then disappear, only to flash and disappear the next time you tap it.

I can put more effort into making them compatible, but it's mostly going to be guessing and fiddling.  Unless you have a reference to a site that handles both seamlessly.

fava/static/sass/_base.scss
```diff
@@ -595,3 +630,36 @@ table.queryresults {
...
+        margin-left: 0;
+    }
+    #aside_button {
+        display: none;
+    }
+}
+
+/* At twice min-width + aside_width, it's safe to make two columns */
+/* If min_width is 560px, $safe_two_column is 1280px */
+@media (min-width: $safe_two_column) {
```
yagebu: I'd prefer to see the desktop CSS code be the default and the changes for small devices scoped instead. It's not like small phones are our main target...

aumayr:
Apart from @yagebu's comments (which I think are accurate), I really like the responsive design. There will be bugs and refinements in the future, as some screens need special adaption, but overall I think this is good to merge (again, except @yagebu's comments)!

yagebu:
> Occasionally the hover event is fired on mobile, but it's very inconsistent, so tapping sometimes makes the menu flash and then disappear, only to flash and disappear the next time you tap it.

I don't have a device to test this on atm, but at least in the chrome inspector the current dropdown works seamlessly. I can't remember a lot of problems with dropdown menus on iPads, mostly they work quite well. In any case, I don't see any reason to change the behaviour on desktop browsers.

fava/templates/_layout.html
yagebu: I don't think the header should be fixed on mobile (it takes up much more space relatively). As a side effect, the stacking context wouldn't be a problem then. Or else the whole header can just be z-indexed above the sidebar, in which case this also isn't a problem. Then we can go back to the CSS dropdown as well.

xentac:
I have made all the changes you suggested and rebased onto current master so the merge should be seamless.

aumayr:
Looks very good, thanks @xentac! Here are some screenshots from the iOS-Simulator:

screenshot-1
screenshot-2
screenshot-3
screenshot-4

yagebu:
Looks good!

PR #322, opened 2016-06-07
Improve testing of parse_date()

yegle:
This will make the intention more explicit and easier to add more test
case.

### DIFF
--- a/fava/util/date.py
+++ b/fava/util/date.py
@@ -94,7 +94,7 @@ def _parse_month(month):
         return months_abbr.index(month)
 
 
-def parse_date(string):
+def parse_date(string, today=None):
     """"Tries to parse the given string into two date objects marking the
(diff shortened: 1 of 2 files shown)


fava/util/date.py
```diff
@@ -94,7 +94,7 @@ def _parse_month(month):
         return months_abbr.index(month)
 
 
-def parse_date(string):
+def parse_date(string, today=None):
```
yagebu: I'm against adding unnecessary arguments just for tests. Or is there some other place where we use it?

yegle: Hmm yes you are right, I should use mock instead. I'll send a PR later today to fix this. Sorry about that.

yagebu: No worries. Thank you for improving the tests!

Issue #313, opened 2016-06-02
Support for general relative dates in time filter


yegle:
I was trying to use something like `3 months ago - 2 months later` as a time filter but failed. This could be useful in additional to existing `last`, `this`, `next` modifier.

I can send a CL once I get a clean commit.

aumayr:
Will close this as the discussion happens in PR #326.

PR #326, opened 2016-06-08
Add a general way to specify relative date in date filter.

yegle:
Examples include:
- 10 days ago
- 1 quarter ago
- -10 weeks
- +1 year


yegle:
This should fix #313

yegle:
So the old time filter used to have placeholder to educate users about possible filter syntax, but it's no longer there.

I don't think people would find the feature in this PR unless they read the source code. How should we address this discover-ability issue?

aumayr:
@yegle Good point, I miss them, too! Maybe we could "manually" push some example strings to the list of values that are presented in this filter (the years). There would be no text explaining that these are examples, but there would be at least some hints as to how powerful this filter is.

yagebu:
> I don't think people would find the feature in this PR unless they read the source code. How should we address this discover-ability issue?

This is exactly what we have the help pages for.

fava/util/date.py
```diff
@@ -35,27 +35,72 @@
...
-def get_next_interval(date, interval):
+rel_date_re = re.compile(
+    # optional space, to support "10d ago" and "10 days ago"
+    r'(\d+)[ ]?'
+    # w, week weeks
+    # d, day, days
+    # q, quarter, quaters
+    # y, year, years
+    r'(d(?=ays?)?|w(?=eeks?)?|q(?=uarters?)?|y(?=ears?)?)[^ ]*'
+    r'[ ](ago|later)')
```
yagebu: "later" is not the correct word here. It's not an antonym of "ago", I think we should rather use "from now".

yagebu:
I don't like the syntax for these new relative intervals. Semantically, they seem wrong. "1 week ago" is a single point in time and shouldn't be an interval. Also, I'd expect it to be exactly 7 days ago, not the beginning of last week.

yegle:
@yagebu That's right, I think the syntax should be improved in some way.

My motive was to add support for "the month before last month", so I came up with either "2 months ago" or "-2 month". I'm not a native English speaker, but I find it hard to express it precisely.

Any suggestions?

yagebu:
I'm for making the syntax more formal. It doesn't have to be a shortened version of the English full text IMHO. It should just be logical and concise. Something like `year+4` maybe, or even just `y+4`. So `year`/`y` would just stand for the current year and it could be shifted. This is just off the top of my head and we would have to see how this fits with the rest of the syntax (I'm not against streamlining the whole date syntax, it's quite a mess atm).

yegle:
For specify a range (since in fava a single day is useless) precisely, we'll need a delta (number of days starting from today, positive or negative), and a span (the length of result range, align to the nearest interval).

When you say `year+4`, you mean the delta is +4 years, but you didn't specify the span. Does my search for "year+4" show transactions of Jun 2020, or transactions from Jan 1 to Dec 31 of 2020?

If we use the implied span, then the PR here already implement support for `+10 year` (delta: 10 year, span: year) and  `-3q` (delta: -3 quarter, span: quarter). Maybe remove support for ambiguous `10 years from now` and `3 quarters ago`. 

A more precise syntax would be `+4 year@month`, which means the delta to be +4 years and the span to be a month, and fava should show me transactions in Jun 2020.

yegle:
Okay here's what I have in mind:

1. remove “X unit ago/from now" syntax since it's not precise
2. support for "[+-]X unit" syntax, use the unit as the span
3. support for "[+-]X unit/span" syntax and use the explicit span

Let me know what you think.

yagebu:
> remove “X unit ago/from now" syntax since it's not precise

Yes, let's remove all the "alternative" date formats IMHO.

> support for "[+-]X unit" syntax, use the unit as the span

This doesn't seem to be the right syntax for me for an interval. Let me explain how I interpret it. We add some number `X` of `unit` to something, say `4` and `year` respectively. Since something is unspecified, taking "today" as the default for that seems sensible. But "today plus four years" is not a whole year in four years time, but rather a single day.

My proposition is declaring `year` and so on as variables for the current year etc. If this variable then supports addition, I think we have the required functionality and to me this is quite logical. For example `year+4` would be interpreted as `2016+4` or rather `2020`, which would then have an obvious interpretation by our simple original date syntax.

The only slight problem with this is making this really flexible. For example it would be ambiguous if `year-10-10` is October 10th of this year or rather all of October 10 years ago. But anyway, even if have to restrict somewhat, this syntax addition seems powerful enough.

My suggestion of course includes reworking the whole date parser, so it would be bigger than this PR and should possibly be discussed in a separate issue. So I'm currently against merging this PR (but we might be able to reuse your code later).

yegle:
I'm a bit confused. Let's consider the following scenario:

On Jun 09 2016, I'm viewing transactions of an account with time filter "last month - next month". I should be viewing the transactions between May 01, 2016 and Jul 31, 2016.

At this time, when I search for "year+4", am I viewing transactions between:
1. Jan 1, 2020 and Dec 31, 2020
2. May 1, 2020 and Jul 31, 2020
3. Something else
4. Consider the search invalid

Are you implying 2?

yagebu:
1., as explained `year+4` would be interpreted like `2020` which is all of 2020.

yegle:
Sorry for the delay. I've removed support for "X days ago" and only support `d+1` and `day+1` etc.

Please let me know what you think and we can continue improving on this.

yagebu:
I think this could actually be simpler if we implement it as a sort of pre-parser that replaces the variables with their values and do the arithmetic there as well. Then we don't have to work with dates, but just with numbers. I can do this and then I would also remove some of the current alternative syntaxes.

yegle:
Awesome, I'll wait for your PR then.

PR #334, opened 2016-06-14
Load settings from custom directives

aumayr:
This implements #214.

Settings from custom directives (`fava-config`) override the settings found in the default- and user-settings-file:

```ruby
# 2016-04-01 custom "fava-option" "[setting_name]" [setting_value]
2016-06-14 custom "fava-option" "language" "de"
2016-06-14 custom "fava-option" "theme" "alternative"
2016-04-14 custom "fava-option" "charts" FALSE
2016-04-14 custom "fava-option" "journal-show" "transaction open"
2016-04-14 custom "fava-option" "editor-print-margin-column" 10
```

As a result of using directives, settings are/can be beancount-file-specific, not global to fava. 

The current implementation is a kind of "proof-of-concept" if this is practical, until we have enough evidence to talk blais into updating the Beancount syntax to allow for custom option directives (currently only a predefined  set of Beancount options is allowed). 

To implement:

- [x] Tests
- [x] Documentation


fava/api/__init__.py
```diff
@@ -90,6 +90,19 @@ def _sidebar_links(entries):
...
 
 
+def _settings(entries):
+    """ Parse custom entries for fava settings. They have the following format:
+
+    2016-04-01 custom "fava-option" "[setting_name]" [setting_value]
+
+    """
+    settings_entries = [
+        entry for entry in _filter_entries_by_type(entries, Custom)
```
yagebu: Maybe filter out the custom entries in `BeancountReportAPI` and store them in `self.custom_entries` there. We now have three functions that only need custom entries, but run over all entries to filter them out.

fava/api/__init__.py
```diff
@@ -90,6 +90,19 @@ def _sidebar_links(entries):
             for entry in sidebar_link_entries]
 
 
+def _settings(entries):
```
yagebu: This function should also parse and validate the options (and generate beancount-errors if they're incorrectly specified)

yagebu:
I'm for making this change now and only supporting custom directives for settings (and not waiting for beancount to change). I think custom directives, although not perfect, are much better than the INI-style format as it's following the syntax that the user knows. If you're not against it @aumayr, I'll take over this PR.

This would work much better for `fava-electron`, which currently has no way to specify settings (and getting the current system to work would be a huge amount of work).

aumayr:
@yagebu Please, go ahead!

I thought that @blais would implement this change faster, so I did let this hang for a while…

PR #346, opened 2016-08-02
Editor: Add button (and keyboard shortcut) for invoking bean-format

aumayr:
This adds a `Format` button next to the `Save` button to run the current editor content through `bean-format`. The keyboard shortcut is `cmd+d`.  It does not save the file. 

bildschirmfoto 2016-08-02 um 11 17 04

Is the styling OK?


yagebu:
The styling is ok. If we add more buttons we might have to rethink the layout of the top bar, but so far I guess it's fine.

This currently introduces some breakage.

aumayr:
> This currently introduces some breakage.

Do you mean Travis CI?

yagebu:
> Do you mean Travis CI?

Yes. The scraper tries to access `/source/format/` which throws an error as it only allows `POST`. 

Maybe we should start grouping these interfaces for the JS-code under some common endpoint like `/api/` (saving/loading the source file could also go there).

aumayr:
I fixed the scraping-issues, but `flake8` is doing something strange: Checking all the pip-dependencies, and only in one of the two Travis CI runs (see above). It also does this locally when I run `tox`, but not when I run only `flake8`. @yagebu Do you have any idea why this is suddenly happening?

yagebu:
@aumayr: No idea why this is happening. Maybe the way that `tox` sets up the environment confuses `flake8`. Anyway I also noticed this (and added a workaround in d974a3da1cf170d0398f1c563a7693db0e216b35).

PR #356, opened 2016-08-27
Fix an issue that zero balance may not show on chart.

yegle:
When one of the commodities an account holds goes to zero, it will not
show in `balance` field in the corresponding entry. A previous attempt
to workaround the issue didn't consider when an account holds more than
two commodities. This change fixed the workaround. Test case has been
updated to cover it.


aumayr:
Thanks!

PR #358, opened 2016-09-03
Changes to support low-res modes again

xentac:
Now that some of the formatting has changed.  Here are a few patches that fix newly introduced problems.


aumayr:
Honestly I didn't test the changes lately on low-res, so thanks for the PR!

PR #360, opened 2016-09-11
ENH: Add ability to expand tree-table one level

davidastephens:
Adds feature suggested in #359


yagebu:
On macOS, Ctrl-Click works as a Right-Click, so it would be nice if Cmd-Click would work in that case.

PR #362, opened 2016-09-15
Fix highlighting of tags

aumayr:
The `beancount`-mode for codemirror does not match tags correctly, like this example:

```sql
2016-09-15 * "" "Test" #tag1
    Expenses:Test 10.00 USD
    Assets:Test 10.00 USD
```

### DIFF
--- a/fava/static/javascript/codemirror-mode-beancount.js
+++ b/fava/static/javascript/codemirror-mode-beancount.js
@@ -38,7 +38,11 @@ CodeMirror.defineSimpleMode('beancount', {
       token: 'whitespace',
     },
     {
-      regex: /[*!&?%PSTCURM]|txn/,
+      regex: /#[A-Za-z0-9\-_\/.]+/,
+      token: 'tag',
+    },
+    {
+      regex: /[*!&#?%PSTCURM]|txn/,
       token: 'directive transaction',
     },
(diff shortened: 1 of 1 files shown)


fava/static/javascript/codemirror-mode-beancount.js
```diff
@@ -38,7 +38,7 @@ CodeMirror.defineSimpleMode('beancount', {
       token: 'whitespace',
     },
     {
-      regex: /[*!&#?%PSTCURM]|txn/,
+      regex: /[*!&?%PSTCURM]|txn/,
```
aumayr: The `beancount`-mode for codemirror does not match tags correctly, like this example:

```
2016-09-15 * "" "Test" #tag1
    Expenses:Test 10.00 USD
    Assets:Test 10.00 USD
```

I found two ways to fix this:

1) Make the change in this line, not matching `#` as a part of the transaction directive, or 
2) moving the regex for matching tags above this line/rule. 

@yagebu What do you think is the best solution?

yagebu: I'd prefer 2). The regex above in this line is copied from beancount's `lexer.l` so changing it risks reintroducing this bug if one tries to match `lexer.l` in the future.

Issue #347, opened 2016-08-06
POST handlers pose a security risk


jbms:
I believe that the two POST handlers, for writing source files and uploading documents, pose a cross-site scripting security risk: if the user visits a malicious website while running fava, the malicious website could send a POST request and overwrite source files or upload documents.  The only information required is the path to the source file; while this might not be trivial to guess in some cases, it is hardly a sufficient security measure.

Any of a number of security measures designed for this purpose could probably be employed: e.g. a random token generated upon startup sent to the browser, or perhaps referer/origin checking would also be sufficient.

aumayr:
Thanks @jbms for mentioning this. 

The easiest way would be to include a CSRF-token (http://flask.pocoo.org/snippets/3/). 

@yagebu discussed (https://github.com/aumayr/fava/pull/346#issuecomment-2373338229 moving all the JS-endpoints to it's own sub-path (`/api` or something) - in this step we should also introduce a mechanism to fix this bug.

yagebu:
A CSRF-token is not really good for the JSON API as it would become invalid on a reload on changes to the source file.

yagebu:
For me the protection we have now (#367) is enough. It's seems like it's only exploitable with Flash and possible older browsers (as I expect newer versions to have addressed this problem).

See http://stackoverflow.com/a/22533680 for details.

If anyone can think of a painless way to add and Origin check, that would be even better...

aumayr:
@jbms Is #367 enough for you?

jbms:
I'm a bit concerned about it still being exploitable with Flash.  Perhaps the CSRF token is obtained via a regular GET XMLHttpRequest.  If any request that requires the CSRF token fails (presumably because the token was stale due to a server restart), a new token can be requested, and then the request retried.

yagebu:
> I'm a bit concerned about it still being exploitable with Flash.

Thanks for insisting, I guess you're right. It sucks that we have to work around a browser bug/vulnerability...

Turns out that adding an origin check is not so bad, so I'll do that.

jbms:
Can X-Forwarded-Host not be set from vulnerable Flash in the same way that X-Requested-With can be set?

PR #367, opened 2016-09-18
JS api: move things for consistency; basic csrf protection

yagebu:
Checking for the `XMLHttpRequest` in the header should safeguard us against all CSRF attacks possible in Javascript. An attack using flash might still be possible, against which we could check the origin in the request header. However this would complicates setting up fava behind a reverse proxy... Ref #347

@aumayr - a side note: Could you add documentation for documents upload? It's currently a completely hidden feature.

### DIFF
--- a/fava/application.py
+++ b/fava/application.py
@@ -1,9 +1,9 @@
 # -*- coding: utf-8 -*-
+import datetime
 import os
-from datetime import datetime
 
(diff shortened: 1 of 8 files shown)


aumayr:
Will do 👍🏻

aumayr:
@yagebu See 7b1b689 for the help section about the documents upload.

Issue #332, opened 2016-06-13
Add Docker images


aumayr:
(This is the successor to #314)

@yegle has made a Docker image for fava: https://hub.docker.com/r/yegle/fava/

@xentac already provided example-files how to use OAuth with nginx: 

> https://github.com/xentac/beancount-oauth2
> 
> Here is a skeleton config showing what I had to do.  It doesn't support the url limiting.  To do that, you need to change the nginx config like they say here: https://github.com/bitly/oauth2_proxy/#configuring-for-use-with-the-nginx-auth_request-directive

I propose the following:

- Add two Dockerfiles/images to the `contrib/docker`-folder: One with the latest stable version of fava, and one to build fava from source
- Add configuration (via env-variables) to enable/disable OAuth

Unfortunately, I do not have any experience with Docker, so any contributions are welcome. 

@yegle Can we use your images? Are you interested in this?

xentac:
I have a Dockerfile that clones beancount and fava, builds them, and installs the vim module and a few other things.  I use it as the base for my hosted solution and to enter all my transactions.

I wouldn't call it particularly awesome, but if you wanted to go that route, I'd be willing to contribute that also.  It's not exactly appropriate for an entirely static hosted solution, yegle's image is better for that.

aumayr:
Sounds good! Maybe wait with any work until @yegle comments as well. 

As said, I think it would be awesome to have two images: One for the latest stable release, and one that builds from source (as yegle's one does).

yegle:
Sure, feel free to use it. https://github.com/yegle/fava-docker it's licensed under MIT.

Since docker images are using the user's ID as namespace (so my fava image would be yegle/fava), @aumayr do you want to register a separate account to host the "official“ images that are built from this github repo?

As an alternative I can also modify my docker settings to use the Dockerfile in this repo to build the image.

aumayr:
I created https://hub.docker.com/r/fava/fava/ as a namespace. Can I add you @yegle as a team member to push the images? 

Is anyone else interested in docker-hub-team-membership?

yagebu:
Is this going anywhere?

adamgibbins:
I'd be happy to take this on if nobody else is interested. I use @yegle's on a regular basis, if there's interest in making that official it'd be nice to just shift it over to the fava namespace instead of duplicating, if @yegle is happy with that.

yegle:
I'm all fine with the changes.

yagebu:
[...]

PR #378, opened 2016-10-09
Add Dockerfile

adamgibbins:
This adds @yegle's Dockerfile to the contrib directory, preserving all history.
Once merged can you please configure an automated hub build in Dockerhub? I'm unable to do this without contrib rights on the repository oddly. This would mean a new image is built on every commit without someone needing to manually trigger.

1. Head to https://hub.docker.com/u/fava/dashboard/
2. Create (top left) -> Create Automated Build
3. GitHub
4. Click the repository
5. "Click here to customize"
6. Set the Dockerfile location

### DIFF
--- a/contrib/docker/Dockerfile
+++ b/contrib/docker/Dockerfile
@@ -13,8 +13,9 @@ RUN cd /root \
         && python3 -mpip install ./beancount \
         && git clone https://github.com/aumayr/fava.git \
         && (cd fava && git log -1) \
-        && make -C fava build-js \
-        && rm -rf fava/fava/static/node_modules \
+        && make -C fava \
+        && make -C fava clean \
+        && rm fava/CHANGES \
         && python3 -mpip install ./fava \
         && python3 -mpip uninstall --yes pip \
         && find /usr/local/lib/python3.5/site-packages -name *.so -print0|xargs -0 strip -v \
--- a/contrib/docker/Dockerfile
+++ b/contrib/docker/Dockerfile
@@ -1,0 +1,30 @@
+FROM python:3.5.1-alpine
+
+ENV BEANCOUNT_INPUT_FILE ""
+ENV FAVA_OPTIONS "-H 0.0.0.0"
+ENV FINGERPRINT "3f:d3:c5:17:23:3c:cd:f5:2d:17:76:06:93:7e:ee:97:42:21:14:aa"
+ENV BUILDDEPS "libxml2-dev libxslt-dev gcc musl-dev mercurial git nodejs make g++"
+ENV RUNDEPS "libxml2 libxslt"
+
+RUN cd /root \
+        && apk add --update $BUILDDEPS $RUNDEPS \
+        && hg clone --config hostfingerprints.bitbucket.org=$FINGERPRINT https://bitbucket.org/blais/beancount \
+        && (cd beancount && hg log -l1) \
+        && python3 -mpip install ./beancount \
+        && git clone https://github.com/aumayr/fava.git \
+        && (cd fava && git log -1) \
+        && make -C fava build-js \
+        && rm -rf fava/fava/static/node_modules \
+        && python3 -mpip install ./fava \
+        && python3 -mpip uninstall --yes pip \
+        && find /usr/local/lib/python3.5/site-packages -name *.so -print0|xargs -0 strip -v \
+        && apk del $BUILDDEPS \
+        && rm -rf /var/cache/apk /tmp /root \
+        && find /usr/local/lib/python3.5 -name __pycache__ -print0|xargs -0 rm -rf \
+        && find /usr/local/lib/python3.5 -name *.pyc -print0|xargs -0 rm -f
+
+
+# Default fava port number
+EXPOSE 5000
+
+CMD fava $FAVA_OPTIONS $BEANCOUNT_INPUT_FILE


yagebu:
Can you squash the commits into one?

adamgibbins:
Sure - done. Unfortunately cannot retain the original credit doing that, hence I opted to retain originally.

yagebu:
> Unfortunately cannot retain the original credit doing that, hence I opted to retain originally.

You can actually, I usually use `rebase -i` to do the squashing which will retain the original commit author and just add you as the committer.

contrib/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+
+RUN cd /root \
+        && apk add --update $BUILDDEPS $RUNDEPS \
+        && hg clone --config hostfingerprints.bitbucket.org=$FINGERPRINT https://bitbucket.org/blais/beancount \
+        && (cd beancount && hg log -l1) \
+        && python3 -mpip install ./beancount \
+        && git clone https://github.com/aumayr/fava.git \
+        && (cd fava && git log -1) \
+        && make -C fava build-js \
+        && rm -rf fava/fava/static/node_modules \
```
yagebu: better use make clean here

contrib/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+ENV RUNDEPS "libxml2 libxslt"
+
+RUN cd /root \
+        && apk add --update $BUILDDEPS $RUNDEPS \
+        && hg clone --config hostfingerprints.bitbucket.org=$FINGERPRINT https://bitbucket.org/blais/beancount \
+        && (cd beancount && hg log -l1) \
+        && python3 -mpip install ./beancount \
+        && git clone https://github.com/aumayr/fava.git \
+        && (cd fava && git log -1) \
+        && make -C fava build-js \
```
yagebu: the `build-js` can be removed

adamgibbins: Hm? I can't build it without that:
```
Exception:
  File "/usr/local/lib/python3.5/site-packages/pip/basecommand.py", line 215, in main
    status = self.run(options, args)
  File "/usr/local/lib/python3.5/site-packages/pip/commands/install.py", line 299, in run
    requirement_set.prepare_files(finder)
  File "/usr/local/lib/python3.5/site-packages/pip/req/req_set.py", line 370, in prepare_files
    ignore_dependencies=self.ignore_dependencies))
  File "/usr/local/lib/python3.5/site-packages/pip/req/req_set.py", line 587, in _prepare_file
    session=self.session, hashes=hashes)
  File "/usr/local/lib/python3.5/site-packages/pip/download.py", line 798, in unpack_url
    unpack_file_url(link, location, download_dir, hashes=hashes)
  File "/usr/local/lib/python3.5/site-packages/pip/download.py", line 676, in unpack_file_url
    shutil.copytree(link_path, location, symlinks=True)
  File "/usr/local/lib/python3.5/shutil.py", line 353, in copytree
    raise Error(errors)
shutil.Error: [('/root/fava/CHANGES', '/tmp/pip-y5627i91-build/CHANGES', "[Errno 95] Not supported: '/tmp/pip-y5627i91-build/CHANGES'")]
```

adamgibbins: Oh maybe not, apologies - seems something in the pipeline has started flickering at the most inconvenient time, sigh.

adamgibbins: @aumayr sorry for the delayed turn around on this. I'm having a trouble building recent versions of fava with `python3 -mpip install ./fava` - I get the above error, my python knowledge is near non-existant so it's highly possible I'm missing something obvious.

It looks like it's attempting to process the CHANGES file for some odd reason? Do pip packages have associated changelogs? Either way, it's not happy with it - any guesses as to why? Irritatingly I'm not able to reproduce this outside of the alpine container - it works fine on my box.

yegle: I was having the same issue here when trying to update yegle/fava docker image. I'm familiar with Python but not very much with Docker, so I'm not sure what happened here.

A quick&dirty solution would be delete the CHANGES file before `python3 -mpip install ./fava`, but I'm curious to know why it fail here.

adamgibbins: @yegle Yeah I'd considered doing that but thought I'd raise wider to see if anyone had any ideas first, as it is a hack. I note CHANGES in the root a symlink that was put there on the 11th, the same day I noticed the issues (I'm certain it worked prior) -- I'm guessing for some reason something in the alpine/pip/docker setup doesn't like the symlink although I haven't a clue to why.

yagebu: The symlink seems to cause problems with ancient versions of `pip`, which version are you using?

`build-js` is part of the default build target, so `make -C fava` is enough

adamgibbins: The upstream image uses pip 7.1.2, latest seems to be 8.1.2, maybe that's the issue.

re build-js ah, sorry misinterpreted your original comment.

yegle: https://github.com/pypa/pip/issues/1006 It seems to be fixed long ago, not sure why it resurfaced now.

adamgibbins:
Alright, force pushed CHANGES workaround and suggested changes.

yagebu:
> Once merged can you please configure an automated hub build in Dockerhub?

I'm against this as long as the Dockerfile installs from master. We don't want to distribute that.

adamgibbins:
@yagebu even if you tag it as unstable?  IMO the big advantage to getting it within the official fava repo was so we could do this.  Without this people have to maintain their own forks and manually hit build every time fava changes -- there's no way of making it automated through the hub. This means you end up with binary blob images that aren't traced back to dockerfiles, so there's no way to verify the integrity.  This was the problem I had with yegle's, I'd have to poke him to hit build on the hub, use a stale HEAD or push to my own hub user and build it there.

Would you be happy with this if we had one that tracked the stable releases too?

aumayr:
I'm for maintaining/building two versions:

- A stable one for each release, and tagged as such (as many other Docker images do, with `-latest` and `-1.0.0`, etc.).
- An unstable one, tagged as such, for the exact reasons @adamgibbins gives.

yagebu:
Can someone address my concerns that I already posed in #332?:

> The dockerfile only installs Fava and doesn't really do anything that running Fava from the command line can't do, right? I've not used Docker before so I might not be aware of some of the advantages of it in this use-case, but it seems to me that it just adds a layer of complexity. How is it easier than running pip3 install beancount-fava?

Of all the ways of distributing an end-user command-line executable, Docker just seems to be the shittiest one.

yagebu:
One should only have to install from master to code or help debugging, which is impossible with the Dockerfile. We had too long gaps between releases so it took too much time for features to land in the released versions, but since 1.0 I plan to something like monthly releases, so this shouldn't be a problem anymore.

aumayr:
> How is it easier than running `pip3 install beancount-fava`?

To avoid problems with dependencies there are AFAIK two solutions:
- Use `virtualenv`
- Use Docker

When I do have to install it on machines that are not dev-machines (like my mom's laptop) it is really a PITA to install all dependencies: Get Python 3 onto that machine and then all the dependencies (I don't think Xcode is necessary anymore for building/installing beancount).

With Docker this get's way easier: Download Docker, and then just get the Fava image and set Docker to start on startup. 

That is also the reason I started to wrap Fava in Electron => to make it easier to install and more accessible.

I agree that this might not be necessary for an unstable version (as the described use case is just relevant for stable releases), but I guess there are people who just want to try out Fava or something new we develop on master to join the discussion, without the hassle of setting up a proper development environment.

aumayr:
> We had too long gaps between releases so it took too much time for features to land in the released versions, but since 1.0 I plan to something like monthly releases, so this shouldn't be a problem anymore.

🎉

yagebu:
You have some good points. I still think that using Docker is the wrong solution to these problems. PyInstaller offers us a way to build a single executable that users then just have to run, I find that much superior to Docker. If we want to provide a simpler way to install it, we should setup a way to build those, not use Dockerfile as a half-hearted stopgap measure.

aumayr:
> PyInstaller offers us a way to build a single executable that users then just have to run, I find that much superior to Docker.

Agreed! Do you have ideas how to automate building them?

aumayr:
(Aside: The absolute dream-setup would be combining PyInstaller with Electron => Fava could be a simple `.app`/`.exe` that normal users can download and run, without worrying/knowing about Python/pip/dependencies/etc.)

yagebu:
> Agreed! Do you have ideas how to automate building them?

Not sure how to best do it. [borgbackup/borg](https://github.com/borgbackup/borg) offers PyInstaller binaries and uses Vagrant to build them. Alternatively, we could build them on Travis.

yagebu:
> (Aside: The absolute dream-setup would be combining PyInstaller with Electron => Fava could be a simple .app/.exe that normal users can download and run, without worrying/knowing about Python/pip/dependencies/etc.)

fava-electron does already use PyInstaller ;) And electron-builder does a great job of building an installer, but similar to here, we have to find a good way to build cross-platform... (and make fava-electron just slightly more usable)

yegle:
(Chime in to offer my opinion regarding docker vs package fava into an app)

I started dockerize fava due to various reasons. First, it used to be the case that beancount have new features in tip but not on PyPI, and beancount from tip might be incompatible with fava from PyPI (I forgot if I did encounter an incompatibility though, it was long ago). Second, I need to share fava UI with my wife while I'm the one adding transactions to beancount file, with a NAS at home I can host fava in docker easily.

With docker, I edit the bean file on my laptop and then use Dropbox to sync it to my NAS.  My wife can then open fava on the NAS to view all my changes in near realtime.

Using a tool to package fava into an executable doesn't seem to solve my problem, I still need to find a way to sync the bean file to my wife's laptop.

yagebu:
> Using a tool to package fava into an executable doesn't seem to solve my problem, I still need to find a way to sync the bean file to my wife's laptop.

How does Docker help with this? I thought you use Dropbox for the syncing, why wouldn't that work with a Fava executable?

yegle:
@yagebu Syncing a set of files using Dropbox to multiple devices to be used by fava binary is not easy unless every device uses the same root directory, which is impossible with different OS (Windows/macOS).

Hosting it on a server in a docker also has the advantage that I can access via my tablet/phone etc.

yagebu:
> Syncing a set of files using Dropbox to multiple devices to be used by fava binary is not easy unless every device uses the same root directory, which is impossible with different OS (Windows/macOS).

I don't quite understand, if you start the binary on a different machine, you can specify a different path to the file.

> Hosting it on a server in a docker also has the advantage that I can access via my tablet/phone etc.

Ok but than that's quite a different use-case from the one that this Dockerfile here tries to solve.

aumayr:
For me, having fava in Docker allows me to easily add HTTPS and Auth to it, so this is really helpful for a more "complex" setup.

PR #393, opened 2016-10-30
Document upload: Invoices

aumayr:
Related issue with discussion: #386

To try out the plugin add this to your beancount file:

```ruby
plugin "fava.plugins.statements_from_metadata"

2016-10-30 * "" "Test"
  statement: "2016-10-30 test.pdf"
  Expenses:Other                                             10.00 EUR
  Assets:Cash
    statement: "2016-10-30 test 2.pdf"
```

Things to do related to the plugin:

- [ ] Link the transaction and the generated Document entry
- [x] Support multiple statements per transaction (currently metadata is a dict), like
  `statements `, `statements-2`, `statements-3`, ...
- [ ] Add description of the feature to help pages, including which paths are searched
- [ ] Testcases

General things to do:

- [ ] Support "relative" paths suggested by @corani (https://github.com/beancount/fava/issues/386#issuecomment-256267212) when downloading files (in the Journal)
- [ ] Drag'n'drop: Add support for multiple `document_root`
- [ ] Drag'n'drop: Add support for dropping on a Journal transaction or posting (upload; add the corresponding metadata entry)

### DIFF
--- a/contrib/docker/Dockerfile
+++ b/contrib/docker/Dockerfile
@@ -1,0 +1,31 @@
+FROM python:3.5.1-alpine
+
+ENV BEANCOUNT_INPUT_FILE ""
+ENV FAVA_OPTIONS "-H 0.0.0.0"
+ENV FINGERPRINT "3f:d3:c5:17:23:3c:cd:f5:2d:17:76:06:93:7e:ee:97:42:21:14:aa"
+ENV BUILDDEPS "libxml2-dev libxslt-dev gcc musl-dev mercurial git nodejs make g++"
+ENV RUNDEPS "libxml2 libxslt"
+
+RUN cd /root \
+        && apk add --update $BUILDDEPS $RUNDEPS \
+        && hg clone --config hostfingerprints.bitbucket.org=$FINGERPRINT https://bitbucket.org/blais/beancount \
+        && (cd beancount && hg log -l1) \
+        && python3 -mpip install ./beancount \
+        && git clone https://github.com/aumayr/fava.git \
+        && (cd fava && git log -1) \
+        && make -C fava \
+        && make -C fava clean \
+        && rm fava/CHANGES \
+        && python3 -mpip install ./fava \
+        && python3 -mpip uninstall --yes pip \
+        && find /usr/local/lib/python3.5/site-packages -name *.so -print0|xargs -0 strip -v \
+        && apk del $BUILDDEPS \
+        && rm -rf /var/cache/apk /tmp /root \
+        && find /usr/local/lib/python3.5 -name __pycache__ -print0|xargs -0 rm -rf \
+        && find /usr/local/lib/python3.5 -name *.pyc -print0|xargs -0 rm -f
+
+
+# Default fava port number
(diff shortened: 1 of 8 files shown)


yagebu:
I think the file-matching should be quite strict. It should only match if it's actually in the folder of one of the accounts of the transaction. Also, all filenames for documents have to begin with a date, so for the example transaction, there should only be two possibilities where the file could be. Additionally, I think it's a bit odd, if the filename in the metadata is `test.pdf` if this can't be the actual filename.

Can you rebase this onto master?

> Make the metadata-key to be used configureable

Please no.

aumayr:
> I think the file-matching should be quite strict. 

After thinking about it, you are right. Will change it as suggested.

> Can you rebase this onto master?

As always with git, I did something but don't know if I got it right. Does it look like I rebased successfully?

> > Make the metadata-key to be used configureable
> 
> Please no.

OK

yagebu:
> As always with git, I did something but don't know if I got it right. Does it look like I rebased successfully?

It looks even weirder than before :D Do you remember which commands you ran? If I just checkout the `invoices` branch and run `git rebase master` that seems to do the wanted rebase.

I always try to keep all branches I'm working on close to master to keep merging troubles to a minimum. So usually running `git rebase master -i` (while on the branch) should show only the "real" commits of the branch and allow to squash/rebase them on master (i.e. apply all these commits after the latest commit on master).

aumayr:
> It looks even weirder than before :D 

👼🏼

> Do you remember which commands you ran?

I tried several in succession: 

```bash
$ git rebase origin/master
$ git pull --rebase
$ git rebase -i master
```

Can I/should I somehow "repair" this?

yagebu:
> Can I/should I somehow "repair" this?

If you run something like `git rebase master`, always make sure that `master` is up to date first, otherwise you might just be wasting effort on rebasing onto some outdated local version. The following should repair it:

    git checkout master
    git pull
    git checkout invoices
    git rebase master

aumayr:
Regarding filenames/fuzzy-matching:

> Also, all filenames for documents have to begin with a date, so for the example transaction, there should only be two possibilities where the file could be. Additionally, I think it's a bit odd, if the filename in the metadata is `test.pdf` if this can't be the actual filename.

Actually, naming files without a date is possible (and working). The date-part is only necessary for the auto-discovery-by-beancount feature. Even the official docs contain an example without the date-part:

https://docs.google.com/document/d/1wAMVrKIA2qtRGmoVDSUBJGmYZSygUaR0uOMW1GV3YE0/edit#heading=h.w1ins9jk4mq3

yagebu:
> Actually, naming files without a date is possible (and working). The date-part is only necessary for the auto-discovery-by-beancount feature. Even the official docs contain an example without the date-part.

Yes, but to me this plugin only really makes sense with the auto-generated document entries. In any case, I think the metadata should contain the full filename.

aumayr:
I completely destroyed this branch when trying to rebase it, so I created a fresh branch/PR in #398. Please head over there to continue discussing this feature.

Issue #388, opened 2016-10-24
New fetch API doesn't send cookies by default


xentac:
As I commented in https://github.com/aumayr/fava/commit/b026ebc5a6f2af52b3fb869f71aad10be123b484, it looks like fetch doesn't send browser cookies by default.  This breaks anything that depends on oauth proxy or any authentication at all.

xentac:
New PR https://github.com/beancount/fava/pull/408

PR #397, opened 2016-10-30
Add credentials: 'include' to all fetch calls

xentac:
Fixes https://github.com/beancount/fava/issues/388.


aumayr:
Thanks @xentac!

yagebu:
Would be nicer to handle this in one place...

yagebu:
A helper function simply called `fetch` in `helpers.js` would be best IMHO. Also, it should set `credential: 'same-origin'`. I'll revert this PR, @xentac could you provide a PR updated with my suggestions?

xentac:
Yeah, I'll write one when I get a chance.

PR #398, opened 2016-10-30
Document upload: Invoices (2)

aumayr:
_I completely destroyed #393 when trying to rebase it, so I created a fresh branch to continue working. There are some important discussion comments there._

Related issue with discussion: #386

To try out the plugin add this to your beancount file:

``` ruby
plugin "fava.plugins.statements_from_metadata"

2016-10-30 * "" "Test"
  statement: "2016-10-30 test.pdf"
  Expenses:Other                                             10.00 EUR
  Assets:Cash
    statement: "2016-10-30 test 2.pdf"
```

Things to do related to the plugin:
- [x] Link the transaction and the generated Document entry
- [x] Support multiple statements per transaction (currently metadata is a dict), like
  `statements`, `statements-2`, `statements-3`, ...
- [x] Add description of the feature to help pages, including which paths are searched
- [x] Testcases

General things to do:
- [x] Support "relative" paths suggested by @corani (https://github.com/beancount/fava/issues/386#issuecomment-256267212) when downloading files (in the Journal)
- [x] Drag'n'drop: Add support for multiple `document_root`
- [x] Drag'n'drop: Add support for dropping on a Journal transaction or posting (upload; add the corresponding metadata entry)


aumayr:
OK now comes the IMHO hardest part: When draganddropping a file onto a Journal row (Transaction or Posting) the file should be uploaded (easy) and a corresponding metadata-entry inserted into the beancount file (hard).

The last part is somewhat scary, as we programmatically change the beancount file.

@yagebu Do you have any thoughts on how to best do that?

fava/application.py
```diff
@@ -325,6 +326,17 @@ def api_add_document():
...
+        if request.form.get('bfilename', None):
+            # TODO use next_statement_key to calculate the metadata key to
+            # support multiple statement keys
+            relpath = os.path.relpath(filepath, os.path.dirname(request.form['bfilename']))
+            insert_line_in_file(
+                request.form['bfilename'],
+                int(request.form['blineno'])-1,
+                'statement: "{}"'.format(relpath))
+
+        g.api.changed(force=True)
```
aumayr: @yagebu The change in the beancount file/documents folder is not picked up when uploading a file, so we have to set the changed-flag manually here. Does this sound like a bug to you?

fava/application.py
```diff
@@ -325,6 +326,17 @@ def api_add_document():
             return api_error('{} already exists.'.format(filepath))
 
         file.save(filepath)
+
+        if request.form.get('bfilename', None):
+            # TODO use next_statement_key to calculate the metadata key to
```
aumayr: I wrote a helper-function in `fava.util.__init__` to get the next key from an array of metadata-keys, so if `statement:` is already present it will return `statement-2` and so forth. 

At this point in the code we only know which filename and lineno this metadata key should go to, but nothing about the entry itself, eg. the existing metadata-keys for it. 

@yagebu Do you have an idea how we could get an array of the metadata-keys for that entry at this point in the code?

fava/application.py
```diff
@@ -325,6 +326,17 @@ def api_add_document():
...
+            # support multiple statement keys
```
aumayr: I wrote a helper-function in `fava.util.__init__` to get the next key from an array of metadata-keys, so if `statement:` is already present it will return `statement-2` and so forth. 

At this point in the code we only know which filename and lineno this metadata key should go to, but nothing about the entry itself, eg. the existing metadata-keys for it. 

@yagebu Do you have an idea how we could get an array of the metadata-keys for that entry at this point in the code?

fava/application.py
aumayr: And there is a general UI bug here as well: When the Journal get's refreshed subsequently, all filters (via GET-params, like "show metadata") are cleared, which is not what the user intended.

aumayr:
c0e1c5d does what this PR intends to do, except a few rough edges and bugs (see comments in the code of the PR itself).

There is an usability issue:

When uploading a file by dropping on a Journal row (Transaction or Posting), the file is uploaded correctly and the metadata entry is also correctly inserted where it should go. But as the files are in one of the specified `documents`-folders they will show up as Document directives on their own anyway, which is not what I intended, as they cannot be hidden without hiding all other "standalone" Document entries as well, and are not linked to the corresponding transaction (what the plugin this PR contains would do). 

So basically I want two things:
- Beancount shouldn't generate Document directives when the filepath is found in a `statement:` metadata key.
- If I want to see the statements as Document directives as well I can enable the plugin, which also links the Document directive to the Transaction where the statement was found.

fava/api/__init__.py
```diff
@@ -173,12 +173,12 @@ def filter(self, **kwargs):
     def hash_entry(self, entry):
         return compare.hash_entry(entry)
 
-    def changed(self):
+    def changed(self, force=False):
```
yagebu: revert

fava/api/__init__.py
```diff
@@ -173,12 +173,12 @@ def filter(self, **kwargs):
...
         """Check if the file needs to be reloaded. """
         # We can't reload an encrypted file, so act like it never changes.
         if self.is_encrypted:
             return False
-        changed = self.watcher.check()
+        changed = True if force else self.watcher.check()
```
yagebu: revert

fava/application.py
yagebu: revert

yagebu: If the change is not picked up, that's a bug either because the watcher doesn't work correctly or that maybe the reloading is called too early from Javascript. Let's not hide bugs with hacks.

fava/plugins/documents_from_metadata.py
```diff
@@ -0,0 +1,87 @@
...
+
+def statements_from_metadata(entries, options_map):
+    errors = []
+
+    if 'documents' not in options_map or len(options_map['documents']) == 0:
+        return entries, errors
+
+    def filenames(date, name):
+        d_str = date.strftime("%Y-%m-%d")
+        return [name, '{} {}'.format(d_str, name), '{}.{}'.format(d_str, name)]
```
yagebu: I think the matching should be strict, only the given name should match.

fava/plugins/documents_from_metadata.py
```diff
@@ -0,0 +1,87 @@
...
+
+    for i, entry in enumerate(entries):
+        type = entry.__class__.__name__.lower()
```
yagebu: Never do this in Python, use isinstance checks

fava/plugins/documents_from_metadata.py
```diff
@@ -0,0 +1,87 @@
...
+            continue
+
+        paths = []  # set of tuples (account, path)
+        b_dir = dirname(options_map['filename'])
+
+        for documents in options_map['documents']:
+            statements = []
+
+            # If the `statement` metadata key is set on the transaction, assume
+            # it belongs to the first posting
```
yagebu: This is surprising behaviour to me. The "fuzzy" part should cover this and find the right posting IMHO.

fava/application.py
yagebu: We can just find the corresponding entry. Also this Beancount-file related logic _really_ shouldn't be in `application`

yagebu:
Mhh I don't really like that this plugin would duplicate functionality that's already in beancount (namely generating Document entries), which seems quite unnecessary and seems like something that will lead to tons of bugs, like duplicate entries.

What I think this plugin should do: Find all statement metadata and link it to the document entries (and add a `#statement` tag so that we can filter it). The Document directives should really be generated by Beancount. The way it's currently 

Also it would be neater if it supported the fuzzy matching suggested by @corani in #386 (minus prepending the date to the filename) and not assume that the statement belongs to the first posting.

> If I want to see the statements as Document directives as well I can enable the plugin, which also links the Document directive to the Transaction where the statement was found.

Enabling/disabling a plugin is bad from the UI perspective to change what is displayed in the Journal IMHO. If we add a `#statement` tag to the Document directive, we can do it much more elegantly.

fava/api/__init__.py
aumayr: reverted

fava/api/__init__.py
aumayr: reverted

aumayr:
OK so I created two new plugins that should replace the first one:

```ruby
# Tags beancount-discovered Document directives with #discovered
plugin "fava.plugins.tag_discovered_documents"

# Tags Document directives that correspond to statement-metadata-entries with #statement and link them to the Transaction
plugin "fava.plugins.tag_statement_documents"
```

(I did not remove the code for `documents_from_metadata.py` yet, but will when there is really no need for it, before the PR is merged or even earlier.)

What is interesting is that the two tags that are added by the plugin, `#discovered` and `#statement`, are not picked up by the Tags-filter!?

The only "ambiguous" place where I don't know what to do is when a user drops a document directly on a Transaction, and not on one of it's Postings: Where should the file go? Right now it goes to the first Posting, eg. the file will be stored in the folder corresponding to the account of the first Posting, but the metadata entry will be set on the Transaction as a whole.

What is still open:

- [ ] Moving the beancount-file-modification-specific code from `application` somewhere else. @yagebu Where should this go?
- [ ] Bug regarding reload after dropping on the Journal page
- [ ] Bug when dragging on a Account page where a Journal table is shown and therefore two droptargets are ontop eachother
- [ ] Support for generating multiple statement-metadata-entries

aumayr:
> How about calling it `link_statements` as this seems to be the main thing it does, while the tagging is secondary.

That's a way better name - thanks!

> > What is interesting is that the two tags that are added by the plugin, #discovered and #statement, are not picked up by the Tags-filter!?
> 
> That's odd. But they show up in the journal?

They certainly do.

> The documents upload could/should have a field for the account. I'd find it nice if this always shows up as it would provide a way to confirm (for all document uploads) that I'm uploading a document to the right account, which is not possible currently.

If it is unambiguous where the document should go (eg. dropped on an Account, and only one document root specified, and a date-prefix in the filename) I do not really see the need for the confirmation. Maybe we can do this like many Mac applications work: When holding the option-key during dropping the file the dialog/confirmation will always be shown (This is possible in JavaScript, see http://stackoverflow.com/a/8875522 for example). 
 
> > Moving the beancount-file-modification-specific code from application somewhere else. 
> 
> `api/file.py` I'd suggest.

👍

yagebu:
> If it is unambiguous where the document should go (eg. dropped on an Account, and only one document root specified, and a date-prefix in the filename) I do not really see the need for the confirmation.

Yes, I meant that it always shows the input when the form shows up.

aumayr:
> Yes, I meant that it always shows the input when the form shows up.

Oh, now I got it: You mean the Account field, right? 

That should be easy to implement, as well as the holding-the-option-key-feature I described.

yagebu:
> Oh, now I got it: You mean the Account field, right?

Yes :)

aumayr:
Now that we have tags for `#discovered` and `#statement`, do you have ideas how (UI/UX-wise) we could add filters and options to the Journal to hide these Document entries?

yagebu:
> Now that we have tags for #discovered and #statement, do you have ideas how (UI/UX-wise) we could add filters and options to the Journal to hide these Document entries?

Maybe do something similar to the flag filters for transactions?

aumayr:
Now I think the functionality of this PR is complete:

- Drop a file on a Journal row (Transaction or Posting)
- Document get's uploaded
- Metadata-entry is added to the beancount file
- Document entries get tagged with `#statement` (via beancount-plugin)
- Filter in Journal to sort out `#statement`-tagged Document entries

Additional features:

- Tag beancount-auto-discovered Documents with `#discovered` (via beancount-plugin)
- Filter in Journal to sort out `#discovered `-tagged Document entries
- Some bugfixes for other issues

Still open: 

- [ ] Help
- [ ] Support for "relative" paths in `statement`-metadata suggested by @corani (https://github.com/beancount/fava/issues/386#issuecomment-256267212) and insert Document directives if found

Future features somewhat related to this PR but should be implemented outside of this PR:

- Always show the "Account" field (dropdown?) in the document upload dialog (upon ambiguity)
- Holding-the-option-key-feature when dropping a file to always show the upload dialog (even if not ambiguous)

aumayr:
Should we hide the filters for `#discovered` and `#statement` if the corresponding plugins are not enabled in the Beancount file?

(Also: Hide/disable keyboard shortcuts)

yagebu:
Nice work overall! I'll add some comments to the code, here are some more general remarks:

- We allow posting-metadata, but don't use any information about the posting then. We should only allow transaction-level metadata, let's not create more places for bugs to occur.
- The plugins needs test. We should strive for ~100% coverage in the code we add.
- I'd really like to see filename matching in the first version. I'd not use the statements if they have to contain relative or absolute paths (absolute and relative should still be possible though). How I think this should work:
 - Collect all statement paths (which is absolute, relative or just the filename)
 - Collect all accounts for the transaction
 - Match
    - if the complete path matches (would work for absolute and relative paths)
    - or if the filename (not path) matches and the account of the document in question is contained in the accounts

> Should we hide the filters for #discovered and #statement if the corresponding plugins are not enabled in the Beancount file?

No, the amount of `if`ed out code in Fava is already causing problems IMHO.

fava/api/__init__.py
```diff
@@ -365,16 +366,24 @@ def is_valid_document(self, file_path):
         """Check if file_path is the path to a document.
 
         It should either occur in one of the Document entries or in a
-        "statement"-metadata in a Transaction entry.
+        "statement"-metadata in a Transaction entry or one of it's Postings.
```
yagebu: We should restrict to documents. For each statement, there should exist a (generated) document entry.

fava/api/__init__.py
```diff
@@ -424,3 +433,29 @@ def account_metadata(self, account_name):
...
+                   entry.meta['lineno'] == lineno:
+                    key = next_key(basekey, entry.meta)
+                    break
+                for pos in entry.postings:
+                    if pos.meta and pos.meta['filename'] == filepath and \
+                       pos.meta['lineno'] == lineno:
+                        key = next_key(basekey, pos.meta if pos.meta else [])
+                        break
+
+        assert key != None
```
yagebu: lint

yagebu: Actually this should raise an Exception that is properly handled in the application

fava/api/file.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+def insert_line_in_file(filename, lineno, content):
+    """Inserts the specified content in the file below lineno, taking into
+    account the whitespace in front of the line that lineno."""
+    with open(filename, "r") as file:
+        contents = file.readlines()
+
+    # this is beancount-specific: if the line has no whitespace in front of it,
+    # use the whitespace of the following line, else use double the whitespace
```
yagebu: doubling the indentation is weird

fava/docs/features.md
```diff
@@ -121,33 +122,64 @@ found in the journal:
...
 
 If no setting is specified, Fava tries to guess the language from your browser
 settings.
 
 ## Documents upload
 
+One or more documents can be uploaded by drag and drop on account names or
+Journal rows.
+
+### Uploading normal documents
```
yagebu: remove "normal", "normal document" doesn't make sense

fava/docs/features.md
```diff
@@ -121,33 +122,64 @@ found in the journal:
...
-The file will be then stored in your Beancount documents folder, in a sub-folder 
-named after the account. You can set the path to the Beancount documents folder 
-by specifying the `option "documents" "/User/test/invoices"`-option (absolute 
+The file will be then stored in your Beancount documents folder, in a sub-folder
+named after the account. You can set the path to the Beancount documents folder
+by specifying the `option "documents" "/Users/test/invoices"`-option (absolute
 or relative to your Beancount file) in your Beancount file.
+
+Beancount will automatically discover files in your `"documents"`-folders and
+emmit corresponding Document entries.
```
yagebu: "generate Document entries for them"

fava/plugins/link_statements.py
```diff
@@ -0,0 +1,68 @@
...
+    for i, entry in all_transactions:
+        statements = [value for key, value in entry.meta.items()
+                      if key.startswith('statement')]
+
+        for posting in entry.postings:
+            if posting.meta:
+                statements.extend([value for key, value in posting.meta.items()
+                                   if key.startswith('statement')])
+
+        _hash = hash_entry(entry)
```
yagebu: Maybe the full hash is a bit much for the link. How about just the first 8 or 10 characters?

aumayr:
> We allow posting-metadata, but don't use any information about the posting then. We should only allow transaction-level metadata, let's not create more places for bugs to occur.

I never thought of doing that, but you are absolutely right: When dropping on a posting, just the upload path should be adjusted accordingly, but the metadata entry should be added to the transaction instead.

> The plugins needs test. We should strive for ~100% coverage in the code we add.

I will add more tests, but am not very experienced in writing them.

> I'd really like to see filename matching in the first version.

The algorithm you suggested sounds good, will implement that.

And thanks for the comments on the code - will review them all.

fava/api/file.py
aumayr: As adding the metadata key to the posting, this is not needed anymore anyway.

fava/plugins/link_statements.py
yagebu: https://en.wikipedia.org/wiki/Birthday_problem#Probability_table

So with 8 characters, you'd need ~9300 links to get a 1% chance of a collision. I think we can take those chances ;)

yagebu:
> When dropping on a posting, just the upload path should be adjusted accordingly, but the metadata entry should be added to the transaction instead.

👍 

> I will add more tests, but am not very experienced in writing them.

For the plugin, it's easiest to use the `load_doc` test fixture probably. As you can write the Document entries in the doc string too, I guess you don't need to write/read any files (if you do, there's the `tmpdir` fixture). Generally one should try to write some test for every possible code path. For this, checking the coverage is really helpful, you can run `pytest --cov=fava --cov-report=html [...]` to generate a html report that shows line-by-line coverage (or wait for the codecov coverage that's generated after successful Travis runs).

Coverage is not everything and realistic tests are best, but it's a helpful indicator that highlights code that is either unnecessary or should be tested.

aumayr:
While writing tests I think I found a counterintuitive behavior by beancount:

When you manually add a `document` directive to your beancount file that points to a file that would otherwise be automatically picked up by beancount ("discovered"), two document entries are emitted by beancount. I would expect that only documents that conform to the discovery-feature-prereqisites  [1] but are not explicitly covered by `document`-directives are added by beancount.

@blais Is this intended behavior?

[1] in the documents folder; with a date at the beginning; in a folder corresponding to an account

yagebu:
That is a known bug in Beancount if I recall correctly.

fava/plugins/link_statements.py
```diff
@@ -37,7 +37,8 @@ def link_statements(entries, options_map):
             statement_p = normpath(join(dirname(entry.meta['filename']),
                                         statement))
             documents = [(j, document) for j, document in all_documents
-                         if document.filename == statement_p]
+                         if document.filename == statement_p and
+                         document.meta['lineno'] == 0]
```
yagebu: Remove. All document entries should be supported, not just auto-generated ones.

fava/plugins/link_statements.py
```diff
@@ -20,6 +22,8 @@ def link_statements(entries, options_map):
     errors = []
 
     if 'documents' not in options_map or len(options_map['documents']) == 0:
+        errors.append(LinkStatementError(
+            'link_statements requires "documents" option to be set'))
```
yagebu: The plugin should also work if there's no document folder (and all documents are given as entries in the file), so this shouldn't be an error.

blais:
I'm not sure it's a bug - you're assuming the user doesn't want dups, and
they're telling you they do - but I think it would make sense to avoid
creating multiple identical directives by default, or at least to have an
option to suppress them.

yagebu:
> I'm not sure it's a bug - you're assuming the user doesn't want dups, and
they're telling you they do - but I think it would make sense to avoid
creating multiple identical directives by default, or at least to have an
option to suppress them.

You're right - it's not a bug really. As you describe it, the current behaviour is not wrong but might be surprising.

aumayr:
> > I'm not sure it's a bug - you're assuming the user doesn't want dups, and
> they're telling you they do - but I think it would make sense to avoid
> creating multiple identical directives by default, or at least to have an
> option to suppress them.
> 
> You're right - it's not a bug really. As you describe it, the current behaviour is not wrong but might be surprising.

Surprising it is!

I vote for adding the following condition to the beancount-auto-document-directive-generation-code: 
If there is a document directive in the beancount file with the same date, account and filename, do not auto-gen a document directive.

aumayr:
Support for relative metadata paths (absolute, relative, and just the filename) is now implemented.

Only problem: When the path is just the filename, the download link is not working, as it is just the filename: 

https://github.com/beancount/fava/blob/36bbf77a900e326c05d4c57f3131e1681333f933/fava/application.py#L186-L190

@yagebu Any idea how - and more importantly: where - to fix this?

aumayr:
> @yagebu Any idea how - and more importantly: where - to fix this?

I tried to solve this in bf470bc. Please have a look at the code.

fava/plugins/link_statements.py
```diff
@@ -0,0 +1,65 @@
...
+    for i, entry in all_transactions:
+        statements = [value for key, value in entry.meta.items()
+                      if key.startswith('statement')]
+
+        _hash = hash_entry(entry)[:8]
+        for statement in statements:
+            statement_p = normpath(join(dirname(entry.meta['filename']),
+                                        statement))
+            documents = [(j, document) for j, document in all_documents
+                         if (document.filename == statement_p) or
```
aumayr: It is the former case :)

yagebu:
> I tried to solve this in bf470bc. Please have a look at the code.

It would be best to split this off into a separate route. This currently mixes very unrelated logic for checking if it's a valid document and finding the correct path for statements. I guess we have to pass something like the lineno to make sure we generate the path from the correct statement. Also I wouldn't modify the statement filepath in the template. The account name of the first posting is not necessarily the right one so it add some extra complexity without really helping.

yagebu:
It's also possible to get the full path when rendering the template but that might be too expensive.

aumayr:
@yagebu I fixed/implemented all of this in fa85e8d

yagebu:
Apart from the lint, this looks good!

aumayr:
@yagebu I fixed the lint problems. Can you merge the PR?

yagebu:
> Can you merge the PR?

Will do!

PR #418, opened 2016-11-22
added notification for upcoming events

fokusov:
default


fava/api/__init__.py
```diff
@@ -73,7 +73,7 @@ class BeancountReportAPI():
         'charts', 'custom_entries', 'date_first', 'date_last', 'entries',
         'errors', 'fava_options', 'filters', 'is_encrypted', 'options',
         'price_map', 'queries', 'root_account', 'sidebar_links', 'title',
-        'watcher']
+        'watcher', 'upcoming_events']
```
yagebu: Maybe keep this list sorted

fava/api/__init__.py
```diff
@@ -157,6 +157,15 @@ def _apply_filters(self):
             self.date_first = self.filters['time'].begin_date
             self.date_last = self.filters['time'].end_date
 
+        self.upcoming_events = []
```
yagebu: This whole block should not be run on every filter, it belongs in `load_file`. It would also be a bit neater if this was a function (like `_sidebar_links`)

fava/templates/_aside.html
```diff
@@ -13,7 +13,11 @@
     {% if title_ %}<h3>{{ title_ }}</h3>{% endif %}
     <ul class="navigation">
     {% for id in menuitems %}
-        <li><a{% if id == active_page %} class="selected"{% endif %} data-key="{{ globals.all_pages[id].1 }}" href="{{ url_for('report', report_name=id) }}">{{ globals.all_pages[id].0 }}</a>
+        <li><a{% if id == active_page %} class="selected"{% endif %} data-key="{{ globals.all_pages[id].1 }}" href="{{ url_for('report', report_name=id) }}">{{ globals.all_pages[id].0 }}
+            {% if id == "events" and api.upcoming_events|length > 0 %}
```
yagebu: `api.upcoming_events|length > 0` -> `api.upcoming_events`

fokusov: I'm sorry for this, but my python skills is almost zero... 
I understood other changes and please give me advise how do I must change this block?

yagebu: Simply replace `api.upcoming_events|length > 0` by `api.upcoming_events`. In Python an empty list has the implicit boolean value `False`.

aumayr:
I really like this! Two suggestions:

- I would change the color of the bubble to something grey - it is not an error, therefore red is somewhat "aggressive". But that is something I can change after being merged as well.

- The `4` in the `if` (day-range when the notification should be shown) should be configureable. And if it is configured to `0`, this feature would be disabled.

fokusov:
It would be great!
Thank you :)

Issue #413, opened 2016-11-12
Journal: Show checkmark for transactions that have a statement attached to them


aumayr:
Currently one has to display metadata, and go through each line in the Journal to check this. 

I want to display a small grey checkmark on the right side of a transaction to indicate that at least one statement is attached to this line. 

This makes it easier to spot transactions that are missing a statement.

yagebu:
To make this more universal, we could show - next to the posting bubbles - the initial characters of the metadata keys.

PR #423, opened 2016-11-25
Journal: Show checkmark for transactions that have a statement attached to them

aumayr:
implements #413

Shows the first two characters of metadata keys as little bubbles.

Looks like this:

screenshot

### DIFF
--- a/fava/static/sass/_journal-table.scss
+++ b/fava/static/sass/_journal-table.scss
@@ -241,13 +241,27 @@
     display: inline-block;
     height: 6px;
     margin-right: 4px;
-    margin-top: 8px;
+    margin-top: 10px;
     padding: 0;
     width: 6px;
 
     &:last-child { margin-right: 10px; }
 
     &.leg-pending { background-color: $color-journalentry-postingwarning; }
+  }
+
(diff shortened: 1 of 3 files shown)


fava/static/sass/_journal-table.scss
```diff
@@ -244,6 +244,20 @@
     &.leg-pending { background-color: $color-journalentry-postingwarning; }
   }
 
+  .metadata-indicator {
+    font-size: 10px;
```
yagebu: could you make `sass-lint` happy here?

aumayr: Which specific error do you mean? I only get warnings from `sass-lint`, not errors.

yagebu: I mean the warnings

aumayr: All of them? The warning for this line is in about 40 other lines as well…

yagebu: There's only one other place where it occurs currently as far as I can see. In any case, there should never be more warnings after a commit really, only less, otherwise we'll never get rid of them.

PR #424, opened 2016-11-25
add loading indicator for async page load

### LINKED ISSUE #399: Partial page loads don't offer an indicator
After I click on something, I'm not quite sure if it's loading.  We should find a sweet loading spinner to use when a partial is loading.

yagebu:
Replace the page logo with a spinner.

Close #399.


yagebu:
indicator

Issue #407, opened 2016-11-04
Editor: Keyboard shortcut for commenting out a line/selection


aumayr:
Add a keyboard shortcut for commenting out the selected lines or the line the cursor is in (if there is no selection).

On a Mac this would be Shift + Cmd + 7

yagebu:
> On a Mac this would be Shift + Cmd + 7

I guess you mean `Cmd`+`/`, that's what Atom and Sublime apparently, or what do you mean by "On a Mac"?

aumayr:
That's what I meant 🙈

aumayr:
Should be trivial: 

* http://codemirror.net/doc/manual.html#addon_comment
* https://github.com/codemirror/CodeMirror/commit/d4f05936c278ea12ff41eb766a1cd9935e37c965

PR #425, opened 2016-11-26
Editor: Keyboard shortcut for commenting out a line/selection

aumayr:
Implementation of #407 

`Cmd`+`/` does not work for me on a Mac (Chrome/Safari), as it opens the help menu of the browser. Mapped it to `Cmd`+`y` for now. Any suggestions about that?


yagebu:
> Mapped it to Cmd+y for now. Any suggestions about that?

I don't have a better suggestion ;)

PR #441, opened 2016-12-25
Query

yagebu:
The first commit adds our own syntax and completion for the Beancount query language.

I quite like the interactive `bean-query` shell, so I've tried to bring Fava's query page up to par with it (by basically wrapping it). Fava's query page can now run all the shell commands, e.g. the help commands. Additionally, all query results are in collapsible boxes and old results are kept on the page when executing a new query. These changes also allow us to easily access the history of shell commands which can be re-run with one click.

This it what it looks like:
query


aumayr:
This looks very useful and absolutely gorgeous!

PR #443, opened 2016-12-27
Initial stab at a simple hook design (for autocommit, initially)

ngoonee:
Here's my current (working for me) implementation.

No docs yet (quite simple, basically just to say that the following)

1. The custom option "post-add-hooks" should be named the module name (e.g. 'myhooks'), with the corresponding module/file being in the same working directory as the beancount file.
2. The myhooks.py file contains functions with names starting with 'add_transaction' or 'add_document_metadata' depending when the hook should run. Arguments are as stated in hooks.py.

An example myhooks.py looks like this:-

    import git, os
    def add_transaction_git_commit(transaction):
        gitobj = git.cmd.Git(os.getcwd())
        retval = gitobj.pull()
        print(retval)

I'd include the example in this PR as well as docs, but not sure if it's 'proper' code wise, and I don't know where such examples would go in the code anyway.


ngoonee:
@yagebu any comments?

yagebu:
> any comments?

I didn't have a detailed look at your code as I'll implement a more general extension architecture, so the only part of this PR I would merge would be an example. Will see what parts of this PR I can use for that - for example I quite like that the hooks can be in a module relative to the Beancount file. Currently I'm imaging something similar to what [beets](https://github.com/beetbox/beets) has - each extension would inherit from, say, `FavaExtension`, which would provide a decorator to register event listeners a.k.a. hooks.

> An example myhooks.py looks like this

Would it be possible to only use the git command line interface and not have some other dependency? Then we could more easily ship this with Fava (at `fava.ext.auto_commit`).

ngoonee:
Thanks, that's awesome! Looking forward to it.

To use the git command line interface I guess subprocess would be fine, let me test and update this to reflect that.

ngoonee:
This works fine for me, using subprocess. Unlike the previous simpler example this is an auto-commit.

    import os, subprocess
    '''
    Example hook for fava.

    The hook add_transaction_git_commit will automatically add all files with the
    beancount suffix in fava's root folder and commit them with a message based on
    the transaction's narration, payee, and date.

    To limit external dependencies, this uses the subprocess module (2.4+). Please
    read the documentation for that[1], especially if you're considering using
    shell=True (probably a bad idea).

    [1] - https://docs.python.org/2/library/subprocess.html
    '''


    def add_transaction_git_commit(transaction):
        subprocess.call(["git", "add", "*.beancount"])
        message = u''
        if transaction.narration:
            message += '"{}"'.format(transaction.narration)
        if transaction.payee:
            if message:
                message += ' paid to "{}"'.format(transaction.payee)
            else:
                message = 'Payment to "{}"'.format(transaction.payee)
        if message:
            message += ' on {}'.format(transaction.date)
        else:
            message = 'Transaction on {}'.format(transaction.date)
        subprocess.call(["git", "commit", "-m", message])

yagebu:
Will be implemented in #477, so closing this.

PR #450, opened 2017-01-03
Collapse inventories into units when displaying journal balances and changes

xentac:
If you make many transactions containing different commodities within a journal
screen (e.g. parent investment account), each journal line becomes dominated by
the list of individual positions.  This will reduce all of the amounts of a
single commodity into a single commodity into a single line.

### DIFF
--- a/fava/templates/_journal_table.html
+++ b/fava/templates/_journal_table.html
@@ -102,7 +102,7 @@
 {%- endmacro %}
 
 {% macro render_inventory(inv) -%}
-    {% for pos in inv %}
+    {% for pos in inv.units() %}
         {{ pos.units|format_amount }}<br>
     {% endfor %}
 {%- endmacro %}


aumayr:
Can you post a screenshot?

xentac:
I will try to do it tonight.

xentac:
PR template
Current template

xentac:
Top screenshot is the PR, bottom is the existing functionality.

It's still not great, but it is better.

blais:
Yeah, this sucks indeed. I've been wondering for a long time what to do.
One easy cop-out is to convert to cost if there is more than a single
commodity in an account, and to somehow render that fact. I don't know.

xentac:
Even if you have a single commodity in an account that you purchase frequently, the lines will still become large an unreadable.  From the journal, knowing I have two lots of 0.50 and 0.60 VTI doesn't help me any more than knowing I have a balance of 1.10 VTI.

yagebu:
Thanks!

Issue #453, opened 2017-01-03
Filtering by time displays different net worth lines than unfiltered views


xentac:
I have two operating currencies (USD and CAD), this may be contributing.

Looking at an unfiltered net worth graph vs. a filtered one for the same time period gives similarly shaped graphs with different total amounts.

I will try to create a minimal example, but I think I need a few months worth of data... we'll see.

xentac:
I have narrowed this down to the way we calculate net worth at particular dates: https://github.com/beancount/fava/blob/master/fava/core/holdings.py#L184

Turns out that the summarizing code does take into account unrealized transactions, but this net worth code doesn't.  That's why filtering changes the apparent values.  I'm pretty sure the correct way to do this is to drop the FLAG_UNREALIZED check in this case.  The original code was probably copied from the holdings function which correctly ignores unrealized transactions (because the unrealized values are displayed as market value).

Given that this method looks to exist only for displaying the balance chart, I'm going to write a patch to remove the filter.

blais:
So this problem highlights a flaw in Beancount: The idea of materializing
unrealized gains into transactions was probably a mistake.  It fails when
materializing gains at different points in time and it causes the need to
filter those here and there. TBH, it was a bit of an experiment; I wanted
to see if I couldn't keep the reporting code as simple as possible by
converting the flow of transactions in this way, but I think it causes too
many problems.

It seems it would be easier to always do the conversion to market value at
reporting time. Reports that care about market value would explicitly have
to make the conversion when necessary, at the desired date, and I should
add functions to make that super easy (underway in branch "conversions").

Perhaps in the interest of avoiding confusion I should remove the
unrealized_gains plugin?
Any thoughts?

xentac:
Like I had mentioned in the other thread, I agree with the eventual goal
and I still plan on working on it. Until that point I'm going to keep using
my unrealized gains plugin ;). To fix the immediate problem I'll write the
patch, but I do agree more and more that the unrealized plugin probably
isn't best.

yagebu:
Ref. the discussion in #467 - It seems to me that adding unrealized gains at the end of the reporting period is still fine and that the issue here was monthly unrealized gains which `summarize` mangled (as it didn't expect to have to deal with unrealized entries).

@xentac: Without your monthly unrealized plugin, do the net worth charts still look different?

xentac:
[...]

Issue #129, opened 2016-02-15
Balance summaries (balance sheet, trial balance) are based on lot price, not current price


xentac:
I would have expected that the prices at the end of the period would be used in the conversion instead of price when I purchased the commodity.  This way makes it much harder to figure out the state of my accounts.

Is there a reason to do it this way?

I can create an example file to show what I'm talking about if necessary.

yagebu:
From an accounting standpoint, the current behaviour is the correct thing to do. The balance sheet should display the "book value" of your assets (and liabilities), that is the original cost of the asset (minus depreciation possibly). Wikipedia has [more](https://en.wikipedia.org/wiki/Book_value) on the meaning of book value and maybe the [Beancount trading doc](https://docs.google.com/document/d/1WjARst_cSxNE-Lq6JnJ5CC41T3WndEsiMw4d46r2694/edit#heading=h.5o43cxhng7b7) gives some further insight.

I agree with you that it would be useful to have a report that reports the account balances based on the market values (and displays charts based on the market value), but it should be separate.

aumayr:
Can I close this, @xentac?

> I agree with you that it would be useful to have a report that reports the account balances based on the market values (and displays charts based on the market value), but it should be separate.

@yagebu Any ambitions to follow up on that with a PR? If yes: Please create a new ticket for it. If no, just discard this comment :-)

yagebu:
> Any ambitions to follow up on that with a PR? If yes: Please create a new ticket for it. If no, just discard this comment :-)

Will do. Although I haven't really thought about how to best do this, so if anyone else is keen or has some ideas, feel free to work on it.

xentac:
Looking at the trading doc, it pointed me at the unrealized_gains plugin.  This is good enough for my purposes right now.  I'm imagining a modification to the plugin to be able to compute unrealized gains every month (or maybe even every price db entry).

aumayr:
@xentac Please have a look at the new views in the Holdings report

PR #467, opened 2017-01-10
Don't exclude unrealized transactions when calculating net worth

xentac:
Fixes https://github.com/beancount/fava/issues/453.  Seems that filtering unrealized transactions when calculating holdings makes sense, but not when calculating the Balance Sheet's chart's net worth values.

### DIFF
--- a/fava/core/holdings.py
+++ b/fava/core/holdings.py
@@ -180,8 +180,7 @@ def convert_inventory(price_map, target_currency, inventory, date):
 
 def net_worth_at_dates(entries, dates, price_map, options_map):
     transactions = [entry for entry in entries
-                    if (isinstance(entry, Transaction) and
-                        entry.flag != flags.FLAG_UNREALIZED)]
+                    if isinstance(entry, Transaction)]
 
     types = options.get_account_types(options_map)
 


yagebu:
I'm not using the unrealized gains plugin, so I might be wrong here, but as far as I can tell, it will insert only one (set of) transaction(s) after the last transactions. How could this affect the net worth chart (except for maybe the last data point)?

Also, currently (in Fava and Beancount) the net worth for a given operating currency is  the sum (over all holdings) of all market values converted to the operating currency. So it is consistent to use the same list of entries (i.e. all but the unrealized ones) - why should we differ from Beancount here?

xentac:
My unrealized plugin is an improvement on the default one (https://github.com/xentac/beancount-plugins-xentac/blob/master/beancount_plugins_xentac/plugins/unrealized_periodic.py).  It posts unrealized gains every month.  That way I can see how my net worth is affected as time goes on.

Can you point me to the code that uses market value and not cost?  I thought that all commodities were being rendered at cost and not at market value, which is why the unrealized plugin was necessary to begin with.  The whole reason I started using the unrealized plugin was because I was seeing all my commodities at cost and not seeing their market values.

xentac:
(Original discussion where I got this impression https://github.com/beancount/fava/issues/129)

yagebu:
> It posts unrealized gains every month. That way I can see how my net worth is affected as time goes on.

Monthly unrealized gains shouldn't be necessary for the net worth chart as it should already take into account the market value.

> Can you point me to the code that uses market value and not cost?

The `convert_inventory` function in `fava.core.holdings`. It first determines the price (in the cost currency, if there is one), then the price of the cost currency in the target operating currency and calculates "number * price * price_of_cost_currency" (the first two factors are the same as the market value).

> I thought that all commodities were being rendered at cost and not at market value, which is why the unrealized plugin was necessary to begin with.

That's true in e.g. the Balance Sheet and Income Statement but not for the net worth chart, which is a holdings-like report (and calculated as explained in my last post). Sorry that I didn't express this clearly in #129 - there I was only referring to the account tree tables.

xentac:
Ok then.  I see.  I'm a little sad that my net worth is not as high as I thought it was ;)

In that case, I will probably drop my usage of the unrealized plugin.  I am a bit sad that I don't currently have a way to look at an individual account's market value.  But this does fit decently well with my original plan of having a market value vs. cost value toggle on the balance sheet to make it easier to see both sets of values.

If I recalibrate my thinking based on this data, it looks like the balance sheet and account monthly changes/balances display cost value.  What do you think about adding a toggle somewhere that displays market vs cost values everywhere?  It would be less confusing than this half one way and half the other way scenario.

I didn't realize there was already code in fava to do market conversion, a patch like that should be much easier then.

yagebu:
A toggle like you describe would definitely be useful. I'm just not sure where it should go and if it being a "global" thing would be the best idea. 'Balance Sheet' has a specific meaning and if we display the market value it would rather turn into a "holdings account tree" (which would be useful, but maybe belongs on a different report)

> I didn't realize there was already code in fava to do market conversion, a patch like that should be much easier then.

I'm pretty sure there's a function in Beancount that calculates the market value (for the net worth, we might need an additional conversion, so it's done in Fava)

xentac:
See the discussion here: https://groups.google.com/d/msg/beancount/zK6sFnzY4fg/msTTPVDTCwAJ

It does have various ways to do it but nothing canonical and nothing to evaluate at a specific date.  Martin has started to refactor it, but he's not done yet.

I think having separate reports might be a bit tricky because the "holdings account tree" would have to link to a "holdings account journal" but the balance sheet might link to the regular account journal.  That's why I was thinking that an always visible toggle would make the displays less ambiguous.

I'll leave it to Martin to argue whether we should hew strongly to accounting terms with beancount.

yagebu:
There's already a `get_inventory_market_value` function...

xentac:
I'll start a patch that creates a different holdings accounts (maybe call it Market Balances or something) page and adds two new Changes/Balances links in the account journal so we can see what it looks like and kick the can down the road on the global toggle decision.

blais:
prices.get_inventory_market_value() and prices.get_position_market_value() will disappear. They're being refactored into something nicer. Tune in to branch "conversions". I'm removing the various incomplete conversion functions from all over and centralizing them all into beancount.core.convert. (Ongoing.)

PR #475, opened 2017-01-12
Some tree-table styling improvements.

yagebu:
The first commit makes Fava use flexbox to wrap on small page-width so we don't have to have a hard-coded media query when to switch from two columns to one.

The second commit tries to save horizontal space in some other places than the number columns (padding between tables, also resizing the accounts column).

Currently the width and font size of the columns for three or more operating_currencies is much too small, I'd be for only having two cases: one, or more columns.

There's no such thing as one phone screen size anyway and mobile browsers seem to ignore the styles in these cases and try to resize it to the right width anyway (which yields a much better result as the resizing is more even). Or it's just a tap away.

### DIFF
--- a/fava/static/sass/_journal-table.scss
+++ b/fava/static/sass/_journal-table.scss
@@ -121,7 +121,7 @@
   dl {
     font-size: .9em;
     margin: 0;
-    padding: 2px 0 2px 8rem;
+    padding: 2px 0 2px 8em;
 
     dt,
     dd {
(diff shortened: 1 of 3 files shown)


xentac:
One of the things I've noticed initially is that the tables don't expand more than a certain size.  This is probably fine, I can't think of a reason we wouldn't want them to, but this is different from what used to happen.  It does make the screen seem a little lopsided when I leave the screen at some of my default resolutions.

xentac:
Overall I'm happier with this than my original attempt.

hoostus:
This looks like a nice improvement on my data :smile:

yagebu:
> One of the things I've noticed initially is that the tables don't expand more than a certain size. This is probably fine, I can't think of a reason we wouldn't want them to, but this is different from what used to happen. It does make the screen seem a little lopsided when I leave the screen at some of my default resolutions.

Yes it can look weird on some resolutions. But without a max-width, the tables can get terribly wide which makes them harder to read. I think having a slightly lopsided layout is preferable to full-width tables on wide pages (I've tried centering them, but that looks even weirder IMO). Should I increase the max-width?

xentac:
I think it's fine.  It fits all the data, which is the important part.

PR #479, opened 2017-01-16
Add "incognito" fava-option to hide digits

aumayr:
Follow-up to PR #457, see discussion there.

### DIFF
--- a/fava/application.py
+++ b/fava/application.py
@@ -159,16 +159,19 @@ def _perform_global_filters():
         g.filters[exception.filter_type] = None
         flash(str(exception))
 
+
 @app.after_request
 def _incognito(response):
     """Replace all numbers with 'X'."""
     if (g.ledger.fava_options['incognito'] and
-       response.content_type.startswith('text/html') and
-       not (request.endpoint == 'report' and
-            request.view_args['report_name'] == 'editor')):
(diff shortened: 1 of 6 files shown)


aumayr:
This does not implement a command line option when starting Fava, as I do not know how to elegantly pass the option from `cli.py` to the app object.

fava/application.py
```diff
@@ -159,6 +160,19 @@ def _perform_global_filters():
         flash(str(exception))
 
 
+@app.after_request
+def _incognito(response):
+    """Replace all numbers with 'X'."""
+    if (g.ledger.fava_options['incognito'] and
```
yagebu: this shouldn't fail if `g.ledger` doesn't exist

yagebu:
A test would be nice

tests/test_application.py
```diff
@@ -76,3 +76,14 @@ def test_jump_handler(app, test_client, referer, jump_link, expect):
             expect)
         assert result.status_code == 302
         assert get_url == expect_url
+
+
+def test_incognito(app, test_client):
+    with app.test_request_context():
+        app.preprocess_request()
+        flask.g.ledger.fava_options['incognito'] = True
```
yagebu: This should probably be reset after this test as the ledger object will be shared between tests

PR #485, opened 2017-01-22
Transaction form: Support metadata

aumayr:
This PR implements input fields in the transaction form to enable the entering of metadata. 

The key and value are inserted into the file unchecked, so the rules in the Beancount Language Specification (keys should start with a lowercase character from a-z, etc.) are not checked or enforced.

(The next goal is to enable the upload of statements when adding a new transaction with the form.)

### DIFF
--- a/fava/core/file.py
+++ b/fava/core/file.py
@@ -153,6 +153,10 @@ def _render_transaction(transaction):
         transaction.date, transaction.flag, transaction.payee,
         transaction.narration)]
 
+    if transaction.meta and len(transaction.meta.keys()) > 0:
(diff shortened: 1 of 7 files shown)


yagebu:
I've been offline for a few days and didn't have time to have a look at this yet. Should be able to review it in the next few days.

yagebu:
"data" is already plural so `metadatas` should be changed to `metadata`. Otherwise this looks good 👍

aumayr:
`metadata` is already used as a class elsewhere, that's why I did choose `metadatas`.

PR #488, opened 2017-01-31
Add Chinese translations

tichai:
Translations in Simplified Chinese


aumayr:
Thanks @tichai for the translation!

We currently manage all translations here, including a Chinese one (has not been merged yet): https://poeditor.com/join/project/TlraSxOCxt

Can you join there and check if the current Chinese one is correct?

PR #494, opened 2017-02-02
Fixing the build process on Mac OS

johannesjh:
Fava's `make binaries` build target reproducably did not work on my machine. As it turned out, there were several problems involved, which I fixed in the proposed pull request.

Apart from minor fixes or improvements, the most notable changes are:

- Python 3.5 is required because of pyinstaller compatibility, but a simple `brew install python3`  installs python3.6. I therefore changed the provisioning script to install the correct version using pyenv:

    ```bash
    brew install pyenv
    env PYTHON_CONFIGURE_OPTS="--enable-shared" pyenv install 3.5.3
    pyenv global 3.5.3
    pyenv init
    echo 'eval "$(pyenv init -)"' >> ~/.profile
    ```

- Two Jinja2 modules cause problems with Pyinstaller because they only work with Python >= 3.6. For details, see https://github.com/pyinstaller/pyinstaller/issues/2393. The problem can be solved by excluding these modules in the pyinstaller spec file:

    ```python
    a = Analysis(['../fava/cli.py'],
             # ...
             excludes=['jinja2.asyncsupport','jinja2.asyncfilters'],
             # ...
             )
    ```


yagebu:
Thanks for the work on this! The development version of PyInstaller now has support for 3.6 (and also fixed the jinja2-related bug). I'd prefer to install the dev version and build the binary with that.

Vagrantfile
```diff
@@ -10,6 +10,10 @@
 # - CentOS 7 - build for 64bit Linux
 # - OS X Yosemite (10.11) - build for 64bit MacOS
 
+unless Vagrant.has_plugin?("vagrant-scp")
```
yagebu: :

Vagrantfile
```diff
@@ -33,13 +42,14 @@ Vagrant.configure("2") do |config|
...
     config.vm.define "centos" do |b|
         b.vm.box = "centos/7"
+        b.vm.provision "Fix shared folder permissions", type: "shell", inline: "chown -R vagrant /vagrant"
         b.vm.provision "Install dependencies", type: "shell", inline: centos_install
         b.vm.provision "Build binary", type: "shell", inline: build, run: "always"
     end
 
     config.vm.define "darwin" do |b|
-        b.vm.box = "jhcook/osx-elcapitan-10.11"
+        b.vm.box = "jhcook/macos-sierra"
```
yagebu: Please revert this. For building the binaries, older is better, as it ensures that we build a binary that can run on both El Capitan and Sierra.

johannesjh:
Ok, in commit b4b4bc6 I switched to using the dev version of pyinstaller. In consequence, my changes in the `pyinstaller.spec` file are not necessary anymore, so I reverted them. And I switched back to simply installing python using `brew install python3` instead of using pyenv. I am not entirely sure about this last change (pyenv vs. direct installation via brew) - what do you think?

johannesjh:
...and in commit dbf672a I fixed a copy paste error related to  folder permissions. I just gave it a spin and it all build and runs nicely on macos.

yagebu:
I've merged this (together with some other slight changes to the {Vagrant,Make}files).

Issue #464, opened 2017-01-08
Better Assets / "Net worth" reporting


redstreet:
In #129, it was stated that balance summaries should be based on book value, not market value. This covers the account standpoint. However, in addition, a net worth reporting standpoint is useful to answer such question as:

- What is the current state (market value) of each of my accounts? (#129 also asks this)
- How are my assets spread across {accounts, currencies, lots} based on market value?

The holdings page shows the basic data on this. Minor extensions (#462) would make it better, but what would be really useful is to have a version of the treemap and sunburst charts for better net worth reporting. Perhaps as a start, the "Holdings by Account" and  "Holdings by Currency" would benefit from these two charts.

Anyone else have thoughts on this?

xentac:
My plan was to add a market/cost toggle to the balance sheet, so that
balances could be calculated and displayed with the graph there. I just
haven't had time to write any of it.

redstreet:
@xentac that sounds like a good idea, and covers all the basic net worth info.

xentac:
Given the discussion in https://github.com/beancount/fava/pull/467, my current plan forward is to create new pages that display market value.  One "Market Balances" page similar to the "Balance Sheet" but only displaying market values and two new links in the account journal that calculates market value for Changes and Balances.

xentac:
Consider https://github.com/beancount/fava/pull/468 my first attempt.  I picked some random variable names (like conversion being either 'cost' or 'market').  It should cover my use cases though.

yagebu:
The 'cost/value' select is now implemented (#504)

Issue #242, opened 2016-03-27
Convert expenses to single currency


goshakkk:
I am trying to get on the CLI accounting bandwagon and trying out beancount as it seems to be the most semantically cohesive (also has fava!)

Is there a way to have expenses converted into a single currency (USD) in the charts (account page & income statement)? As well as everything on the trial balance page

<img width="1481" alt="screen shot 2016-03-27 at 5 41 22 pm" src="https://cloud.githubusercontent.com/assets/315596/14065882/bda7b4f8-f443-11e5-9871-fd309d2323ed.png">

Setting two operating currencies results in ugly charts which are basically useless (given that in terms of my national currency, 1 USD = 20,300 BYR; and only credit card transactions are in BYR, everything else is dollars only)

<img width="981" alt="screen shot 2016-03-27 at 5 46 13 pm" src="https://cloud.githubusercontent.com/assets/315596/14065888/dfc9d66a-f443-11e5-81b6-f53a0aa5194d.png">

FWIW, there are `price` statements in the file, and conversion works manually using `convert(sum(position), 'USD')`.

Ideally, I'd love to have everything converted into dollars and plotted as such. Is there a way to achieve that currently in fava?

Thanks!

yagebu:
This is currently not possible in fava, but it's a useful feature that we should have. There's an [issue](https://bitbucket.org/blais/beancount/issues/103/make-the-balance-sheet-and-income) on the beancount issue tracker for this as well.

This is something I've been meaning to implement for some time, but it doesn't have the highest priority for me. So no idea if I'll get round to do it anytime soon. If you know Python and want to get your hands dirty, I could give you some pointers, otherwise you'll have to wait for me or someone else to implement this.

goshakkk:
I do know some Python and would appreciate any pointers on this feature, might get some time on the weekend to work on that

yagebu:
- We need a function to convert an inventory to the specified currency: `convert_inventory(price_map, target_currency, inventory)`. This can probably be done similarly to `beancount.ops.holdings.convert_to_currency`, as a Position and Holding have basically the same attributes (and an Inventory is a list of Positions). `ops.prices`, `ops.holdings` and the basic beancount data types should be all that's needed. This function could go in `fava.api.helpers`.
- `fava.api.serialization.serialize_inventory` should take the currency as an optional argument and then use the above conversion.
- To link this to the frontend, the currency could be passed to the api as a "filter".
- There should be some select/dropdown form to select the conversion in the interface.

[...]

PR #504, opened 2017-02-16
Conversion

yagebu:
[The first commit updates our code to use the new shiny conversion utilities in `beancount.core.convert`. @blais, could we get a new release of Beancount so that we can depend on this?, edit: this has been cherry-picked into master]

More importantly, this allows us to offer a toggle to show market values instead of cost values (ref #464).

This should also allow us to convert everything to a single currency (later, ref #242).

### DIFF
--- a/fava/application.py
+++ b/fava/application.py
@@ -186,6 +186,8 @@ def _inject_filters(endpoint, values):
         return
     if 'interval' not in values:
         values['interval'] = request.args.get('interval')
+    if 'conversion' not in values:
+        values['conversion'] = request.args.get('conversion')
     for filter_name in ['account', 'from', 'payee', 'tag', 'time']:
         if filter_name not in values:
             values[filter_name] = g.filters[filter_name]
@@ -199,6 +201,7 @@ def _pull_beancount_file(_, values):
     if g.beancount_file_slug not in app.config['FILE_SLUGS']:
         abort(404)
     g.ledger = app.config['LEDGERS'][g.beancount_file_slug]
+    g.at_value = request.args.get('conversion') == 'at_value'
 
 
 @app.errorhandler(FavaAPIException)
--- a/fava/core/charts.py
+++ b/fava/core/charts.py
@@ -4,12 +4,14 @@ import datetime
 from beancount.core import flags, convert, realization
 from beancount.core.amount import Amount
 from beancount.core.data import Transaction
-from beancount.core.number import Decimal
+from beancount.core.display_context import DisplayContext
+from beancount.core.number import Decimal, MISSING
 from beancount.core.position import Position
 from beancount.core.inventory import Inventory
 from beancount.core.data import iter_entry_dates
 from beancount.utils.misc_utils import filter_type
 from flask.json import JSONEncoder
+from flask import g
 
 from fava.util import listify, pairwise
 from fava.core.helpers import FavaModule
@@ -17,6 +19,7 @@ from fava.core.helpers import FavaModule
 
 class FavaJSONEncoder(JSONEncoder):
     """Allow encoding some Beancount date structures."""
+
     def default(self, o):  # pylint: disable=E0202
         if isinstance(o, Decimal):
             return float(o)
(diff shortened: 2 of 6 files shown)


blais:
Yo got it
2.0b14

blais:
Sorry, made a release on booking branch by accident...
releasing 2.0b15 instead (correctly this time).

yagebu:
Thanks!

ivan-the-guy:
@yagebu how can I make this work? I tried to run the development version of fava. I do see new select field with 'At value' and 'At cost' options, but changing its value does nothing (well, it produces a js exception in console that transition is not a function - I guess it has something to do with the rollup bundler) and providing `'conversion=at_value'` in the address bar changes nothing on the initial graph render.

What am I missing here?

Thanks.

yagebu:
Without more to go on I can't really debug the JS error.

Do you have prices for the commodities that you want to see displayed at market value? Do these show up in the tree tables if you add `conversion=at_value` to the query string?

ivan-the-guy:
> Without more to go on I can't really debug the JS error.

Stacktrace
```
Uncaught (in promise) TypeError: this.selections.cells.transition is not a function
    at TreeMapChart.zoom (http://localhost:5000/static/gen/app.js:13298:35)
    at TreeMapChart.update (http://localhost:5000/static/gen/app.js:13279:10)
    at TreeMapChart.draw (http://localhost:5000/static/gen/app.js:13259:10)
    at HierarchyContainer.draw (http://localhost:5000/static/gen/app.js:13834:10)
    at http://localhost:5000/static/gen/app.js:13999:14
    at Array.forEach (native)
    at initCharts (http://localhost:5000/static/gen/app.js:13908:13)
    at http://localhost:5000/static/gen/app.js:32398:3
    at http://localhost:5000/static/gen/app.js:2036:5
    at Array.forEach (native)
```

> Do you have prices for the commodities that you want to see displayed at market value?

I do. I can see them on Commodities page

> Do these show up in the tree tables if you add conversion=at_value to the query string?

Sorry, I didn't understand what you mean. Could you please share a screenshot?

ivan-the-guy:
If I issue query 
```select narration, position, convert(position, 'RUB');``` I am able to see correct output in the bean-check command which makes me think that the problem is on fava side, not in the journal.

yagebu:
Not sure if the Javascript error is due to incorrect bundling or actually valid but it doesn't seem to do any harm.

Without any more information about what kind of output you would expect from Fava, but which you are not seeing, it's kind of impossible to see what's not working on your end. This feature works as expected for me and as it basically just relies on Beancount's conversion functions so I'm quite certain it works correctly.

yagebu:
> Not sure if the Javascript error is due to incorrect bundling or actually valid but it doesn't seem to do any harm.

actually it seems like the d3-transition dependency was not correctly bundled.

ivan-the-guy:
> Without any more information about what kind of output you would expect from Fava, but which you are not seeing, it's kind of impossible to see what's not working on your end. 

The problem is I don't see any changes on pages at all when I pass `conversion=at_value` param in urls. It makes me think that the feature doesn't work.

yagebu:
Depending on your account balances, there might not be any changes...

Let's say you have `10 HOOLI @ 5 USD` in one account. At cost, this should be displayed as `50 USD`, at market value it will be displayed at `10 HOOLI` if you don't have a price for it or if you, say, have a price of `6 USD` for it, it should show as `60 USD` at market value.

ivan-the-guy:
Would you mind produce a minimal ledger file (say, 2 accs and 3-4 transactions with different) that works on your fava installation?

I am also encountering a lot of other bugs in fava (e.g. sunburst graph doesn't update on click) and it is really hard to debug everything at the same time.

Thank you

ivan-the-guy:
@yagebu you are right, rollup doesn't include d3-transition in the resulting bundle. 
Importing selection from `d3-selection` solves the problem (found it here - https://bl.ocks.org/mbostock/97557a39b4bfc8229786c8bccb54074d ).

I am not sure why it happens. I guess using `yarn` instead of `npm` would help to prevent the problem from reappearing.

yagebu:
> Would you mind produce a minimal ledger file (say, 2 accs and 3-4 transactions with different) that works on your fava installation?

Try the example file in the tests directory. The balance of the Assets account changes when switching between cost and value.

> I am also encountering a lot of other bugs in fava (e.g. sunburst graph doesn't update on click)

I don't use the charts much so if other parts of Fava or JS dependencies change they break sometimes. Already found the bug in the sunburst chart. Please do report any other bugs you encounter.

> I guess using yarn instead of npm would help to prevent the problem from reappearing.

How would that help? Maybe your version of npm is outdated?

ivan-the-guy:
> How would that help? Maybe your version of npm is outdated?

Yarn produces `yarn.lock` file which pins exact versions of all dependencies. It means that dev environment will be the same for all of us.

yagebu:
> Yarn produces yarn.lock file which pins exact versions of all dependencies. It means that dev environment will be the same for all of us.

I don't think it's worth the additional build dep for us though.

ivan-the-guy:
> I don't think it's worth the additional build dep for us though.

This is why fava is riddled with 'it works for me' js bugs

yagebu:
> This is why fava is riddled with 'it works for me' js bugs

I can't remember a single bug being caused by not using a lockfile. If it's so riddled with bugs then please report them, just vaguely criticising Fava without basis doesn't really help.

xentac:
Have you tried running fava within shared docker container?  It's a pretty consistent way to get all the deps and build the project.  I personally use a modified version of it and the conversion branch worked before it was even merged.

ivan-the-guy:
> I will point out that I'm running a very recent version of fava and the "this.selections.cells.transition" line in my app.js is on line 9759. It's odd that you'd have an extra 4000 lines of js in your file.

This is why deterministic builds matter. Could you please remove `node_modules` folder (`rm -rf ./fava/static/node_modules/`) and rebuild the app.js file from scratch (`make`)?

> If you'd like, I can give you access to an example file deployed with my docker image so you can narrow down a difference in your setup.

That would be great.

xentac:
Because I use docker, every time I build I start from a fresh clone and exercise the entire build process.  Last time I built was 2 days ago, but I actually just did a new build as of this morning to pick up the convert to any currency changes.  I don't see any stacktraces anywhere to do with conversion.

I actually have oauth set up by default for my docker images, so if you have a gmail account I can authorize, I can give you access right away.  If you want http access, it'll take me a little while to get the unsecure path opened up.

xentac:
I was able to figure out how to enable the github oauth proxy, so it should be working now without a gmail account.

https://bean-dev.xennet.org

Issue #462, opened 2017-01-08
Add export to Holdings page


redstreet:
Hi,
This is a feature request to:
- make the tables shown on the "Holdings" page exportable to csv/xls
- add a totals row to this table

Thanks!

blais:
A couple of comments (won't directly address your request but the topic is
related):

- bean-query can now export any of its outputs using the --format=csv
option.

- I have been using this experimental script to move my uploads to a
spreadsheet for a while:
https://bitbucket.org/blais/beancount/src/0db89eb691221543e6a76eb77cb72fb6a1d23aeb/experiments/upload/upload_assets.py?at=default&fileviewer=file-view-default
After the 2.0 release, I'd like to remove the dependency on gspread and
move it to a tested beancount.scripts.upload script.

- Eventually the Holdings pages should be generated from SQL, not with the
custom code that's currently in use. I'll eventually make changes so that
this is possible.

redstreet:
@blais thanks, good to know, especially about bean-query --format=csv

redstreet:
@blais, is there an easy way to use bean-query --format=csv and get the output for what fava shows as "Holdings by account"? I guess I'm trying to find out if the query used to generate this is built into beancount or if fava has it. Thanks!

blais:
bean-report -f csv <filename> holdings -g account

The web pages are mostly just rendering output produced by code from
beancount.reports, which you can also get by invoking bean-report.

(A note about the future: Eventually I want to remove the Holding object
and replace all of these with equivalent SQL queries. More simplier.)

redstreet:
@blais thank you!

yagebu:
Something like the following should get you quite close to the "Holdings by account" (doesn't have the averages though, and I don't currently know of a way to get rid of empty rows).

    SELECT account, units(sum(position)), cost(sum(position)), value(sum(position))
      WHERE account ~ "Assets|Liabilities"
      GROUP BY account, account_sortkey(account)
      ORDER BY account_sortkey(account)

Regarding your original suggestion: I don't really see how a totals row can make sense for the holdings tables. Summing units doesn't make a lot of sense and neither does any other number in the table if cost currencies are different.

redstreet:
Thanks for the query, @yagebu. Question about the query (cc: @blais): some of the cost or value columns are shown in USD even though the actual units are comprised of various currencies, while in other rows, they are shown natively. Is this because I haven't told beancount how to convert one to another?

[...]

PR #510, opened 2017-03-07
Holding query

yagebu:
Implement the Holdings page using Beancount queries. This allows us to remove a lot of code that basically duplicated functionality in Beancount. For the aggregated holdings this removes some unneeded columns in the holdings tables. Also, this makes exporting and writing holdings-related queries a breeze (ref #462).


yagebu:
Note: this includes that makes Fava's filters (like FROM, Time, etc.) apply to queries too. From the commit:

> This allows combining Fava's filters with a Query. This seems more
consistent as the filters are shown on the Query page and they
already apply everywhere else.

PR #513, opened 2017-03-12
limit the number of circle in the list of transaction

pfrancois:
fava show all the posting in one transaction.
usually it is ok, but i have a case where it is problematic.
business case:
I buy each month some share of a fund. each one or two year, i see what is the result and sell the share bought. as fava show all the different posting, it could have many posting, so i propose a to have a maximum to not break the screen.


yagebu:
I'm against this change as hiding some of the indicators is unexpected and might hide important information for some other user. If a high number of indicators breaks the layout, a fix to the layout would be preferable, could you try providing one instead?

Issue #507, opened 2017-02-22
Budget Accumulation Issues


shmcgrath:
When using the budgeting option (either yearly, monthly, etc.) Fava does not seem to calculate the budget until there is at least one transaction in the Expenses. An example from my files is below:

The budget line in my Beancount file reads like this:
```
2017-01-01 custom "budget" Expenses:Tech:Hardware "yearly" 750.00 USD
```

I expect the program to allocate 750.00 USD for the year. It is fine (excellent, actually) that this does it per day. The problem is that the budgeting does not seem to allocate properly until there is a transaction in that category. This is problematic for large, annual or semi-annual expenses. (I don't pay my car insurance until September) I would expect my budget to be accumulating from the start date (i.e. January 01).

Now for my example:
The transaction that kicks off my Hardware category is $94.99 on February 21. When I look at the weekly balances, the balance for that week in the budget is a total of $15.97. That is not correct. 

If I add a dummy transaction of $0.01 for January 2nd, the remaining budget drops to $9.79 for that week, which is closer to the correct value. If I look at the budgeting daily, the Jan 02 dummy transaction for $0.01 leaves me with a budget of $2.04. 

A couple of clarifying questions: 
- If I am kicking off my annual budget on January 01, shouldn't the January 02 budget have $4.09 in it?
       - Note: If I add another dummy transaction for $0.01 on Jan 01, the budget then properly allocates on Jan 02 ($4.09).
- Why does the weekly budget fluctuate so much with and without a dummy transaction?
- Is there something that I am doing wrong in my files? (There probably is)

Let me know if you have any more questions or if I should be asking this elsewhere. Thank you for all of the hard work on Fava, it is awesome!

aumayr:
Sorry for the late response - I did not have time to look into this yet.

yagebu:
Thanks for the detailed report! I'm currently travelling so I've only had time to really look at this now. I don't use budgets myself and so I don't test this part of Fava's functionality in my daily usage - I've found the following while trying to reproduce this issue.

When the balance for an account is empty, the budget wasn't displayed - this should be fixed now. Otherwise I think these problems come from the reporting intervals that Fava considered. If no time filter is set, it will only consider dates between the first and last transaction, so since the budgeting is computed daily, this might lead to wrong budgets. Can you try setting the time filter to `2017` to see if these issues persist? (I couldn't reproduce this issue with the filter set) If that fixes it, I guess the complete fix will be to also consider the first and last budget for the reporting interval.

[...]

PR #518, opened 2017-03-28
Fix budget accumulation regression.

TomJohnZ:
Fix bug introduced in b94a2c6 where budgets accumulate daily but never
decrement when transactions are posted.

Ref #507.

### DIFF
--- a/fava/templates/_tree_table.html
+++ b/fava/templates/_tree_table.html
@@ -127,7 +127,9 @@
             {% endfor %}
             {% if budget %}
               {% for currency, number in budget.items() if currency not in balance_children.currencies() %}
-                {{ render_budget(budget, currency) }}
+                <span class="balance-children">
+                  {{ render_budget(budget, currency) }}
+                </span>
               {% endfor %}
             {% endif %}
             {% for pos in balance_children %}
--- a/fava/templates/_tree_table.html
+++ b/fava/templates/_tree_table.html
@@ -126,12 +126,13 @@
                 </span>
             {% endfor %}
             {% if budget %}
-              {% for currency, number in budget.items() if currency not in balance.currencies() %}
+              {% for currency, number in budget.items() if currency not in balance_children.currencies() %}
                 {{ render_budget(budget, currency) }}
               {% endfor %}
             {% endif %}
             {% for pos in balance_children %}
                 <span class="balance-children">
+                    {{ render_budget(budget, pos.units.currency, pos.units.number) }}
                     <span class="number">{{ pos.units|format_amount }}</span>
                 </span>
             {% endfor %}


fava/templates/_tree_table.html
```diff
@@ -126,12 +126,15 @@
...
-                {{ render_budget(budget, currency) }}
+              {% for currency, number in budget.items() if currency not in balance_children.currencies() %}
+                <span class="balance-children">
+                  {{ render_budget(budget, currency) }}
+                </span>
               {% endfor %}
             {% endif %}
             {% for pos in balance_children %}
                 <span class="balance-children">
+                    {{ render_budget(budget, pos.units.currency, pos.units.number) }}
```
yagebu: The budget of an account doesn't include any budgets of its children accounts, so comparing it  to `balance_children` is incorrect.

fava/templates/_tree_table.html
```diff
@@ -126,12 +126,15 @@
                 </span>
             {% endfor %}
             {% if budget %}
-              {% for currency, number in budget.items() if currency not in balance.currencies() %}
-                {{ render_budget(budget, currency) }}
+              {% for currency, number in budget.items() if currency not in balance_children.currencies() %}
```
yagebu: same here

yagebu:
The budget of an account only includes budgets posted directly to it, not too any of its children. So it should only be compared to the account `balance`, not `balance-children`.

With that in mind, I don't see the regression in b94a2c6b85cc9f70c077979b47ad80c93c4a206b. If I'm wrong, could you elaborate (maybe with an example) what the problem is with the current code?

yagebu:
> If I'm wrong, could you elaborate (maybe with an example) what the problem is with the current code?

Ahh, saw your comment on the issue. Could you give some details on which accounts you have the budgets and the transactions respectively?

TomJohnZ:
Thanks for comments! You're right; I hadn't realized, but it works fine for transactions posted to the account.

However, in the past, posting to children accounts also decreased the budget of parent accounts. For instance, see [https://gist.github.com/TomJohnZ/153109bdd5628776eab6b4f98fbb73f6.js](url) and go to account/Expenses/balances/?time=2017-01-01+-+2017-01-03&interval=day.

Before b94a2c6, the category1 budget (which should accumulate $1.00 per day) would correctly go negative on 2017-01-01 and increase from there. After that commit, the category1 budget never goes negative and instead increases at $1.00 per day.

I don't know if this was the way budgets were meant to work, but having parent accounts inherit their children's postings was incredibly useful for me; for instance, I would set a budget on an Expenses:Luxuries account, and then postings to Expenses:Luxuries:Food or Expenses:Luxuries:Entertainment would all decrease the Expenses:Luxuries budget. If this was not a design feature, might I suggest including it again as an option?

yagebu:
> Before b94a2c6, the category1 budget (which should accumulate $1.00 per day) would correctly go negative on 2017-01-01 and increase from there.

But generally, the previous behaviour was incorrect as comparing the budget to `balance_children` is invalid. Say you have a budget of 30 USD on `Expenses:Luxuries` and of 60 USD on `Expenses:Luxuries:Food` and on transaction of 50 USD on the Expenses:Luxuries:Food account in the relevant interval. Then this would have been shown as violating the budget on `Expenses:Luxuries` (as it would compare to the budget (30 USD) of just the parent account and not an accumulated budget of 90 USD).

So we need a "budget_children" to compare to `balance_children`. A PR adding a function that computes "budget_children" in `fava.core.budgets` would be welcome! (I can also add one at some point, but the budget functionality is low-priority for me as I don't use it)

TomJohnZ:
I see; that makes sense to me. I'll try implementing your `budget_children` suggestion.

Regarding implementation, do you think it would be better to have a global config option to change `budget` to `budget_children`, or would on additional option on the opening of each budget be preferrable?

yagebu:
No option or anything needed for this. Just like Beancount computes `balance_children` by considering all postings to either the account or its children we can compute `budget_children` by considering all budgets on either the account or its children (instead of just on the account itself).

TomJohnZ:
Sure, computing `budget_children` wouldn't need an option or anything. I was just thinking of the next step; after `budget_children` is implemented, it would need to present its data to `_tree_table.html` somehow, no? But that's the next step, so no need to worry about it right now, I suppose.

yagebu:
Getting it the template would work by duplicating [line 116](https://github.com/beancount/fava/blob/276b01695375cf147eeb1790d88d3e59bdee3185/fava/templates/_tree_table.html#L116) in `_tree_table.html` for `budget_children`.

TomJohnZ:
I think I'm still misunderstanding what you're envisioning. I was able to create a method `calculate_budget_children` in `budgets.py` that accumulates the budget of all children accounts; using your example above:

> Say you have a budget of 30 USD on Expenses:Luxuries and of 60 USD on Expenses:Luxuries:Food and on transaction of 50 USD on the Expenses:Luxuries:Food account in the relevant interval

This method would return a budget of 90 USD on `Expenses:Luxuries` for the relevant interval. Am I correct in assuming this is what you were describing? If so, I can commit the patch.

However, after writing the method and getting more familiar with Fava's data structures, it doesn't seem this will accomplish what I was attempting to convey above. Namely, rather than have parent accounts accumulate the children's *budgets* (as in the 90 USD budget above), they should instead accumulate the children's *postings*. That is, `render_budget` in `_tree_table.html` should be calculating the `diff` between the budget of the parent (like usual) and the position `pos` in all children (as opposed to just the parent).

I don't know if that was any clearer; I apologize for making this so difficult for you!

yagebu:
Just open a PR with the changes to `core.budgets` and I'll handle the template part.

TomJohnZ:
Ok, PR opened. Thanks for you help with all of this!

TomJohnZ:
@yagebu You can probably close this PR now, as its clear the "regression" wasn't actually a regression (whoops).

Howevr, I'll admit I'm still not 100% clear on how the method I added in the other PR will help. At risk of further clogging the PR queue, I submitted a sample of what I had in mind. I don't actually mean for it to be accepted, but I think a bit of working code will better convey what I meant than a vague explanation.

PR #519, opened 2017-03-30
Budget children

TomJohnZ:
Accumulates budget from parent accounts and children accounts.

### DIFF
--- a/fava/core/budgets.py
+++ b/fava/core/budgets.py
@@ -70,7 +70,7 @@ def _matching_budgets(budgets, account_name, date_active):
     """Find matching budgets.
 
     Returns:
-        The budget that is active on the specifed date for the
+        The budget that is active on the specified date for the
         specified account.
     """
     last_seen_budgets = {}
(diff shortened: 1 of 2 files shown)


yagebu:
Thank you for the PR! I'll try to get the template part sorted this weekend.

We've got some pretty strict linting set up and the lint errors should be fixed. See travis or run `make lint` to list the problems. You can also use [yap](https://github.com/google/yapf) to fix them automatically.

Also, some tests would be good.

TomJohnZ:
Not a problem! Oops, forgot to check for linting; it should be conforming now. (Sorry for the multiple lint commits; I can't run lint since pylint [doesn't support python3.6](https://github.com/PyCQA/pylint/issues/1113) yet, so I needed to do trial and error with Travis).

I'd be happy to make some unit tests. What about the documentation on the dev page? Is that compile automatically from the comment blocks, or should I submit something for that as well?

yagebu:
> What about the documentation on the dev page? Is that compile automatically from the comment blocks, or should I submit something for that as well?

The documentation get's compiled from the comment automatically (you can run `make docs` to generate them locally if you want to see the output)

yagebu:
Merged the test into #521, so closing this one.

PR #520, opened 2017-03-31
add global option for budget-children.

TomJohnZ:
Add global option causing budgets to accumulate positions from children


yagebu:
We should always recursively accumulate either both budgets and positions or none of the two.

PR #521, opened 2017-04-01
Budget children

yagebu:
Ref #519, #520.

### DIFF
--- a/fava/core/budgets.py
+++ b/fava/core/budgets.py
@@ -46,11 +46,14 @@ class BudgetModule(FavaModule):
 def parse_budgets(custom_entries):
     """Parse budget directives from custom entries.
 
+    Args:
+        custom_entries: the Custom entries to parse budgets from.
+
     Returns:
-        A tuple of Budget entries and possible parsing errors.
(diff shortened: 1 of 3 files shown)


TomJohnZ:
Thanks a lot for working on this @yagebu! I submitted a unit test in the other PR.

I just tried your commit on the MWE I posted earlier, but it looks like something's still off; now, neither the child_budget nor the posting_sum is displaying. My HTML is too rusty to fix it myself, but it seems that the `<span class="balance-children">` in line 137 is doing something weird; removing it (or changing it to `class="balance"`) seems to fix it.

TomJohnZ:
Ohh, I see. I wasn't aware that's how it worked. And I really like your changes in 65c1228, that's quite clever.

Thanks a lot for all your work on this. It was a pleasure working with you!

yagebu:
> I wasn't aware that's how it worked.

Yes it's not totally obvious so maybe we should add comments to the help pages about features like this.

> Thanks a lot for all your work on this.

You're welcome. Thank you for taking the initiative.

PR #522, opened 2017-04-11
settlement-date plugin

aumayr:
Quick and dirty implementation of the plugin mentioned in this thread: https://groups.google.com/forum/#!topic/beancount/d6SfLshoo1E

With this plugin, the following transaction:

```ruby
plugin "fava.plugins.settlement_date" "Assets:Savings:Transfer"

2017-04-01 * "" ""
    Assets:Savings:US    -100.00 USD
    Assets:Savings:JP
        settle: 2017-04-03
```

becomes

```ruby
    2017-04-01 * "" "Doing some saving transfers" ^settle-43be1c
        Assets:Savings:US       -100.00 USD
        Assets:Savings:Transfer
            settle: 2017-04-03

    2017-04-03 * "" "Settle: Doing some saving transfers" ^settle-43be1c
        Assets:Savings:Transfer -100.00 USD
        Assets:Savings:JP        100.00 USD
```

### DIFF
--- a/fava/plugins/settlement_date.py
+++ b/fava/plugins/settlement_date.py
@@ -1,0 +1,68 @@
+"""Beancount plugin to split transactions which are in transit.
+
+It looks through all Transaction entries with the `settlement-date`-metadata on
+one of it's postings and splits those into two transactions.
+
+Example:
+    plugin "fava.plugins.settlement_date" "Assets:Savings:Transfer"
+
+    2017-04-01 * "" ""
+        Assets:Savings:US    -100.00 USD
+            settle: 2017-04-03
+        Assets:Savings:JP
+
+    ; becomes
+
+    2017-04-01 * "" "Doing some saving transfers" ^settle-43be1c
+        Assets:Savings:US    -100.00 USD
+            settle: 2017-04-03
+        Assets:Savings:Transfer
+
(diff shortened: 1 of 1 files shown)


blais:
Nice!

yagebu:
Are you planning to get this PR into a mergeable state? (tests, lint, and from a quick look it seems that the code uses a different metadata key than the documentation).

aumayr:
> Are you planning to get this PR into a mergeable state?

I don't know if it should be part of Fava or not… what do you think?

yagebu:
> I don't know if it should be part of Fava or not… what do you think?

Not sure. We already ship two plugins, which are more tightly integrated with Fava somehow. But it also wouldn't do any harm to have one more. If we accumulate more plugins, it might make sense to move them out of the package though at some point.

aumayr:
> We already ship two plugins, which are more tightly integrated with Fava somehow

We should not include this one, then. It has virtually nothing todo with Fava, so this should live somewhere else. I'd like to keep the PR open until I have found a suitable home for it.

aumayr:
@yagebu Are you moving it over to https://github.com/beancount/fava-plugins?

yagebu:
@aumayr: I'll leave that to you since it needs more work - but yes, that's where it should go :)

Issue #448, opened 2017-01-02
Balance sheet display signs


knoxx093:
Propose to add options that will flip the signs in the balance sheet display.
Option to show liabilities as positive.
Option to show Equity with same sign as net-worth.
Perhaps Similar options for Income Statement display.

Reason:
It's driving me nuts that my negative net worth is showing up as a positive number.

blais:
On Mon, Jan 2, 2017 at 4:22 PM, knoxx093 <notifications@github.com> wrote:

> Propose to add options that will flip the signs in the balance sheet
> display.
>
That's a very good idea! I've had this on my todo list for bean-web for a
while.
In addition, when this is turned on, I'd recommend indicating to the user
that this feature is turned on by using a slightly different background
color or font for those numbers which have been "adjusted" to be normally
positive, so that one doesn't have to wonder whether it's on or not.


> Option to show liabilities as positive.
> Option to show Equity with same sign as net-worth.
> Perhaps Similar options for Income Statement display.
>
IMHO a single option for all numbers is sufficient.
The implementation is simple: all numbers are associated with an account
anyway: when a number is rendered, flip the sign if the account is either
of Liabilities, Equity or Income.



> Reason:
> It's driving me nuts that my negative net worth is showing up as a
> positive number.
>

knoxx093:
I think a single option would be equally helpful at least for my case. My initial reasoning for providing them separately is that it's apparently not uncommon to show liabilities as negatives. 
The suggestion to provide a visual indicator that this feature is turned on is a great idea as well.

If I had more time I'd attempt a PR for this but I unfortunately don't have much to spare now.

aumayr:
Good idea (initial idea and font-color)!

Maybe, when enabled, show the sign anyway but in a light-gray color so it is visually easier to distinguish (instead of hiding/flipping it completely)?

blais:
The problem with too many options is that it would quickly get confusing.
Don't forget contra accounts, which might make this even more confusing,
that is, accounts whose signs are opposite of the expected ones. A single
option means that there's only two possibilities: either the signs are
flipped, either they're not. If you flip only the Liabilities that would
make Income incoherent.

I'd advise against supporting the full flexibility of enabling arbitrary
flips, even if there's some usage for it somewhere (AFAIK you don't need
it, you mention that "it's not uncommon"; I'd ignore that). Keep it simple.
Start with a single option. See if that fulfills your immediate needs.

knoxx093:
[...]

PR #523, opened 2017-04-12
Option to invert balance sheet numbers

aumayr:
This PR implements #448:

A new option named `invert-balance-sheet` was added to invert all numbers on the balance sheet ;-) 

```ruby
2017-04-12 custom "fava-option" "invert-balance-sheet" "true"
```

With this option enabled, the balance sheet looks like this:

screenshot

\cc @knoxx093 @blais

### DIFF
--- a/fava/static/css/base.css
+++ b/fava/static/css/base.css
@@ -454,6 +454,13 @@ kbd {
 
   & h3 {
     text-align: center;
+
+    & .byline {
+        color: var(--color-background-darkest);
+        font-size: .9em;
+        font-weight: normal;
+        margin-left: .2em;
+    }
   }
 }
 
--- a/fava/static/css/base.css
+++ b/fava/static/css/base.css
@@ -456,10 +456,10 @@ kbd {
     text-align: center;
 
     & .byline {
-        color: var(--color-background-darkest);
-        font-size: .9em;
-        font-weight: normal;
-        margin-left: .2em;
+      color: var(--color-background-darkest);
+      font-size: .9em;
+      font-weight: normal;
+      margin-left: .2em;
     }
   }
 }
--- a/fava/core/fava_options.py
+++ b/fava/core/fava_options.py
@@ -30,7 +30,8 @@ DEFAULTS = {
     'journal-show-document': ['discovered', 'statement'],
     'language': None,
     'interval': 'month',
-    'incognito': False
+    'incognito': False,
+    'invert-balance-sheet': False,
 }
 
 BOOL_OPTS = [
@@ -42,6 +43,7 @@ BOOL_OPTS = [
     'show-accounts-with-zero-transactions',
     'show-closed-accounts',
     'use-external-editor',
+    'invert-balance-sheet',
 ]
 
 INT_OPTS = [
--- a/fava/docs/options.md
+++ b/fava/docs/options.md
@@ -184,3 +184,11 @@ entry, then the grey uptodate-indicator is shown.
 Default: `false`
  
 If set to `true` all digits will be replaced with "X".
+
+---
+ 
+## `invert-balance-sheet`
+ 
+Default: `false`
+ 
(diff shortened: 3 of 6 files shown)


fava/static/css/base.css
```diff
@@ -454,6 +454,13 @@ kbd {
 
   & h3 {
     text-align: center;
+
+    & .byline {
+      color: var(--color-background-darkest);
```
yagebu: Don't use a background color variable for a font color.

yagebu:
👍  This option should affect more reports (at least the income statement) and should hence have a different name.

knoxx093:
Going to agree with @yagebu about the other reports. 
I'd also like to remind that in the discussion for #448 the end goal was to flip signs displayed for Liabilities, Income, and Equity, not everything. See @blais first comment there.

aumayr:
I'm confused. 

@yagebu and @knoxx093 can you elaborate on what you exactly envision for this feature?

knoxx093:
The idea when I initially suggested the feature is to show 'credit normal'
accounts with a positive balance on reports. Since it's a bit counter
intuitive when a positive income or net worth value shows up as a negative
number due to it being a 'credit'.

aumayr:
> I'd also like to remind that in the discussion for #448 the end goal was to flip signs displayed for Liabilities, Income, and Equity, not everything. See @blais first comment there.

So only the "Total"-line should be inverted? *still confused*

knoxx093:
Only values for Income, Equity, and Liability accounts should be inverted.

aumayr:
> Only values for Income, Equity, and Liability accounts should be inverted.

So as in the screenshot above?

knoxx093:
In that screenshot it looks like you also have assets inverted since they
are all negative

On Apr 12, 2017 12:13 PM, "Dominik Aumayr" <notifications@github.com> wrote:

Only values for Income, Equity, and Liability accounts should be inverted.

So as in the screenshot above?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/beancount/fava/pull/523#issuecomment-293646501>, or mute
the thread

aumayr:
Ahh now I got you - I misread for "all balance sheet accounts", and did not get "Income, Equity, and Liability accounts". 

@yagebu Apart from the Balance Sheet and Income Statement, which reports should be affected as well?

yagebu:
> Apart from the Balance Sheet and Income Statement, which reports should be affected as well?

All tree tables basically for consistency.

@knoxx093: Do you have a suggestion for the name of this option?

knoxx093:
I'm terrible at naming. 'show_credit_normal_as_positive' is the best I can
think of.

yagebu:
@aumayr: Are you planning to continue with this?

aumayr:
> @aumayr: Are you planning to continue with this?

I won't have time to continue this for 1.4, so if you think it should be part of the next release please have a go at it. In 2-3 weeks time I'll have time again to work on it.

yagebu:
> I won't have time to continue this for 1.4

This is certainly not a blocker, so take your time ;)

aumayr:
>  In 2-3 weeks time I'll have time again to work on it.

Ups, that was more than 2-3 weeks 👼 

> All tree tables basically for consistency.

_All_ treetables? Because if "yes", then this is not a big change to implement I think.

yagebu:
> All treetables?

In all places, where an amount belongs to some account which is a subaccount of Income, Equity, or Liabilities, that amount should be multiplied by -1. So in the treetables, but also in the corresponding charts and the account journal.

yagebu:
Since this has gotten quite stale and unmergeable (and there's not much useful code here for the proper implementation), I'll close this PR

aumayr:
> Since this has gotten quite stale and unmergeable

Sorry for dropping the ball on this.

Issue #481, opened 2017-01-17
Better transaction-insertion


yagebu:
Currently, Fava determines where to insert a transaction in a less-than-ideal way. It abuses the insert marker, which is really only intended for the editor as it's a per-file setting whereas transactions are always inserted globally. Also, it doesn't allow one to insert transactions for different accounts in different places.

One way I've thought we could solve this would be to provide a new option, say `insert`, which can be a regular expression for an account name. When inserting a transaction, Fava would then go through the postings in reverse order<sup>[1](#footnote-1)</sup>, and try to match the `insert` regular expressions.

For example if I had two insert fava-options, one with `Assets:Cash`, the other with `Liabilities:CreditCard`, transactions for those two accounts would be inserted above the respective options.

I don't know how others order transactions in their files, would this approach be flexible enough? Alternatively we could try to determine the file positions completely algorithmically, but I don't see an easy way that's flexible enough to do it.

1<a name="footnote-1"></a>: I don't know if this is a convention for Beancount too, but in ledger the "source" account is contained in the last posting, which for me at least is the most relevant account for determining the position.

aumayr:
> I don't know how others order transactions in their files, would this approach be flexible enough? 

I also sort the postings this way. What I don't quite get is how this option would look like - could you provide a short example? Is this a comment like the current insert marker?

If this is a `custom` directive, then it would have a date, which is not really relevant (and quite confusing), as the _position_ in the file would be relevant, where otherwise only the _date_ of the entry is relevant (one of the positive aspects of Beancount, IMHO, is that the ordering in the file does not matter).

adamgibbins:
I place all my transactions in one file, until I get around to automating the import of that account at which point I move it out into its own (as it's easier to compare then with new imports etc). I'm unlikely to ever use the add transaction feature to add transactions in their final resting place - it's likely to be "quick, get it recorded" type entries which I'll then follow up on and move before committing. So I'm perfectly happy with having the ability to write to a scrap/scratch file or similar.

yagebu:
@aumayr: I'd imagine this as a `fava-option`. Since the date is irrelevant for *all* our options and we're already using the position in the file for the `default-file` option, I don't see why this would be an issue here in particular. The following example would lead to all transactions to subaccounts of `Assets:Cash` being inserted above the option (if that's the first regex they match).

    2000-01-01 custom "fava-option" "insert" "Assets:Cash:.*"

ngoonee:
[...]

PR #530, opened 2017-04-17
Transaction insertion

yagebu:
Ref #481


aumayr:
👍🏻

yagebu:
> Request (if possible) that if the last posting of a transaction does not match, the second-last (all the way to the front) be checked as well (in that way, there's a priority order).

👍

ngoonee:
Works for me. Can't wait for this to be merged =)

PR #531, opened 2017-04-17
Conversion to a single currency

yagebu:
Ref #242

### DIFF
--- a/fava/application.py
+++ b/fava/application.py
@@ -196,7 +196,7 @@ def _pull_beancount_file(_, values):
     if g.beancount_file_slug not in app.config['FILE_SLUGS']:
         abort(404)
     g.ledger = app.config['LEDGERS'][g.beancount_file_slug]
-    g.at_value = request.args.get('conversion') == 'at_value'
+    g.conversion = request.args.get('conversion')
 
 
 @app.errorhandler(FavaAPIException)
--- a/fava/core/charts.py
+++ b/fava/core/charts.py
@@ -11,9 +11,9 @@ from beancount.core.inventory import Inventory
 from beancount.core.data import iter_entry_dates
 from beancount.utils.misc_utils import filter_type
 from flask.json import JSONEncoder
-from flask import g
 
 from fava.util import listify, pairwise
+from fava.template_filters import cost_or_value
 from fava.core.helpers import FavaModule
 
 
@@ -44,13 +44,9 @@ def _inventory_units(inventory):
     }
 
 
-def _inventory_cost_or_value(inventory, date=None):
+def _inventory_cost_or_value(inventory, date):
     """Renders an inventory at cost or value to a currency -> amount dict."""
-    if g.at_value:
-        inventory = inventory.reduce(convert.get_value, g.ledger.price_map,
-                                     date)
-    else:
-        inventory = inventory.reduce(convert.get_cost)
+    inventory = cost_or_value(inventory, date)
     return {p.units.currency: p.units.number for p in inventory}
 
 
--- a/fava/template_filters.py
+++ b/fava/template_filters.py
@@ -33,8 +33,11 @@ def cost(inventory):
 
 def cost_or_value(inventory, date=None):
     """Get the cost or value of an inventory."""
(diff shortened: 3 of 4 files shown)


aumayr:
👍🏻

xentac:
This isn't working quite the way I thought it would...

Most charts keep displaying only my operating currencies... though choosing to convert to one of them seems to work ok.

If I'm looking at an expense account that has both USD and CAD transactions, the Account Balance graph doesn't change, the Changes graph changes (only if I select one of the two operating curriences), but the Changes graphs doesn't change if I choose a non-operating currency.

None of the transactions are converted when displayed in the journal, so they always show their original currency.

I also see some weird entries in the Converted to dropdown.  Specifically I'm seeing some things listed that aren't commodities.  A handful of transactions I add document metadata to, representing an identifier from a pdf.  Those values also show up in the dropdown...

Shall I file bugs for these or do you have easy fixes for them?  I may be able to work on some of them, but it'll be slowly over the next few weeks.

aumayr:
> Shall I file bugs for these or do you have easy fixes for them? I may be able to work on some of them, but it'll be slowly over the next few weeks.

@xentac Yes please do file bugs if you find some! If you start working on them please assign yourself to them in the Github issue, so there is no double-work.

yagebu:
Thanks for the feedback @xentac!

> Most charts keep displaying only my operating currencies... though choosing to convert to one of them seems to work ok.

The bar charts currently only display operating currencies, I guess if one converts to another currency that should also be displayed.

> If I'm looking at an expense account that has both USD and CAD transactions, the Account Balance graph doesn't change, the Changes graph changes (only if I select one of the two operating curriences), but the Changes graphs doesn't change if I choose a non-operating currency

Currently the conversion only applies to amounts that are normally displayed at cost. I do agree that when selecting "convert" one would probably expect that the account balance in the chart and the account journal is also converted. This would introduce the following (probably acceptable) inconsistency: When "At Cost" is selected, the units of the account balance are shown so the option doesn't seem to apply but with, say, "Converted to USD" it *would* apply. 

>None of the transactions are converted when displayed in the journal, so they always show their original currency.

As the journal is just a fancy rendering of the input file, the numbers there shouldn't be converted IMHO. In the account journal however the above applies to the change and balance columns.

> I also see some weird entries in the Converted to dropdown. Specifically I'm seeing some things listed that aren't commodities. A handful of transactions I add document metadata to, representing an identifier from a pdf. Those values also show up in the dropdown...

This is either a problem with your input or Beancount. What kind of entries make these unexpected currencies show up?

xentac:
> This is either a problem with your input or Beancount. What kind of entries make these unexpected currencies show up?

It's sort of easy to find the exact transaction because there is a single reference to the "currencies" in my file.

```
2015-01-29 * "S04128836C"
  document: RSUA148148
  Assets:US:Schwab:HOOL 3 HOOL {9 USD}
  Income:US:Hooli:RSU -45.00 USD
  Assets:Hooli:RSURefund
  Expenses:Taxes:TY2015:US:Medicare 5.00 USD
  Expenses:Taxes:TY2015:US:Federal 6.00 USD
  Expenses:Taxes:TY2015:US:SocSec 4.00 USD
```

This is a simplified transaction that probably also has the problem.

yagebu:
Strings need to be quoted.

xentac:
Ha! Thanks for that.

xentac:
> Currently the conversion only applies to amounts that are normally displayed at cost. I do agree that when selecting "convert" one would probably expect that the account balance in the chart and the account journal is also converted. This would introduce the following (probably acceptable) inconsistency: When "At Cost" is selected, the units of the account balance are shown so the option doesn't seem to apply but with, say, "Converted to USD" it would apply.

I'm not sure I understand what you mean by "the units of the account balance are shown".  Don't the graphs and balances always show units?

> As the journal is just a fancy rendering of the input file, the numbers there shouldn't be converted IMHO. In the account journal however the above applies to the change and balance columns.

I was talking about the Balance and Change columns.  I agree that the detailed journal rows should be untouched.

PR #540, opened 2017-04-30
Fix Min/Max LineChart values and add two more tree levels

scanta2:
Set LineChart min/max to time period values.
Add two more account tree levels.

### DIFF
--- a/fava/static/css/tree-table.css
+++ b/fava/static/css/tree-table.css
@@ -72,6 +72,18 @@
       min-width: 7em;
     }
 
+    &.depth-8 {
+      margin-left: 8em;
+      max-width: 22em;
+      min-width: 6em;
+    }
+
+    &.depth-9 {
+      margin-left: 9em;
+      max-width: 21em;
+      min-width: 5em;
+    }
+
     & a {
       margin-left: 1em;
     }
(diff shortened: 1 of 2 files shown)


yagebu:
Thanks for the PR. Will add two account levels.

I usually find charts that don't include the x-axis harder to understand so I would prefer to keep it like it is, in which case the precision of 2 for the y-axis labels is also enough. Also I find 2 digits after the decimal point to be a better default precision for the tooltips.

scanta2:
I probably explained myself wrong. This is about the y-axis. It is set to the min/max value of the account during that time period. I have had issues with 2 digits in the precision in this case, because the y-axis span could be much shorter.

yagebu:
> I probably explained myself wrong. This is about the y-axis.

I think I understood correctly. My comment was maybe slightly imprecise. By x-axis I mean the line at `y = 0`, which I think the chart should include.

scanta2:
Got it. Would it be possible to add an option for that? It makes it really tough for me to look at daily variations in my net worth graph when looking at shorter time periods. Maybe an option to set min/max y or fallback to the default?

yagebu:
I feel like an option for this is a bit overkill (and also not the best from a usability standpoint). It wouldn't really solve the problem in general either, if the chart globally has a high variation but one is maybe interested in a small local change.

So instead making the chart zoomable would be the better approach here IMHO.

aumayr:
I'm also for the zoom-solution. 

The current zoom-state could be added to the URL-hash, so already-zoomed charts would be bookmark-able.

yagebu:
Closing the PR as the account levels change has been merged already and the charts change won't be. Any work on zoomable charts is welcome!

PR #554, opened 2017-06-04
Tag-filter: Exclude tags and links by filtering for `-#tag`

aumayr:
Adds the ability to exclude certain tags for filtering for `-#tag` and certain links by filtering for `-#link`.


yagebu:
While it makes sense that `#tag` only includes entries with that tag (and hence automatically excludes all entry types that have no tags), it is surprising (and probably wrong) that `-#tag` excludes entries without tags.

PR #561, opened 2017-06-26
Hide context overlay when clicking on the Location link in it

### LINKED ISSUE #560: Context-overlay: Clicking the Location link does not close modal overlay
When clicking on the date in a Journal row, the context-overlay is shown. 

When clicking on the Location-link in the context-overlay, the editor is loaded, but the context-overlay stays, and the content of it is inserted a second (and third, etc.) time. Ideally, when clicking the Location-link, the editor is loaded, and the overlay disappears.

aumayr:
Fix #560

### DIFF
--- a/fava/static/javascript/router.js
+++ b/fava/static/javascript/router.js
@@ -76,6 +76,10 @@ function handleHash() {
       .then((data) => {
         $('#context-overlay .content').innerHTML = data;
         initReadOnlyEditors();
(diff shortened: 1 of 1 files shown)


yagebu:
This should be solved more generally in the router IMHO: Clicking on any link should close (and empty) all overlays.

aumayr:
Will be fixed in #563

Issue #534, opened 2017-04-23
Suggestions for Editor view


ngoonee:
A few suggestions for the Editor:-

1. Start (maybe optionally) with all folds folded.

2. Close folds with any click on the number line rather than only on the head/title of fold.

3. When clicking save, check whether beancount file has changed in the meantime.

Number 1 should be self explanatory, much more convenient for dealing with large files.

Number 2 eases navigation in large files.

Number 3 is an attempt to deal with the asynchronous access issue. As it is if youre in the Editor page and use the transaction entry popup to enter a transaction, the editor page still contains the old version of the file. Can also be triggered if using the editor in one tab while entering transactions in another. This can result in data loss without warning. Perhaps a md5sum of the file can be saved on opening, and on saving check the md5sum of the file currently on disc against that saved md5sum. If not matching, provide option to override, to not save, or to save as another file name.

aumayr:
Number 3 is definitely something we should add - this happens sometimes, and is frustrating each time.

yagebu:
All good suggestions! Any work on this would be welcome :)

yagebu:
I'm working on this.

yagebu:
1. and 3. are addressed IMHO:

1.  "fold all" and "unfold all" are available from the menu. However they are quite slow for large files so I'm against starting with all folds closed as that makes the page lock up for a few seconds (on my file).

3. Saving will now fail if the source file was changed externally.

I had a look at 2. but didn't find a good way to do it. I'll not spend more time on it myself so I'm closing this issue, but if anyone wants to work on it, feel free!

Issue #456, opened 2017-01-03
Add ability to clear transactions


adamgibbins:
When I import transactions I guess the second leg of the transaction and mark it with a ! flag until I've confirmed it. I generally find browsing the fava transaction screens to be more visually pleasing and efficient than the raw beancount files so I do my reviews in fava, It'd be nice if I could just click the flag to change its status so I don't have to continually alt-tab or click through to the editor + refresh on 100s of transactions.

aumayr:
This seems like a useful feature, but a "hidden" one, because now clicking on the row (and with that the little dots) just expands the row. Maybe an alt-click or ctrl-click could do that. 

More broadly, this also goes into the direction of changing the beancount file directly, which seems to be the direction many of Fava's new features are going. 

@adamgibbins Would you prepare a PR for this?

adamgibbins:
On second thought it'd be pretty nice to make it a keyboard shortcut, that'd keep it out the way cluttering the interface more too.

@aumayr I'd love to, but have written ~100 lines of python ever, and no JS, so it'd take me some while and probably be terrible. But if you can point me in some direction(s) I'd be happy to take a stab.

aumayr:
As this requires the direct manipulation of the Beancount file, some thought has to go into how to approach this (the add-transaction-feature just appends text, but does not change it). 

@yagebu Do you have any thoughts on how to approach this?

yagebu:
#563 should make changing the flag much quicker than before. I don't to edit the flag very often so I won't implement an even more direct way to edit them.

GuillaumeHM:
This subject is still open at  #1151.

PR #563, opened 2017-06-27
Entry editor

yagebu:
- Use file hash to check that edited file is not changed externally.
- Replace the context overlay with an editor for the lines of the entry.

Ref #534; Somewhat related to #456.

### DIFF
--- a/fava/core/__init__.py
+++ b/fava/core/__init__.py
@@ -5,7 +5,7 @@ import datetime
 import os
 
 from beancount import loader
-from beancount.core import getters, prices, realization
+from beancount.core import getters, interpolate, prices, realization
 from beancount.core.flags import FLAG_UNREALIZED
 from beancount.core.account_types import get_account_sign
 from beancount.core.compare import hash_entry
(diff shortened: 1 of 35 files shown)


aumayr:
Looks awesome and works as advertised! And makes a huge difference in my workflow, as I very often go to a specific transaction in the editor through the Context page. 

One question about UX: Should a successfull "Save" close the overlay?

yagebu:
> One question about UX: Should a successfull "Save" close the overlay?

Yes that is the plan, will go together with some reworking of the overlay-closing logic.

aumayr:
Bug: The z-index of the account-suggestion-overlay when typing in the inline-editor is behind the context-overlay.

aumayr:
Let's bake a new release?

yagebu:
Sounds good - although I would give it a few days to see if some bugs turn up. I pushed new translateable strings to POEditor and notified the contributors there - let's also give them a few days ;)

aumayr:
> Sounds good - although I would give it a few days to see if some bugs turn up.

I did use this feature somewhat in the last week, and no big issues on my side.

Issue #573, opened 2017-07-30
Add --version/-v flag that prints the version


mhansen:
A neat little thing that'll help me figure what version I'm running. Would probably help when folks file bugs too, to help them easily find out which version they're running.

aumayr:
Fixed in #574

PR #574, opened 2017-07-30
Add --version flag support.

mhansen:
Click already had an implementation.

It looks like:
$ python3 fava/cli.py --version
fava, version 1.6-dev

### DIFF
--- a/fava/cli.py
+++ b/fava/cli.py
@@ -8,6 +8,7 @@ from werkzeug.wsgi import DispatcherMiddleware
 
 from fava.application import app, load_file
 from fava.util import simple_wsgi
+from fava import __version__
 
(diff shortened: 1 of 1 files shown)


mhansen:
Fixes #573

aumayr:
Thanks @mhansen for the PR!

This works as advertised, and is a very welcome addition! The only thing holding me back from merging is that the tests on Travis fail. Changing 

```python
from __init__ import __version__
```
to 
```python
from fava import __version__
```
does fix these issues. Can you incorporate this change?

mhansen:
Done, thanks very much!

aumayr:
Thank you, @mhansen!

PR #575, opened 2017-08-07
Short version of quarter and month range of current year

Darion:
Now, if you want to specify week or quarter period of current year - you don't need to specify a year

### DIFF
--- a/fava/docs/features.md
+++ b/fava/docs/features.md
@@ -9,9 +9,9 @@ There are four filters available in Fava:
   Beancount Query Language.
 - **Time**: Filter entries by their date. You can specify dates and intervals
   like years, quarters, months, weeks, and days (for example `2015`, `2012-Q1`,
-  `2010-10`, `2016-W12`, or `2015-06-12`). You can specify a range of dates like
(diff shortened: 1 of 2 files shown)


yagebu:
Hey, thanks for working on Fava!

Are you aware that specifying a week or quarter of the current year is already possible in Fava, e.g. `year-W10`? One problem with extending the syntax like you suggest is that it would make it ambiguous: `year-W10` could mean an interval (from the start of the year to the 10th week) or a single week (the 10th week of the current year).

Darion:
No, I didn't now that, thanks. But `year-W10` is not much easier to type than `2017-W10`. Can we make things like `W10` or `Q2` (outside of period expressions) something like special forms of syntax, since queries with week or quarter of current year must be very popular?

yagebu:
Sorry for the late response.

If it can't be used inside date ranges, that would be a bit annoying and wouldn't seem that useful to me. I'd really much prefer it if it could be made unambiguous somehow. Any suggestions?

Darion:
Maybe we can add possibility to set user shortcuts for filters. But this is for another pull request, we can close this, I think.

PR #580, opened 2017-09-09
Reduce change and balance counters when creating journal table

xentac:
Instead of potentially copying the same objects many times on every iteration. It turns out that this one weird trick speeds up journals with large balances of commodities because it saves copying many Inventory objects.


yagebu:
Thanks, that's a good find!

PR #583, opened 2017-09-11
Faster balance sheets

xentac:
Two improvements, when mixed with https://bitbucket.org/blais/beancount/issues/191/beancountcoreinventory-should-be-a-dict, reduce my balance sheet computation time from over 3 seconds to around 1 second.

### DIFF
--- a/fava/core/charts.py
+++ b/fava/core/charts.py
@@ -1,11 +1,12 @@
 """Provide data suitable for Fava's charts. """
 import datetime
+import collections
 
 from beancount.core import flags, convert, realization
 from beancount.core.amount import Amount
 from beancount.core.data import Transaction, Cost
 from beancount.core.display_context import DisplayContext
-from beancount.core.number import Decimal, MISSING
+from beancount.core.number import Decimal, MISSING, ZERO
 from beancount.core.position import Position
 from beancount.core.inventory import Inventory
 from beancount.core.data import iter_entry_dates
@@ -50,20 +51,13 @@ def _inventory_cost_or_value(inventory, date):
     return {p.units.currency: p.units.number for p in inventory}
 
 
-def _inventory_combine(inv1, inv2):
-    """Combine two inventory dicts adding numbers together -> amount dict."""
-    keys = set(inv1.keys()) | set(inv2.keys())
-    return {currency: inv1.get(currency, 0) + inv2.get(currency, 0)
-            for currency in keys}
-
-
 def _serialize_real_account(real_account, date):
     children = [_serialize_real_account(account, date)
                 for _, account in sorted(real_account.items())]
-    balance_children = _inventory_cost_or_value(real_account.balance, date)
+    balance_children = collections.Counter(_inventory_cost_or_value(
+        real_account.balance, date))
     for child in children:
-        balance_children = _inventory_combine(balance_children,
-                                              child['balance_children'])
+        balance_children.update(child['balance_children'])
     return {
         'account': real_account.account,
         'balance_children': balance_children,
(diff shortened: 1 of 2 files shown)


fava/core/charts.py
```diff
@@ -165,11 +165,16 @@ def net_worth(self, interval):
         txn = next(transactions, None)
         inventory = Inventory()
 
+        today = datetime.datetime.today().replace(
```
yagebu: This should be a date and if the date is irrelevant, use a constant like `date.min`

fava/core/charts.py
```diff
@@ -165,11 +165,16 @@ def net_worth(self, interval):
...
+            hour=0, minute=0, second=0, microsecond=0)
         for date in self.ledger.interval_ends(interval):
             while txn and txn.date < date:
                 for posting in filter(lambda p: p.account.startswith(types),
                                       txn.postings):
-                    inventory.add_position(posting)
+                    inventory.add_amount(
+                        posting.units,
+                        Cost(0, posting.cost.currency, today, None)
```
yagebu: The 0 should be a `Decimal`, so use the constant from `beancount.core.number`

fava/core/charts.py
```diff
@@ -165,11 +165,16 @@ def net_worth(self, interval):
...
         inventory = Inventory()
 
+        today = datetime.datetime.today().replace(
+            hour=0, minute=0, second=0, microsecond=0)
         for date in self.ledger.interval_ends(interval):
             while txn and txn.date < date:
                 for posting in filter(lambda p: p.account.startswith(types),
                                       txn.postings):
-                    inventory.add_position(posting)
+                    inventory.add_amount(
```
yagebu: Maybe a short comment here why we do the calculations in this slightly odd way would be good.

fava/core/charts.py
```diff
@@ -50,16 +50,25 @@ def _inventory_cost_or_value(inventory, date):
...
+    """Combine two inventory dicts adding numbers together -> amount dict."""
+    keys = set(inv1.keys()) | set(inv2.keys())
+    return {currency: inv1.get(currency, 0) + inv2.get(currency, 0)
+            for currency in keys}
+
+
 def _serialize_real_account(real_account, date):
+    children = [_serialize_real_account(account, date)
+                for _, account in sorted(real_account.items())]
+    balance_children = _inventory_cost_or_value(real_account.balance, date)
```
yagebu: If we use a `collections.Counter` here, this code should get much simpler

yagebu:
It would be good to add tests for the functions that were changed.

xentac:
I'll try to write tests, but here's the other requested changes.

fava/core/charts.py
xentac: Whoops, you're right. Done.

fava/core/charts.py
xentac: Done.

fava/core/charts.py
xentac: Done.

fava/core/charts.py
xentac: Neat, I didn't even know that existed.

xentac:
Tests added.

yagebu:
Merged with some minor changes, thanks 👍 

For me, this PR doesn't change the performance of the balance sheets very much, the improvement is at most 5-10%.

xentac:
For reference, the reason mine is so large is because I started using Betterment in 2015 and contributed to it every other week. My Holdings report has 2005 rows in it...

PR #585, opened 2017-09-12
Budgets example

yagebu:
Also, commit two more generated files in git and update pythonanywhere config to use fixed example files.

@aumayr: This will require updating the page on pythonanywhere. Is there some way to grant me access to fava project/account on pythonanywhere?


aumayr:
@yagebu I can give you the credentials - which "delivery method" do you prefer? Email?

yagebu:
> Email?

👍

PR #595, opened 2017-10-02
Export current Journal view as Beancount file

aumayr:
This PR renders a new button in the Journal view, that enables the user to download the entries in the current Journal view (with applied filters, etc.) as a Beancount file.

This is very helpful if the user wants to transfer a certain "slice" of his/her Journal entries over to another Beancount instance, or for export/import workflows.

(I can elaborate on my use case if someone is interested.)


fava/core/__init__.py
```diff
@@ -355,6 +355,47 @@ def context(self, entry_hash):
         source_slice, sha256sum = get_entry_slice(entry)
         return entry, balances, source_slice, sha256sum
 
+    def render_journal(self):
```
yagebu: There's no need for another function on FavaLedger. This can be more cleanly done with a template.

aumayr: Changed this to a template in 48c3239

fava/templates/_journal_table.html
```diff
@@ -29,24 +29,32 @@
...
-        <button type="button" title="{{ doc_type.2 }}" data-type="{{ doc_type.0 }}" data-key="{{ keyboard_shortcuts[doc_type.0] }}" class="small doc-toggle{% if doc_type not in g.journal_show or 'document' not in g.journal_show %} inactive{% endif %}">{{ doc_type.1 }}</button>
-      {% endfor %}
-    {% endif %}
-  {% endfor %}
-  <button type="button" title="Toggle metadata" data-type="metadata" data-key="m"{% if 'metadata' not in g.journal_show %} class="inactive"{% endif %}>{{ _('Metadata') }}</button>
-  <button type="button" title="Toggle postings" data-type="postings" data-key="p"{% if 'postings' not in g.journal_show %} class="inactive"{% endif %}>{{ _('Postings') }}</button>
-</form>
+<div class="journal-menu">
+  <form action="{{ url_for('download_journal') }}">
+    <button type="submit" title="Export">Export</button>
```
yagebu: This isn't really a button but a link, you can use an `<a>` with the button class.

fava/template_filters.py
```diff
@@ -145,3 +145,12 @@ def uptodate_eligible(account_name):
...
         'fava-uptodate-indication') == 'True'
+
+
+def rendered_entries(entries):
+    """Render the entries provided."""
+    from fava.core.file import get_entry_slice
+
+    return [get_entry_slice(entry)[0] for entry in entries
+            if isinstance(entry, (data.Balance, data.Transaction))
+            and entry.meta['filename'] != '<summarize>']
```
yagebu: I think exluding transactions by flag is more robust - I'm not sure which should be excluded. Maybe a flag whitelist is better even so because this won't work for generated entries.

fava/template_filters.py
```diff
@@ -145,3 +145,12 @@ def uptodate_eligible(account_name):
     """Determine whether uptodate-indicators should be shown for an account."""
     return g.ledger.account_metadata(account_name).get(
         'fava-uptodate-indication') == 'True'
+
+
+def rendered_entries(entries):
+    """Render the entries provided."""
+    from fava.core.file import get_entry_slice
+
+    return [get_entry_slice(entry)[0] for entry in entries
```
yagebu: Can't this be done in the template (maybe a method needs to be added on FileModule.

yagebu:
A test and some info about this on the help pages would be good.

aumayr:
> A test and some info about this on the help pages would be good.

cdf7b18

fava/template_filters.py
aumayr: 628813f

fava/templates/_journal_table.html
aumayr: 8e92658

fava/core/file.py
```diff
@@ -110,6 +110,24 @@ def insert_entries(self, entries):
...
+        Only renders entries that are in the original Beancount file(s).
+
+        Args:
+            entries: A list of entries.
+
+        Returns:
+            A string representing the entries rendered in Beancount format.
+
+        """
+        entries = [get_entry_slice(entry)[0] for entry in entries
```
yagebu: This might fail if the entry is plugin-generated. We could check two lines below if `entry.meta.get('filename')` is one of the files in the ledger.

Or maybe we would want to include plugin-generated entries too? Depending on the setup simply ignoring those entries on export might lead to a useless export.

aumayr: > Or maybe we would want to include plugin-generated entries too?

I think we want - but I have no idea how to do that, as right now `get_entry_slice` looks for the "source" in the beancount file. In the case of plugin-generated entries we would have to render an entry from the Python object.

aumayr:
@yagebu I want to get this in before the next release. Can you review the changes again?

fava/core/file.py
yagebu: Sorry, somehow missed this comment. You can use `_format_entry` for entries where get_entry_slice fails. Find out with which exceptions it can fail and wrap it in a try-except block. (Maybe that's easier to do if this function for-loops over the entries `yield`s entry slices and the template joins them)

aumayr: I implemented this in a rudimentary way, ignoring "Summarize"-entries (because I don't want to have them).

fava/core/file.py
```diff
@@ -110,6 +110,28 @@ def insert_entries(self, entries):
...
+        Args:
+            entries: A list of entries.
+
+        Yields:
+            The entries rendered in Beancount format.
+
+        """
+        for entry in entries:
+            if isinstance(entry, (data.Balance, data.Transaction)):
+                if isinstance(entry, data.Transaction) and entry.flag == 'S':
```
yagebu: The 'C' flag should also be excluded and you should check what other flags Beancount uses for generated entries.

aumayr: 6369f70

fava/core/file.py
```diff
@@ -123,9 +123,20 @@ def render_entries(self, entries):
             The entries rendered in Beancount format.
 
         """
+        excl_flags = [
+            flags.FLAG_PADDING,      # P
```
yagebu: If the generated padding transactions are not exported, maybe we should export the Padding directives themselves too in addition to transactions and balances?

fava/templates/_journal_table.html
```diff
@@ -29,24 +29,27 @@
...
-      {% for doc_type in document_type_buttons %}
-        <button type="button" title="{{ doc_type.2 }}" data-type="{{ doc_type.0 }}" data-key="{{ keyboard_shortcuts[doc_type.0] }}" class="small doc-toggle{% if doc_type not in g.journal_show or 'document' not in g.journal_show %} inactive{% endif %}">{{ doc_type.1 }}</button>
-      {% endfor %}
-    {% endif %}
-  {% endfor %}
-  <button type="button" title="Toggle metadata" data-type="metadata" data-key="m"{% if 'metadata' not in g.journal_show %} class="inactive"{% endif %}>{{ _('Metadata') }}</button>
-  <button type="button" title="Toggle postings" data-type="postings" data-key="p"{% if 'postings' not in g.journal_show %} class="inactive"{% endif %}>{{ _('Postings') }}</button>
-</form>
+<div class="journal-menu">
+  <a href="{{ url_for('download_journal', account=account_name) if account_name else url_for('download_journal') }}" class="button" title="{{ _('Export Journal as a Beancount file') }}" data-remote>{{ _('Export') }}</a>
```
yagebu: I think this would be slightly more cleanly solved if the `<a>` just went in the form. Will do this on merge, noticed some unused CSS code related to this too.

yagebu:
@aumayr: can you finish this soon? I'd like to release 1.6 today, we could also push this to 1.7 (which should be in a month)

fava/core/file.py
aumayr: Would that mean that for each account used in a balance directive in the export, we add a pad directive, like `2000-05-28 pad Assets:CA:BofA:Checking Equity:Opening-Balances`?

yagebu: Mhh not sure actually - we should probably do that if we want the exported file to be a self-contained Beancount file without errors. Currently the balance directive in the exported file are a bit awkward since they probably won't balance on their own with some subset of the entries. Also, re could export pad directives if they already exist in the source.

aumayr: True the file would not be "self-contained" regardless, as the user can select any filters and therefore display a subset of the Journal that cannot possibly be "self-contained".

My use-case here is that I export a subset of my journal (Liabilities) and wrote an Importer for my girlfriend's fava, so she can import "my liabilities", and vice versa, as we have two separate journals.

So I guess we cannot implement this sensibly, and keep it the way it works now, until a use case turns up that would need this strict "self-containment".

Issue #589, opened 2017-09-18
Book value vs Market value in account trees


yagebu:
In the discussion on #588, the question of whether we should display balances in the account trees at cost (as we currently) or at market value came up. The reports in Fava go back to the reports in `bean-web` and I don't think there's been any real discussion on this issue so far. Let's take the following example entries.

    2017-01-01 * "Invest"
      Assets:Checking
      Assets:Investment 10 HOOLI {5 USD}

    2017-02-01 price HOOLI 6 USD

 As far as I can see we could display the balances (of `Assets:Investment`) in the following way:

1. Display them at the historical cost (`50 USD`). This is what we currently do (when "At Cost" is selected in the dropdown).
1. Display them at market value (`60 USD`) and post the difference to `Equity:UnrealizedGains` or something.
1. Display them at market value (`60 USD`). This is currently possible by selecting "At Market Value" in the dropdown.
1. Display the units (`5 HOOLI`).

I do quite like 1. as a default. 1. is maybe closest to the traditional accounting approach as it's following the historical cost principle (I'm not an expert on accounting, so this should be taken with a grain of salt of course).  2. is also useful and it's currently not possible to get this sort of report in Fava directly, but the unrealized gains plugin does offer this (however it only adds unrealized gains once at the end of all entries not for any possible filtered interval).

Having 3. or 4. as a default would be quite far from the typical notion of a balance sheet. 3. is already possible and has a simple implementation as nothing has to be moved around to balance things out. The holdings page somewhat handles 4. but maybe it would also be nice to have it in the account tree.

Which of these options should Fava offer and which should be default?

Pinging and quoting @blais from the other discussion:

> I would argue for displaying them at market by default,
and to add an option (a toggle?) to display at book value.

> [..] I'd have reported at market value and added a tax
liability to account for taxes on the unrealized profit, so that the
balance reflect the value on liquidation.

xentac:
Option 2 is something that I tried to better support with unrealized_periodic (https://github.com/xentac/beancount-plugins-xentac/blob/master/beancount_plugins_xentac/plugins/unrealized_periodic.py) but it broke net worth calculation.

The correct way to do it is add unrealized transactions every time a Price shows up in the entry table, then use that list instead of the unmodified one when building the tree and journal (but not the net worth table). I suppose, on load, we could build the market value entry list and store it as another attribute in FavaLedger. From there we would remove the current At Market Value query and instead just use the different list of entries.

[...]

PR #597, opened 2017-10-06
Tree table optimizations; show unrealized gain

yagebu:
First, this contains some improvements to the balance trees are calculated. Since Beancount's inventories are currently based on lists, this helps in particular for balances containing lots of holdings at cost.

Secondly, if "At Value" is selected, the unrealized gains will be shown next to the market value (similar to how budgets are displayed). This looks as follows (I guess the cell widths should be increased):

screenshot-2017-10-6 balance sheet - huge example file

Ref #589

### DIFF
--- a/fava/core/tree.py
+++ b/fava/core/tree.py
@@ -1,0 +1,124 @@
+"""Account balance trees."""
+
+import collections
+
+from beancount.core import account, convert
+
+from fava.core.inventory import CounterInventory
+
+
+# pylint: disable=too-few-public-methods
+class TreeNode(object):
+    """A node in the account tree."""
+    __slots__ = ('name', 'children', 'balance', 'balance_children', 'has_txns')
+
+    def __init__(self, name):
+        #: str: Account name.
+        self.name = name
+        #: A list of :class:`.TreeNode`, its children.
+        self.children = []
+        #: :class:`.CounterInventory`: The cumulative account balance.
+        self.balance_children = CounterInventory()
+        #: :class:`.CounterInventory`: The account balance.
+        self.balance = CounterInventory()
+        #: bool: True if the account has any transactions.
+        self.has_txns = False
+
+
+class Tree(dict):
+    """Account tree.
+
+    Args:
+        entries: A list of entries to compute balances from.
+    """
+    def __init__(self, entries=None):
+        dict.__init__(self)
+        self.get('', insert=True)
+        if entries:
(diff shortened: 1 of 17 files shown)


xentac:
Nice!

This solution still doesn't make Assets - Liabilities = Equity, but I like the performance improvements for sure.

xentac:
I also like the speed improvements :)

fava/core/tree.py
```diff
@@ -0,0 +1,112 @@
...
+        for count in self.values():
+            if count != ZERO:
+                return False
+        return True
+
+    def reduce(self, reducer, *args):
+        """Reduce inventory.
+
+        Note that this returns a simple currency to number dict, not a
+        CounterInventory.
```
xentac: This comment disagrees with line 47. You probably just want a dict and not a CounterInventory.

yagebu: thanks, fixed (the docstring)

xentac:
Nice on the summarize.cap optimization. I had been looking into that myself a few days ago.

This patch set brings the balance sheet rendering time down to 0.4s from 2.0s for me!

I also sent a patch to beancount to convert an Inventory to a dict, so that may speed up other stuff.

adamgibbins:
Really liking the look of this unrealized gain enhancement. The state of unrealize gains in fava still remains a bit mysterious to me, this is a good step in the right direction. It'd be nice to factor this into the networth graph somehow, a different coloured line or something, if sanely doable?

xentac:
Networth already factors in market values. It doesn't make a distinction between unrealized gains though.

adamgibbins:
Right, which I'd like it to - that makes sense if you're selecting the "at value" in the drop down. It should distinguish it to be consistent with this pull request however.

yagebu:
The net worth chart is independent of the dropdown, it is always just the sum of all assets and liabilities converted to all operating currencies separately. I'd prefer to keep it simple - and for simply knowing my "net worth" it is irrelevant how much of the net worth is due to unrealized gains or something else. I think this would be better to show on the account page (the chart there only shows units currently) or as new chart on the balance sheet that follows the conversion specified in the dropdown.

xentac:
Given how awesome this PR currently is, I'd rather merge it as is and look at additional improvements later.

yagebu:
Yes, that's my plan ;) Just need to add an option for the unrealized gains account.

PR #608, opened 2017-11-08
When dropping a document on a transaction, use tx date in filename suggestion

HappyEinara:
When I want to add a document to a transaction (e.g. a trade confirmation for that tx), fava prepends the current date in its suggested filename. It would make more sense to suggest the date of the transaction by default.


aumayr:
Sounds good to me, thanks for the contribution!

Issue #614, opened 2017-12-01
Render commas according to Indian system


balaramadurai:
Hello,
  I love fava and beancount. This is a fantastic and powerful personal finance system.

I live in India and we follow a different numbering system (the placement of the commas in currency) as compared to the rest of the world. Reference - [Indian numbering system](http://en.wikipedia.org/wiki/Indian_numbering_system)

Would it be possible to include an option in render_commas to include the locale like in here - https://stackoverflow.com/a/40952632?

Thanks and have a great day!
Bala

blais:
https://bitbucket.org/blais/beancount/issues/204/add-international-support-for-rendering

PR #617, opened 2017-12-18
Internationalizing rendering of commas

balaramadurai:
Hello all,
  a while back I had raised an issue #614  of a minor feature request for commas based on Indian Numbering system. I've tinkered with the code a bit and it works for me in my local copy. I extended it a little bit to make it work for other international locales as well. Warning: I haven't extensively tested this feature.

Thanks for developing beancount/fava. It's awesome!

Bala


aumayr:
Hi Bala! Thanks a lot for the contribution - always nice to see when people implement their own ideas/wishes! I will add some comments to the code, so we can bring it into shape to merge it.

fava/core/fava_options.py
```diff
@@ -65,6 +66,7 @@
 ]
 
 STR_OPTS = [
```
aumayr: Should be sorted alphabetically.

fava/core/__init__.py
```diff
@@ -187,12 +188,8 @@ def load_file(self):
         self.price_map = prices.build_price_map(self.all_entries)
         self.all_root_account = realization.realize(self.all_entries,
                                                     self.account_types)
-        if self.options['render_commas']:
-            self._format_string = '{:,f}'
-            self._default_format_string = '{:,.2f}'
-        else:
-            self._format_string = '{:f}'
-            self._default_format_string = '{:.2f}'
+        self._format_string = '%.2f'
```
yagebu: if these are just constant, we don't need variables for them

fava/help/options.md
```diff
@@ -89,6 +89,16 @@ click to reload the page.  If the file change is due to user interaction, e.g.,
 uploading a document or adding a transaction, Fava will always reload the page
 automatically.
 
+## `locale`
+
+Default: `en_US.UTF8`.
```
yagebu: I think the default should be `None` (i.e. use the system locale)

fava/help/options.md
```diff
@@ -89,6 +89,16 @@ click to reload the page.  If the file change is due to user interaction, e.g.,
...
+
+If you set `render_commas` to `TRUE`, you can set up your locale to your favourite setup. 
+`2010-01-01 custom "fava-option" "locale" "en_IN"`
+This will change the comma system to the [http://en.wikipedia.org/wiki/Indian_numbering_system](Indian numbering system).
+
+There are other locales described in [http://babel.pocoo.org/en/latest/api/numbers.html](here)
```
yagebu: The site you link to doesn't really describe any locales.

yagebu:
Thanks for the PR!

There are some problems with the current approach: The code now always uses the locale to render the number, so if `render_commas` is False, then the default system locale, with grouping set to True, so commas will actually be rendered in most cases.

Also, it doesn't play with multiple files - setting the locale does so for the whole process. Probably better to use something like `babel.numbers.format_number` instead of the `locale` module.

balaramadurai:
@aumayr and @yagebu, Thanks a lot for your detailed review. Given my very very limited python experience, I have tried my best to address your comments.

The code works for `render_commas` `TRUE` and `FALSE`. The comma stuff works for en_IN and en_US.UTF8 settings. I searched high and low to get the correct codes for other locales, but couldn't. 

Again, thanks a bunch! Have a great day!

aumayr:
>  I searched high and low to get the correct codes for other locales, but couldn't.

Have a look at the `babel` documentation here: http://babel.pocoo.org/en/latest/numbers.html

yagebu:
@balaramadurai thanks for the work on this! In the current state, I won't merge it, but I'll try to take this over and finish it (at some point, hopefully soon).

aumayr:
@yagebu I think this should be part of the next (soonish) release. What do you think?

yagebu:
@aumayr: Don't know if I'll have the motivation soon to implement this. It shouldn't block the release.

PR #619, opened 2017-12-24
Small style changes

aumayr:
Before:
before

After:
after

### DIFF
--- a/fava/static/css/base.css
+++ b/fava/static/css/base.css
@@ -214,6 +214,7 @@ button,
 .button {
   background-color: var(--color-header);
   border: 0;
+  border-radius: 0;
   color: var(--color-background);
   cursor: pointer;
   outline: 0;
--- a/fava/static/css/base.css
+++ b/fava/static/css/base.css
@@ -601,8 +601,16 @@ td .status-indicator {
       color: var(--color-background);
     }
 
-    & span {
+    & span:before {
       background-color: var(--color-autocomplete-match);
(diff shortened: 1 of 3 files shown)


fava/static/css/base.css
```diff
@@ -601,8 +602,16 @@ td .status-indicator {
       color: var(--color-background);
     }
 
-    & span {
+    & span::before {
```
yagebu: Why not just set the background color?

aumayr: This way the background overlaps a little bit to the right and to the left, making it easier on the eyes.

yagebu: Ah, I think that can be done more simply (and slightly better, since it will take the width of letters into account) by padding+negative margin.

aumayr: Can you incorporate this? I'm not quite sure how this works with negative margins…

yagebu: Just add a padding of .1em left and right (or however much you want it to overlap) and a margin of -.1em left and right

aumayr: Ahh, thanks for the hint! Will modify the code.

aumayr: @yagebu This does not work as well, because just setting the padding and margin does show that letters further down the word/sentence are on top of the z-scale, so the background overlaps the letter to the left of the highlighted section a bit, which looks weird. By working with `::before`, the background stays behind the text.

Issue #543, opened 2017-05-07
Group journal entries by day, week, month


aumayr:
Give the option to visually group Journal entries by day, week or month.

PR #620, opened 2017-12-24
Group Journal entries

aumayr:
This PR implements #543: "Group journal entries by day, week, month"

Progress:

- [X] Add template filter
- [ ] Add tests
- [x] Modify JS-code for sorting the table

### DIFF
--- a/fava/template_filters.py
+++ b/fava/template_filters.py
@@ -173,20 +173,18 @@ def format_errormsg(message):
     return new_message.replace('for \'', 'for ').replace('\': ', ': ')
 
 
-def group_by_date(entries, kind='M'):
-    if kind == 'Y':
-        def group_fn(d): return d.year
-    elif kind == 'Q':
-        def group_fn(d): return '{}-Q{}'.format(d.year, (d.month-1)//3+1)
-    elif kind == 'M':
-        def group_fn(d): return '{}-{:02}'.format(d.year, d.month)
-    elif kind == 'W':
-        def group_fn(d): return '{}-W{:02}'.format(d.year, d.isocalendar()[1])
-    elif kind == 'D':
-        def group_fn(d): return '{}-{:02}-{:02}'.format(d.year, d.month, d.day)
-    else:
-        return [(None, entries)]
-    values = set(map(lambda entry: group_fn(entry.date), entries))
-    return [
-        (x, [entry for entry in entries if group_fn(entry.date) == x])
-        for x in values]
+def date_group(entry):
+    """Return the date group an entry belongs to."""
+    kind = g.ledger.fava_options['interval']
+    d = entry.date
+    if kind == 'year':
+        return d.year
+    elif kind == 'quarter':
+        return '{}-Q{}'.format(d.year, (d.month-1)//3+1)
+    elif kind == 'month':
+        return '{}-{:02}'.format(d.year, d.month)
+    elif kind == 'week':
+        return '{}-W{:02}'.format(d.year, d.isocalendar()[1])
+    elif kind == 'day':
+        return '{}-{:02}-{:02}'.format(d.year, d.month, d.day)
+    return None
--- a/fava/template_filters.py
+++ b/fava/template_filters.py
@@ -173,18 +173,17 @@ def format_errormsg(message):
     return new_message.replace('for \'', 'for ').replace('\': ', ': ')
 
 
-def date_group(entry):
+def date_group(entry, interval):
     """Return the date group an entry belongs to."""
-    kind = g.ledger.fava_options['interval']
     date = entry.date
-    if kind == 'year':
+    if interval == 'year':
         return date.year
-    elif kind == 'quarter':
+    elif interval == 'quarter':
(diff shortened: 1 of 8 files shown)


yagebu:
Cool! Some preliminary notes:

- I think this could be enabled by default (and doesn't even need option to disable it IMHO).
- For the grouping interval, I'd just use `interval`.
- The grouping filter is highly inefficient and it adds ~1s for me - the entries are already sorted by group so there's not really any need for it. Just check for each entry if we switched group.

aumayr:
I'm not quite happy with the current state:

- Sorting does not work (and would require big JS effort to implement).
- Hiding empty groups is done in JS, which leads to the rendering of empty groups until the whole journal table is rendered. This looks ugly.

@yagebu Do you have any ideas/suggestions on how to implement this in a better and more efficient way?

fava/template_filters.py
```diff
@@ -173,20 +173,18 @@ def format_errormsg(message):
...
-        (x, [entry for entry in entries if group_fn(entry.date) == x])
-        for x in values]
+def date_group(entry):
+    """Return the date group an entry belongs to."""
+    kind = g.ledger.fava_options['interval']
+    d = entry.date
+    if kind == 'year':
+        return d.year
+    elif kind == 'quarter':
+        return '{}-Q{}'.format(d.year, (d.month-1)//3+1)
```
yagebu: there's no need to render this to a string, just return a tuple

yagebu: or maybe nevermind: for the comparison at least it's not needed, but to render out the headers it is. Might not be worth the time saved to have two functions for this

yagebu:
- Don't hide empty groups. The fact that there aren't any entries shown in a group is a fact worth showing IMHO
- Sorting and grouping don't make sense together except for the initial order. So to sort, simply delete all group headers and then sort as before.

aumayr:
> - Don't hide empty groups. The fact that there aren't any entries shown in a group is a fact worth showing IMHO

This is not really useful, as for every query/filter you set, all `open` statements will be in the Journal, and when hidden by default, empty groups for all `open` statements will be shown.

> Sorting and grouping don't make sense together except for the initial order. So to sort, simply delete all group headers and then sort as before.

Agreed, and implemented in 9e5e2af

aumayr:
@yagebu What do you think of the current implementation (code-/design-wise)?

fava/template_filters.py
```diff
@@ -171,3 +171,20 @@ def format_errormsg(message):
     new_message = re.sub(ACCOUNT_RE,
                          '<a href="{}">'.format(url) + r'\1' + '</a>', message)
     return new_message.replace('for \'', 'for ').replace('\': ', ': ')
+
+
+def date_group(entry):
+    """Return the date group an entry belongs to."""
+    kind = g.ledger.fava_options['interval']
```
yagebu: not the option, but the current value needs to be used

fava/template_filters.py
```diff
@@ -171,3 +171,20 @@ def format_errormsg(message):
...
+    date = entry.date
+    if kind == 'year':
+        return date.year
+    elif kind == 'quarter':
+        return '{}-Q{}'.format(date.year, (date.month-1)//3+1)
+    elif kind == 'month':
+        return '{}-{:02}'.format(date.year, date.month)
```
yagebu: just return the tuple, same for the other cases

fava/templates/_journal_table.html
```diff
@@ -92,100 +92,109 @@
...
+  {% set type = entry.__class__.__name__.lower() %}
+  {% set entry_hash = entry|hash_entry %}
+  {% set metadata = entry.meta|remove_keys(['__tolerances__', '__automatic__', 'filename', 'lineno']) %}
+  <li class="{{ type }} {{ entry.type or '' }} {{ entry.flag|flag_to_type if entry.flag else '' }}
+  {{- ' statement' if entry.tags and 'statement' in entry.tags else '' }}
+  {{- ' discovered' if entry.tags and 'discovered' in entry.tags else '' }}
+  {{- ' hidden' if not entry|show_journal_entry else '' }}
+  {{- ' fail' if entry.diff_amount else '' }}">
+    <p>
+      <span class="datecell" data-group-value="{{ ns.group }}" data-sort-value=-{{ loop.index }}><a href="#context-{{ entry_hash }}">{{ entry.date }}</a></span>
```
yagebu: unused data-group-value - remove

fava/template_filters.py
aumayr: https://github.com/beancount/fava/pull/620#discussion_r159632840 ?

yagebu:
I think overall that the implementation problems this creates make this somewhat not worth it. What do you think @aumayr? Are you planning to continue this (or probably redo since this has gotten somewhat unmergeable in parts)?

aumayr:
Implementing (actually: redoing) this is still worth it I think, because at least once a month I wish for grouping by month. Will push this up on my todo list.

yagebu:
Will close the PR for now, the issue's still open ;)

Issue #603, opened 2017-11-02
filter on metadata values


zacchiro:
As far as I can tall there is no UI element in Fava that allows to query based on metadata values.
Similarly to the "Tag" text entry, one could imagine a "Metadata" text entry that takes a comma-separated list of "key: value" pairs, filtering on the corresponding metadata values.
Bonus point: supporting regular expressions (on both sides?).

Thanks for Fava!

yagebu:
Have you tried writing a `FROM` expression for this? The query language has a `ENTRY_META()` function, so for a fixed key this should be possible.

I'm not really keen on adding any more text entries to the filters form, it already gets quite clunky on smaller screens. I'd rather be up for grouping some of them together - however it seems to me that would duplicate what the query language can already do.

zacchiro:
I confess I'm having hard time understanding which part of the Beancount query language can be written in the FROM input field, the doc says "as supported by BQL" but that doesn't seem to be the case. So I'm guessing you support a subset of the query language (which makes sense), but I can't figure out which subset exactly. Can you clarify?

yagebu:
Any entry filter expression suppported by the BQL can be used (this is referred to as `<entry-filter-expression>` in the [Google Doc](https://docs.google.com/document/d/1s0GOZMcrKKCLlP29MD7kHO4L88evrwWdIO0p4EwRBE0) and `<from_expr>` in the integrated help). You can run `help from` in the query shell to get a list of supported columns and functions.

zacchiro:
On Tue, Nov 07, 2017 at 12:56:32PM +0000, Jakob Schnitzer wrote:
> Any entry filter expression suppported by the BQL can be used (this is
> referred to as `<entry-filter-expression>` in the [Google
> Doc](https://docs.google.com/document/d/1s0GOZMcrKKCLlP29MD7kHO4L88evrwWdIO0p4EwRBE0)
> and `<from_expr>` in the integrated help). You can run `help from` in
> the query shell to get a list of supported columns and functions.

Thanks for this explanation. Do you confirm that you cannot use WHERE
expressions? Because that is my feeling and also the cause of trouble
here.

For instance, a query like:

  select * where any_meta('trip') ~ 'berlin'

does work in native bean-query (and its web based equivalent), but
AFAICT cannot be entered (not even clipped at the start of the
any_meta() call) in the FROM box.

The reason seems to be that the following query does *not* work:

  select * from any_meta('trip') ~ 'berlin'

[...]

PR #622, opened 2017-12-27
Add todo_as_error plugin

aumayr:
## Display 'todo'-metadata-entries as errors

When enabling the `todo_as_error`-plugin, transactions with the `todo`-metadata-key will be added as Beancount errors, displaying the value of the `todo`-metadata-entry as the error description.

```ruby
    plugin "fava.plugins.link_statements"

    2017-12-27 * "" "Groceries"
      todo: "Put the milk into the fridge"
      Expenses:Groceries   150.00 USD
      Assets:Cash
```

### DIFF
--- a/fava/help/features.md
+++ b/fava/help/features.md
@@ -189,3 +189,15 @@ corresponding transaction and can be filtered in the Journal:
 
 When displaying a Journal, including a filtered Journal, the entries displayed
 can be downloaded in Beancount format by clicking the `Export` button.
+
+### Display 'todo'-metadata-entries as errors
+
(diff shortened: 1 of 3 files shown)


yagebu:
Marking them as errors seems to be abusing the Beancount error reporting a bit. When something implementing #603 gets implemented, this would just be a quick filter away.

In any case, I don't think this should live in-tree, like #522.

aumayr:
> In any case, I don't think this should live in-tree, like #522.

What do you think about a sister-repo `fava-plugins`?

yagebu:
> What do you think about a sister-repo fava-plugins?

👍

yagebu:
Closing, I've created the repo

aumayr:
Thanks for implementing/making this!

PR #623, opened 2017-12-27
Entry form: /-key does divide the number by 2

aumayr:
This implements a special requirement I have, and this PR is here to show what the change would implement, and a point of discussion if we merge this into `master` or not. 

Here is the feature: As I split many expenses with my girlfriend, I often have to divide amounts by 2, say an expense on my credit card for example, where half goes to my Expense-account and half to her Liabilities-account. With this change, when the focus is in the `number`-input field, whenever the user presses the `/`-key, the number get's divided by 2. This makes this kind of entries very quick to do.

This change should not really "interrupt" anyone who does not know or need this feature, as it is not very common to enter `/` in the number-input-field. 

@yagebu What do you think of this? Are you OK with merging this into `master`?

### DIFF
--- a/fava/json_api.py
+++ b/fava/json_api.py
@@ -122,6 +122,15 @@ def add_document():
     return _api_success(message='Uploaded to {}'.format(filepath))
 
 
+def _parse_number(num):
+    if not num:
+        return None
+    if '/' in num:
+        left, right = num.split('/')
+        return D(left) / D(right)
+    return D(num)
+
+
 def json_to_entry(json_entry, valid_accounts):
     """Parse JSON to a Beancount entry."""
     # pylint: disable=not-callable
@@ -140,7 +149,7 @@ def json_to_entry(json_entry, valid_accounts):
                 raise FavaAPIException('Unknown account: {}.'
                                        .format(posting['account']))
             data.create_simple_posting(txn, posting['account'],
-                                       posting.get('number') or None,
+                                       _parse_number(posting.get('number')),
                                        posting.get('currency'))
 
         return txn
@@ -148,7 +157,7 @@ def json_to_entry(json_entry, valid_accounts):
         if json_entry['account'] not in valid_accounts:
             raise FavaAPIException(
                 'Unknown account: {}.'.format(json_entry['account']))
-        number = D(json_entry['number'])
+        number = _parse_number(json_entry['number'])
         amount = Amount(number, json_entry.get('currency'))
 
         return data.Balance(json_entry['metadata'], date,
--- a/fava/static/css/base.css
+++ b/fava/static/css/base.css
@@ -78,7 +78,7 @@ h4,
 h5 {
   color: var(--color-headings);
   font-weight: 500;
-  margin: 0 0 1rem;
+  margin: 0 .5rem 1rem 0;
   padding: 0;
 }
 
--- a/fava/static/css/ingest.css
+++ b/fava/static/css/ingest.css
@@ -41,6 +41,7 @@
 
     & label {
       display: inline-block;
+      margin: 0;
     }
   }
 
--- a/fava/templates/_entry_forms.html
+++ b/fava/templates/_entry_forms.html
@@ -1,20 +1,18 @@
 {% macro posting(posting=None) %}
 <div class="fieldset posting">
-  <label for="account">{{ _('Posting') }}:</label>
   <button class="muted round remove-fieldset" data-event="remove-fieldset" type="button" tabindex="-1">×</button>
-  <input type="text" class="account" placeholder="{{ _('Account') }}" list="accounts" value="{{ posting.account or '' }}"/>
-  <input type="number" step="0.01" class="number" placeholder="{{ _('Number') }}" value="{{ posting.units.number if posting else '' }}"/>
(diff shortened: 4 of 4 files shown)


yagebu:
> This change should not really "interrupt" anyone who does not know or need this feature, as it is not very common to enter / in the number-input-field.

It's currently not supported by Fava (as rendering the entry would fail), but something like `100/3` is a valid Beancount number expression. I think it would be neater to just support that and in particular not hard-code division to be division by 2.

aumayr:
> It's currently not supported by Fava (as rendering the entry would fail), but something like `100/3` is a valid Beancount number expression.

This was a main reason for this change. But implementing this would also mean that the `number`-field would not be `type="number"`, so we couldn't take it's advantages in modern browsers.

yagebu:
> This was a main reason for this change. But implementing this would also mean that the number-field would not be type="number", so we couldn't take it's advantages in modern browsers.

I have to admit I never used any of these "advantages" - is there even anything other than using up/down to increase/decrease the number?

aumayr:
> I have to admit I never used any of these "advantages" - is there even anything other than using up/down to increase/decrease the number?

It only accepts numbers, and afaik it also "validates" the number, so no two decimal points, etc.

yagebu:
> It only accepts numbers, and afaik it also "validates" the number, so no two decimal points, etc.

This seems to be very inconsistent between browsers though. For example Firefox allows one to type in any other character (and only marks the input), whereas Chrome ignores characters like `/`. As it doesn't seem to be standardised, this doesn't seem like something to be relied upon (we're not doing any validation for the other fields anyway).

If we want to have it behave like the number input, it should be only a few lines of JS (and that'd work the same across all browsers).

aumayr:
> If we want to have it behave like the number input, it should be only a few lines of JS (and that'd work the same across all browsers).

This, and support for "beancount-math" like `100/3` would be awesome, and make this not needed. I'm not fit enough with RegEx to make the math-things to render (and align) correctly in Fava, though.

xentac:
I think the type="number" also makes mobile chrome give the number pad
input. It would be nice to keep that hint.

aumayr:
> I think the type="number" also makes mobile chrome give the number pad
input. It would be nice to keep that hint.

>> If we want to have it behave like the number input, it should be only a few lines of JS (and that'd work the same across all browsers).
> This, and support for "beancount-math" like 100/3 would be awesome, and make this not needed. I'm not fit enough with RegEx to make the math-things to render (and align) correctly in Fava, though.

These two things don't go together: Either we keep `type="number"`, or we implement "beancount-math". 

I am still in favor for keeping this simple implementation and the `type="number"`, but also implementing the Syntax-highlight-and-alignment-support for "beancount-math".

aumayr:
Not both:

1. Keep the current `input`-type and the functionality of this PR, so no "beancount-math" in the Add-Transaction-form as it is `type="number"`, as I think the value of the `"number"` overweights the value of being able to enter "`100/2`" into the field. (_tldr:_ Merge this PR quasi-as-is.)

2. Fix the Syntax-highlighting (and aligning-functionality) in the Editor, so if someone uses this syntax in his/her ledger-files, it is correctly displayed and aligned. The current state where some parts of the beancount-syntax ("beancount-math") is not properly supported by the Fava-editor is not ideal imho.

fava/static/javascript/transaction-overlay.js
```diff
@@ -42,6 +42,12 @@ export default function showTransactionOverlay() {
   if (!initialized) {
     addPostingRow(form);
     addPostingRow(form);
+    $.delegate(form, 'keyup', '.number', (event) => {
```
yagebu: This should go in `entry-forms.js` (otherwise the ingest and overlay form behave differently)

yagebu:
Could this be changed so that `/2` divides by two (and `/x` divides by x for all digits between 2 and 9)?

fava/static/javascript/transaction-overlay.js
aumayr: I moved this to the "global" level in https://github.com/beancount/fava/pull/623/commits/eeb60a6ef9d4d5c3cfefb865e3bbbf894d4b0574, because the Ingest entry forms are not initialized with JavaScript, but the overlay one is.

aumayr:
> Could this be changed so that `/2` divides by two (and `/x` divides by x for all digits between 2 and 9)?

Implemented in 1981d53. Does not look very clean to me, but works.

yagebu:
I tried this for the first time and it doesn't work at all in Firefox, since it behaves differently when one enters an invalid key in the number input: It will display the slash, but the input.value seems to be empty if the entered string is invalid. I think it's better to use an input that doesn't to any validation (and then we can do the division server-side).

yagebu:
@aumayr: The way I've implemented it now, the division is on the server side before the entry is rendered to a file, so entering `5/2` will result in a `2.5` in the file. A future improvement would be of course to support more than just division and to render out the numerical expression. Using the `type=tel` input should bring up the number keyboard on mobile, but without the drawback of having the input validated (which varies from browser to browser).

aumayr:
> The way I've implemented it now, the division is on the server side before the entry is rendered to a file

Works as expected, thanks for the improvement! Do you want to merge this now?

PR #628, opened 2018-01-01
doc: add deployment chapter and document apache/reverse proxy case

zacchiro:


aumayr:
Thanks @zacchiro for the contribution!

yagebu:
I'll revert this PR for the time being (but move the file to `contrib/`):

- As long as the CLI uses the Werkzeug dev server, it shouldn't be really used on some public servers.
- It doesn't talk about encryption and authentication.

I don't want to have this suboptimal solution as the one and only example on the website. PR(s) addressing these points (and generally extending the deployments doc a bit) would be welcome :)

zacchiro:
On Fri, Feb 23, 2018 at 05:11:56PM -0800, Jakob Schnitzer wrote:
> - As long as the CLI uses the Werkzeug dev server, it shouldn't be really used on some public servers.
> - It doesn't talk about encryption and authentication.

I can easily address the second point here, which is indeed what I use
in production, if you think that'd be enough to keep the doc around.

Cheers
-- 
Stefano Zacchiroli . zack@upsilon.cc . upsilon.cc/zack . . o . . . o . o
Computer Science Professor . CTO Software Heritage . . . . . o . . . o o
Former Debian Project Leader & OSI Board Director  . . . o o o . . . o .
« the first rule of tautology club is the first rule of tautology club »

PR #630, opened 2018-01-03
Move export button to sidebar

yagebu:
There's no reason to only have this in the journal (where it actually often leads to a line break in the form for me)

### DIFF
--- a/fava/application.py
+++ b/fava/application.py
@@ -295,7 +295,8 @@ def download_query(result_format):
 @app.route('/<bfile>/download-journal/')
 def download_journal():
     """Download a Journal file."""
-    filename = "journal_{}.beancount".format(datetime.datetime.now())
+    now = datetime.datetime.now().replace(microsecond=0)
+    filename = "journal_{}.beancount".format(now.isoformat())
     data = BytesIO(bytes(render_template('beancount_file'), 'utf8'))
     return send_file(data, as_attachment=True, attachment_filename=filename)
 
--- a/fava/static/css/base.css
+++ b/fava/static/css/base.css
@@ -309,27 +309,6 @@ header {
   }
 }
 
-.add-transaction-button {
-  background-color: var(--color-sidebar-background);
-  color: inherit;
-  font-size: 25px;
-  height: 100%;
-  line-height: 1;
-  margin: 0;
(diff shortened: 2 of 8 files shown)


aumayr:
Having this button in the sidebar is kind of confusing for me: All sidebar buttons lead to separate pages/views on my data, but this button is a function that is called, and does something when I press it. 

Wouldn't it be better to find a place up in the header, where the button can be shown whenever the current view can be exported? There we also would have room to expand the button when there are multiple formats for export available.

yagebu:
> All sidebar buttons lead to separate pages/views on my data, but this button is a function that is called, and does something when I press it.

The add-transaction button is similar, no? I find it much more logical to have this next to the import link than anywhere else.

The header would also be fine, but I find it a bit too crowded already (in particular if one doesn't have a wide screen).

aumayr:
> The add-transaction button is similar, no?

True!

What about creating a separate "Export" screen where this button in the sidebar leads to? (Again, thinking about different views to be exported, and different formats, eg. "Export the Balance Sheet as a PDF with a graph" or "Export the Journal as a CSV-file", etc.)

yagebu:
> What about creating a separate "Export" screen where this button in the sidebar leads to?

Sounds good, I'd just make it an overlay though (as it would be related to the current report if it worked as you propose). Right now there's just one export available which is the same for all reports, so I'd leave it as a direct link for now.

aumayr:
> Sounds good, I'd just make it an overlay though

👍 

> Right now there's just one export available which is the same for all reports

I would implement CSV, Excel, (Google Spreadsheets?) and PDF if you implement the overlay right away… was just waiting for an opportunity to work on this.

yagebu:
Cool. Will do.

gety9:
@yagebu @aumayr 
Do i need to reinstall fava to see the export button/menu?
Don't see it one my end - trying to understand what i am doing wrong.

adamgibbins:
@gety9 it is presented as an arrow next to the Import link at the bottom of the sidebar, do you not see it?

gety9:
@adamgibbins thank you!

who put it there :) it's impossible to find it if you don't know where to look.

Anyway when i press it it says "Download currently filtered entries as a Beancount file", don't see the google docs options

PR #639, opened 2018-01-15
Fix Editor focus

### LINKED ISSUE #606: Navigation in editor by arrow keys is interrupted by autocomplete popup (focus is lost)
If I move cursor in editor window with keys Up, Down, Home, End, PgUp, PgDown, and cursor happens to stop on word with autocomplete popup, then popup appears and navigation with keys can't be continued: focus is stolen by popup.

Is there any workaround to show popup only by hotkey?

aumayr:
This PR regards the autocomplete popup and how it reacts to different keystrokes. 

- Don't open the popup when navigating the editor with arrow keys (see #606)
- When the popup is shown and the user presses `Esc`, close the popup.

Fixes #606

### DIFF
--- a/fava/static/javascript/editor.js
+++ b/fava/static/javascript/editor.js
@@ -146,8 +146,8 @@ function initQueryEditor() {
   const editor = CodeMirror.fromTextArea(queryForm.elements.query_string, queryOptions);
 
   editor.on('keyup', (cm, event) => {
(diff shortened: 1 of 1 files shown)


yagebu:
This PR is missing some keys that #641 covers, but both use `.keyCode` which is deprecated, I'll implement this using `.key` instead.

Issue #605, opened 2017-11-05
Confirmaton to save changes in editor and autosave editor option


Vanav:
Currently if I make some changes to editor and then switch to other tab or close browser, unsaved changes are lost without confirmation. This often leads to data loss and is not an expected behavior.

I suggest:
1) on switch to other tab or closing browser ask for confirmation: "Save? Yes/No".
2) add an option to autosave any changes on tab switch or browser close.

Other requests for this fix: #191.

yagebu:
Closed in bb5ade85c40cf2f7fcdc29b5f92d01aa9d8fdda1

PR #640, opened 2018-01-15
Confirmaton to save changes in editor

aumayr:
This tries to implement the first part of #605: Adding a confirmation dialog when navigation away from the Editor when there are unsaved changes.

Ref #605

### DIFF
--- a/fava/static/javascript/editor.js
+++ b/fava/static/javascript/editor.js
@@ -222,6 +222,7 @@ const sourceEditorOptions = {
   },
 };
 
+let activeEditor = null;
 // Init source editor.
 export default function initSourceEditor(name) {
   sourceEditorOptions.rulers = [];
@@ -236,6 +237,9 @@ export default function initSourceEditor(name) {
   if (!sourceEditorTextarea) { return; }
 
   const editor = CodeMirror.fromTextArea(sourceEditorTextarea, sourceEditorOptions);
+  if (name === '#source-editor') {
+    activeEditor = editor;
+  }
   const saveButton = $(`${name}-submit`);
   editor.setOption('favaSaveButton', saveButton);
 
@@ -284,4 +288,27 @@ e.on('page-loaded', () => {
   initQueryEditor();
   initReadOnlyEditors();
   initSourceEditor('#source-editor');
+});
+
+const leaveMessage = 'There are unsaved changes. Are you sure you want to leave?';
+
+e.on('navigate', (state) => {
+  if (activeEditor) {
+    if (!activeEditor.isClean()) {
+      const leave = window.confirm(leaveMessage); // eslint-disable-line no-alert
+      if (!leave) {
+        state.interrupt = true; // eslint-disable-line no-param-reassign
+      } else {
+        activeEditor = null;
+      }
+    } else {
+      activeEditor = null;
+    }
+  }
+});
+
+window.addEventListener('beforeunload', (event) => {
+  if (activeEditor && !activeEditor.isClean()) {
+    event.returnValue = leaveMessage; // eslint-disable-line no-param-reassign
+  }
 });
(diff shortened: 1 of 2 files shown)


aumayr:
The current implementation lacks one crucial part: When navigation happens within the page (with the async routing mechanism), this does not get reset in any way. 

@yagebu Is there a way to get an event/notification in JS when the user navigates away from the Editor page on an internal link, so this dialog can be shown/reset accordingly?

yagebu:
> Is there a way to get an event/notification in JS when the user navigates away from the Editor page on an internal link

You can add one to Router.navigate in router.js

aumayr:
OK this is architecturally not simple… in order to keep `Router.navigate` generic there needs to be a mechanism to register callbacks for `beforeNavigate`, so the callback (in this case registered by `editor.js` can cancel the `navigate` if dirty.

fava/static/javascript/editor.js
```diff
@@ -254,6 +255,26 @@ export default function initSourceEditor(name) {
       });
     }
   });
+
+  const leaveMessage = 'There are unsaved changes. Are you sure you want to leave?';
+
+  router.on('navigate', () => {
+    if (!editor.getDoc().isClean()) {
+      const leave = confirm(leaveMessage); // eslint-disable-line no-restricted-globals,no-alert
```
yagebu: If you use `window.confirm`, one of the warnings should go away

fava/static/javascript/editor.js
```diff
@@ -254,6 +255,26 @@ export default function initSourceEditor(name) {
...
+      if (leave) {
+        editor.getDoc().markClean();
```
yagebu: This seems wrong

fava/static/javascript/editor.js
```diff
@@ -254,6 +255,26 @@ export default function initSourceEditor(name) {
...
+      }
+      return !leave;
+    }
+    return false;
+  });
+
+  window.onbeforeunload = () => {
```
yagebu: This and the other event handler persist when one navigates away from the editor. Better use window.addEventListener and add the event handlers once globally (on page-init)

fava/static/javascript/editor.js
aumayr: When navigating away from the (dirty) editor, and then navigating back, the change made before will be gone, but the editor is still marked dirty.

aumayr:
> Also, I'd prefer to just keep the single event handler `e` from events.js.

In 97d4e9f I had to introduce a return value for callbacks to events, because the router (calling `e.trigger(...)` needs to know (from the editor) if the navigation should be allowed.

fava/static/javascript/editor.js
aumayr: The check if the editor is dirty needs a reference to the `editor`-variable, pointing to the CodeMirror instance running the editor. I don't know how to get this reference when putting this code in a global scope without using a global variable (or is that what you suggest?).

yagebu: This doesn't need a global variable, one within the scope of this file is enough. Right now the above attaches tons of event handlers even though at most one is needed

yagebu:
I've moved the place where we trigger the event to `router.loadURL` so that it also interrupts reloads etc. @aumayr: I think this should be ready, but could you test it to check if there's any edge cases this code might be missing?

aumayr:
Three issues:

- JS-Exception `Uncaught TypeError: Cannot create property 'interrupt' on string 'http://fava:5000/aumayr/statistics/'`
- When navigating away from a "dirty" Editor, the popup shows twice.
- The popup also applies/shows when editing a Journal Entry (clicking on the hash, ...)

yagebu:
I can reproduce none of these issues (in neither Chrome nor Firefox). Can you try to debug them?

aumayr:
Arrr, fault on my part. A `git reset --hard origin/editor-dirty-alert-605` and `make` fixed all the issues 👍

PR #641, opened 2018-01-15
Don't show hint on keyboard navigation (#606)

Vanav:


yagebu:
`.keyCode` and `.which` are deprecated, I'll implement this using `.key` instead. (and I'll also trim down the list a bit)

yagebu:
Thanks for the PR though :)

Issue #258, opened 2016-04-09
New chart types: Calendar and Sankey


aumayr:
These two charts look like we could use them:

![sankey](https://cloud.githubusercontent.com/assets/5135450/14403424/ee9b3488-fe55-11e5-8674-2a41127860dc.png)

https://bost.ocks.org/mike/sankey/

![calendar](https://cloud.githubusercontent.com/assets/5135450/14403425/f0800846-fe55-11e5-9460-3876af0ef850.png)
http://bl.ocks.org/mbostock/4063318

Do you think this is worth the trouble?

corani:
Here are almost 2500 sample D3 charts that may give some inspiration: http://christopheviau.com/d3list/gallery.html

aumayr:
Do any particular ones spring to your eye, @corani ?

corani:
No particular ones (yet). I came across the page last night in a D3 tutorial video.

vlaci:
I think a [stacked bar chart](http://bl.ocks.org/timelyportfolio/5586845) would make sense e.g. for the `Monthly Changes` graphs. That would make much easier to see the dynamics how accounts' contributions changed from month to month to the expenses

johannesjh:
Also check the visualizations from this open source portfolio analysis tool: 
http://www.portfolio-performance.info/portfolio/

yagebu:
I'll close this issue as there seems to be little interest in implementing calendar and sankey charts (well for sankey, there was some interest but implementing it seems hard/not worth it #642). Links to collections of a hundred charts are not specific enough to start any work in Fava . More specific chart suggestions are very welcome though (but can have and deserve their own issues).

I've openened a new issue for a stacked bar chart.

@johannesjh: The tool you linked to seems to have some nice charts - if there's one you'd really like to see in Fava, could you open a new issue with some more specific suggestions?

yagebu:
I'm still not sure whether it works well (with many accounts, three-and-more-legged transactions) but maybe vibecode it to see whether it's worth it? ;)

lukasalexanderweber:
+1 on "would love sankey in fava" 😊

to get a better understanding where my money flows is the main motivation I started using beancount in 2026

cbrxyz:
The [fava-dashboards](https://github.com/andreasgerstmayr/fava-dashboards) extension can show these graphs! You can use the [example `dashboards.tsx` file](https://github.com/andreasgerstmayr/fava-dashboards/blob/main/example/dashboards.tsx) to get them.

PR #642, opened 2018-01-16
Sankey diagram

aumayr:
This PR adds a Sankey-style diagram to the balance sheet page:

screenshot

At the moment this is just a proof-of-concept, missing layout code and real data. 

Ref #258

### DIFF
--- a/fava/core/charts.py
+++ b/fava/core/charts.py
@@ -173,3 +173,52 @@ class ChartModule(FavaModule):
                     for currency in self.ledger.options['operating_currency']
                 }
             }
+
+    def sankey(self):
+        """Sample datastructure for sankey diagram."""
+        return {
+            'nodes': [
+                { 'name': 'Expenses' },
+                { 'name': 'Assets' },
+                { 'name': 'Liabilities' },
+                { 'name': 'Income' },
+                { 'name': 'Equity' },
+            ],
+            'links': [
+                {
+                    'source': 0,
+                    'target': 1,
+                    'value': 5.1
+                },
+                {
+                    'source': 0,
+                    'target': 3,
+                    'value': 6.3
+                },
+                {
+                    'source': 1,
+                    'target': 2,
+                    'value': 2.0
+                },
+                {
+                    'source': 2,
+                    'target': 3,
+                    'value': 1.5
+                },
+                {
+                    'source': 4,
+                    'target': 3,
+                    'value': 7.8
+                },
+                {
+                    'source': 4,
+                    'target': 1,
+                    'value': 8.1
+                },
+                {
+                    'source': 4,
+                    'target': 2,
+                    'value': 2.9
+                },
+            ]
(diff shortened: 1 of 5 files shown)


aumayr:
This should display the net flow of value between different accounts.

@yagebu Are you interested in helping me create the correct data structures for this (`charts.py`)?

yagebu:
I don't yet see if this chart is actually useful for the kind of data we have. The example by Mike Bostock is good of course, but all the other (similarly looking) sankey charts that a quick image search turned up seemed to be pretty bad visualisations of the underlying data. In our case, this could just turn into a complete mess, with the number of accounts a usual Beancount journal has. And even if the underlying data is discernible, does it actually display anything useful? What kind of insights are you hoping to gain from this chart?

A more immediate problem is actually that it's not really possible to get the necessary data for this chart: A transaction with more than 3 legs can not be correctly translated to a flow between pairs of accounts.

aumayr:
> A more immediate problem is actually that it's not really possible to get the necessary data for this chart: A transaction with more than 3 legs can not be correctly translated to a flow between pairs of accounts.

I agree, and think that we can discard this PR entirely. I wanted to try out if this would provide something useful, and started with the technical question of implementing the diagram graphically, before thinking about the data behind it. 

Anyhow, if anyone in the future has a good idea about how to use this in a useful way, this is a jumping-off point.

tbm:
Here's a good example of how this would be a useful addition for fava:
https://i.imgur.com/90Dx8zJ.png

Of course, the point about too many account names is correct. Users would somehow need to map their accounts to more higher-level accounts they want to see in the chart (on each level of the chart).

yagebu:
@tbm: I missed your comment here, that's actually a good-looking chart. The 3-or-more postings is still a problem though - with that in mind, this chart doesn't seem to be worth the work (for me at least)

tbm:
@yagebu I'll bring this topic up again in a few months when I understand beancount and fava better. Maybe I can figure out a way to do it.

hoostus:
> A more immediate problem is actually that it's not really possible to get the necessary data for this chart: A transaction with more than 3 legs can not be correctly translated to a flow between pairs of accounts.

I think this misunderstands what the sankey chart is graphing here. It isn't going to graph individual transactions, so the number of legs doesn't matter. At least in a personal finance context, it is really just a different visualization of the tree structure on the income statement.

The chart that @tbm linked is a good example. The left side is an (inverse) tree. "Total income" is the root, with "earned income" and "passive income" as nodes. Earned income is divided into "work", "side hustle", etc. The same tree structure happens on the expense side.

That said, I agree that this kind of visualization -- though it looks cool -- doesn't actually add much. Since it is replicating the tree structure you get the same information -- but much more easily understood -- by just having pie charts that you drill down into. You can see the problems with sankey by looking at the far right side: too much detail and too compare things.

PR #648, opened 2018-01-28
Fix dockerfile cleanup

adamgibbins:
The previous implementation used a bashism, which isn't available in the
sh that Docker's RUN uses so no cleanup was actually happening - npm caches etc were left behind. Fixing removes ~80MB from the image size.


aumayr:
Thanks! Will test and merge this in the coming days.

MarkusTeufelberger:
apk supports a `--no-cache` flag by the way when installing packages.

adamgibbins:
@MarkusTeufelberger I tried that, but for some reason it leaves the APKINDEX files within the cache dir. Haven't done any digging to why.

adamgibbins:
^ I did a stupid and accidentally cut the trailing `/` off when I copy and pasted from my non-forked version. Whoops. Fixed in the above (force pushed).

PR #666, opened 2018-02-21
add fava-options for currency alignment

yagebu:


yagebu:
@blais: A new release of Beancount would be nice so that we can use the additional keyword arguments to align_beancount()

yagebu:
This will be superseded by https://bitbucket.org/blais/beancount/issues/224/bean-format-allow-specifying-just-the and the corresponding option, so closing this PR.

PR #673, opened 2018-03-07
Move to more standard css.

yagebu:
The css color() function doesn't seem to be on a standard track - remove it, by specifying colors using hsl() it's often just as easy to achieve the same. Also custom properties are supported by all modern browsers now, so we don't need to remove them via postcss

### DIFF
--- a/fava/static/css/base.css
+++ b/fava/static/css/base.css
@@ -182,7 +182,7 @@ textarea {
 
 input,
 textarea {
-  border: 1px solid color(var(--color-background-darkest));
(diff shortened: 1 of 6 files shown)


aumayr:
screenshot

I think the "Error-bubble" did not "survive" the changes :-/

yagebu:
> I think the "Error-bubble" did not "survive" the changes :-/

It should have (I changed it to be a 50% transparent white), maybe the hsl() alias for hsla() isn't so well supported yet...

PR #674, opened 2018-03-07
Journal colours

yagebu:
I'm not super happy with the current look of the entry journal and think it can be improved. Both colour and layout can be helpful to distinguish entry types and see the necessary information quickly. I think at the moment colours are overused and otherwise the layout is a bit too unstructured.

- I think only showing the "entry colour" in the date and flag column is enough - this allows us to get rid of some colours like the "document tag colour" and use the same colour for tags for all entries.
- By putting the entry type (or a shortened version thereof) in the flag column, sorting becomes more useful and we have more space for the interesting information in the main column.
- Setting a min-width for account links nicely aligns document names (which now use a monospaced font).

Any other suggestions or is someone unhappy about these changes?


aumayr:
I really like the effort, thanks for working on this! I'll post some comments when I have tested the new Journal design.

yagebu:
I'd consider this to be ready (there's always some more tweaking that can be done - but that can also happen later). Now looks like this:

screencapture-localhost-5000-short-example-journal-1520942555180

I've added a somewhat smaller example file (that should be extended even further when making further UI changes) and a `make run-example` command to test this sort of change. This highlighted quite a few bad bugs with the current design.

aumayr:
I really like the new look! 👍

Only thing I would modify is slightly increasing the `padding-left` of the Narration/Payee-column, to have more space between the colored cells and the text in the Narration/payee-column.

yagebu:
> Only thing I would modify is slightly increasing the padding-left of the Narration/Payee-column, to have more space between the colored cells and the text in the Narration/payee-column.

Done, and merged

PR #681, opened 2018-03-14
Use better default order for holdings

tbm:
The holdings page only sorts by account by default.  The default
ordering can be improved by additionally sorting by the units
(currency) and the acquisition date (cost_data).


yagebu:
Thanks!

PR #687, opened 2018-03-17
Fix BQL error on holdings by account page

tbm:
Commit dcf718e ("Use better default order for holdings") introduced
a BQL error on the "Holdings by Account" page.


tbm:
I'm sorry about this bug. I made one change and tested it and just before submitting it I noticed I could also add it to the second query but I think I forgot to test it.

PR #691, opened 2018-03-23
Statement-download: Keep original filename (#688)

### LINKED ISSUE #688: Document download should preserve original filename
I have attached a document to an account.  The file is called 03.pdf.  In fava, the URL also shows 03.pdf and fava itself shows the file as 03.pdf.

But when I download it, I get filenames like tYJlSTo7.pdf and BViw2yLX.pdf.

Is there a way for the original filename to be preserved?

aumayr:
This PR fixes #688 by suggesting the original filename to the browser when sending the statement. 

It still displays the original file inline in the browser (instead
of forcing a direct download), as IMHO this feature is often used for
previewing invoices (PDFs) and most modern browsers can display PDFs
inline.

### DIFF
--- a/fava/application.py
+++ b/fava/application.py
@@ -33,7 +33,7 @@ from fava.core.charts import FavaJSONEncoder
 from fava.core.helpers import FavaAPIException, FilterException
 from fava.help import HELP_PAGES
 from fava.json_api import json_api
-from fava.util import slugify, resource_path, setup_logging, send_as_inline
+from fava.util import slugify, resource_path, setup_logging, send_file_inline
 from fava.util.excel import HAVE_EXCEL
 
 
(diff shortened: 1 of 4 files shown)


aumayr:
@yagebu Should we do this here as well? https://github.com/beancount/fava/pull/691/files#diff-6d90976a6c500672baef2abdbc814f8fR244

aumayr:
Whups, seems like something in my git went wrong… anyhow, 4c84fec is the relevant changeset.

yagebu:
This needs to deal with unicode filenames too (maybe something like https://github.com/pallets/flask/blob/4309c7c6ea51e5797604c43ba858e5ebe1fb1674/flask/helpers.py#L551)

> Should we do this here as well?

Yes

> Whups, seems like something in my git went wrong…

git rebase should easily fix this

yagebu:
Should probably be in some helper function `send_file_inline` in util

yagebu:
The code crashed for non-ASCII latin-1 filenames. I think it's better and much simpler to just always use UTF-8

PR #692, opened 2018-03-24
Remove GUI

yagebu:
The GUI is currently broken, breaks often and is barely used - let's remove that crap.


aumayr:
I do not have the time currently to maintain the GUI, but would you be OK to at least keeping the code in a branch?

yagebu:
> I do not have the time currently to maintain the GUI, but would you be OK to at least keeping the code in a branch?

I'll probably just move it to contrib - in any case, it would always remain in git's history so no need for a branch.

There's more issues with the GUI than it being just broken - I don't think it will be worth bringing it back. (if we want a GUI again, I think it should be just a simple supervisor app to start/stop Fava)

aumayr:
> if we want a GUI again, I think it should be just a simple supervisor app to start/stop Fava

👍

yagebu:
> I'll probably just move it to contrib

I ended up deleting it after all - but moved the stuff we might need a again (icon and pyinstaller spec) to contrib

PR #705, opened 2018-03-26
Small Journal CSS fixes

aumayr:
- Don't allow date or numbers to wrap
- Fix black separator square width

### DIFF
--- a/fava/static/css/journal-table.css
+++ b/fava/static/css/journal-table.css
@@ -27,6 +27,7 @@
   color: var(--color-text);
   font-family: var(--font-family-monospaced);
   text-align: right;
+  white-space: nowrap;
 }
 
 .tree-table .head p > span,
@@ -136,6 +137,7 @@
 
 .journal .datecell {
   width: 5.5rem;
+  white-space: nowrap;
 }
 
 .journal .flag {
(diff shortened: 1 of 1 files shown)


aumayr:
Another point for discussion: In the Journal table, the payee and narration are vertically centered, as well as tags and the points for metadata entries. The date, flag and numbers are not vertically centered, which looks weird. 

Would be vertically centering everything in the Journal table be a good solution?

fava/static/css/journal-table.css
```diff
@@ -136,6 +137,7 @@
 
 .journal .datecell {
   width: 5.5rem;
+  white-space: nowrap;
```
yagebu: Shouldn't the date cell always have enough space?

aumayr: In Chrome and Safari sometimes it doesn't.

yagebu: Does it look bad with the nowrap setting then? We could also make this cell minimally wider

yagebu:
Noticed the vertical centering too

> Would be vertically centering everything in the Journal table be a good solution?

I guess

fava/static/css/journal-table.css
yagebu: :+1:

yagebu:
Since the vertical centering might not be as trivial as these changes here, I've merged this already.

aumayr:
> Since the vertical centering might not be as trivial as these changes here, I've merged this already.

Are you having a go at this?

yagebu:
> Are you having a go at this?

No, I thought you are?

PR #707, opened 2018-03-27
add a contributing document

yagebu:


yagebu:
> Any reason for removing the info about the .whl file? I found that useful.

It has little to do with setting up a development environment.

tbm:
Ok, good point.

Although it might be relevant if your dev platform is a different system than your test platform.  You could build on A, transfer to B, test on B.

yagebu:
> You could build on A, transfer to B, test on B.

You could, but let's not encourage novice users to make their lives harder than necessary.

PR #711, opened 2018-03-27
filters: correctly print illegal character in filter

tbm:


tbm:
Sorry, I'm not sure how to format this since it's >80 characters even with the wrapping I did.

yagebu:
Can you add a test too?

To format, you can use `yapf` (which you might need to install)

tbm:
Thanks for letting me know about yapf!

yagebu:
Thanks!

PR #713, opened 2018-03-27
tox: allow all Python 3 versions

tbm:
The test suite works fine with Python 3.5.  Since Debian 9 (stretch)
doesn't have Python 3.6, allow other Python 3 versions.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,5 @@
 [tox]
-envlist = lint, py36
+envlist = lint, py3
 
 [testenv]
 deps =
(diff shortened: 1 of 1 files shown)


tbm:
Unfortunately I couldn't find a way to say Python 3.5 or higher.  But I don't think many people still use any Python 3 older than 3.5.

Feel free to reject if you don't like the patch but it makes things easier for people who have Python 3.5 installed.

tox.ini
```diff
@@ -8,6 +8,11 @@ deps =
     pyexcel-ods3
 commands = pytest tests
 
+[testenv:py3-dev]
```
yagebu: One doesn't need to run this locally, this can be removed

yagebu: Will do so in the merge

Issue #726, opened 2018-04-01
Prices on balance sheet for market value still use current date not filter date


xentac:
It looks like https://github.com/beancount/fava/issues/613 didn't quite address the prices at filter date problem. I wrote a PR that may address the issue, but I'm not sure it's the right way to do it, especially because it uses the date of the most recent transaction in the Tree. This isn't 100% analogous to the latest filter date, but I didn't want to depend on the g.ledger object everywhere.

yagebu:
Fixed in ef40734ef41f37d8b4f102cd1ba59aa71a4a18e5

PR #727, opened 2018-04-01
Propagate end date to balance tree

xentac:
Tree and TreeNodes are used to render the balance sheet and they have no access to the global end date filtering. We need a way to specify the date to look up prices for calculating market value. Use the transaction date of the last entry in the TreeNode to calculate the price.

Fixes https://github.com/beancount/fava/issues/726.

### DIFF
--- a/fava/core/tree.py
+++ b/fava/core/tree.py
@@ -9,11 +9,13 @@ from fava.core.inventory import CounterInventory
 
 class TreeNode(object):
     """A node in the account tree."""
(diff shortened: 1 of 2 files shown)


yagebu:
I don't like this solution very much - why not just use the end date of the time filter at rendering time?

xentac:
You're right. I didn't realize I could just access the ledger object within a template. I'll send a different PR soon-ish.

Issue #721, opened 2018-03-31
tree-table numbers no longer wrap when viewing "at Market Value"


xentac:
When viewing at Market Value on the balance sheet, black text is pushed into the next column.

In the fix for https://github.com/beancount/fava/issues/708, https://github.com/beancount/fava/commit/95cd532e2a3f619c952543e733751a75b7847a86 probably shouldn't have been applied to ".tree-table .num" and maybe even ".journal .num".

aumayr:
@xentac Can you post a screenshot?

xentac:
![screenshot 2018-04-01 at 10 20 08 am](https://user-images.githubusercontent.com/1805/38175755-593774fe-3596-11e8-9644-ecf414d0d885.png)

Interestingly, another bug is that incognito mode doesn't work with at Market Value... I'll file a bug for that one too.

PR #728, opened 2018-04-02
Fix scope of no-wrap for numbers

aumayr:
@xentac Can you test if the changes in this PR make your tables look better?

Ref #721


xentac:
Yes, it does. Thanks!

PR #735, opened 2018-04-04
Add time filters to account trees

xentac:
To make sure we calculate market prices correctly when filtering.

### DIFF
--- a/fava/templates/_tree_table.html
+++ b/fava/templates/_tree_table.html
@@ -31,8 +31,8 @@ Hold Ctrl or Cmd while clicking to expand one level.') %}
     </p>
   </li>
   {% for account in ([account_node] if account_node.name else account_node.children) if account|should_show recursive %}
-  {% set balance = account.balance|cost_or_value(ledger._filters['time'].end_date) %}
-  {% set balance_children = account.balance_children|cost_or_value(ledger._filters['time'].end_date) %}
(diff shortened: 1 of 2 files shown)


xentac:
A much simpler PR to do the same thing for market value price calculations.

https://github.com/beancount/fava/issues/726

fava/templates/_tree_table.html
```diff
@@ -31,8 +31,8 @@
     </p>
   </li>
   {% for account in ([account_node] if account_node.name else account_node.children) if account|should_show recursive %}
-  {% set balance = account.balance|cost_or_value %}
-  {% set balance_children = account.balance_children|cost_or_value %}
+  {% set balance = account.balance|cost_or_value(ledger._filters['time'].end_date) %}
```
yagebu: `_filters` is a "private" attribute of the ledger class, so it shouldn't be used

xentac: Ok cool. How would you like me to expose the end_date in the FavaLedger class? Should I just create a property called end_date or do you want it to be more complicated than that? I could see a use for making all filters accessible, but then we're essentially renaming _filters to filters.

I suppose we could create a read-only property that exposed a copy of the filters to prevent accidental modification.

xentac:
Change made.

yagebu:
Thanks! Merged

PR #736, opened 2018-04-04
Initial attempt at documenting my fava+oauth2_proxy+letsencrypt setup

xentac:
I'm sure there are minor mistakes, but hopefully there are no major ones.

### DIFF
--- a/contrib/docker/README.md
+++ b/contrib/docker/README.md
@@ -44,8 +44,17 @@ Going to http://localhost:5000/ will display your fava instance.
 
 ## Advanced
 
-### Commit hashes
+Hosting a local docker instance is nice and all, but what we really want is a
+globally available, authenticated, secure deployment of fava.  To do that we
+need two other parts: oauth proxy and letsencrypt.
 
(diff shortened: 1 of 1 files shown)


aumayr:
I think I can contribute a bit as well by

- a `docker-compose.yml`-file to make deployment easier and
- using `https://github.com/SteveLTN/https-portal/` to cut down the number of containers needed for Let's Encrypt

@xentac What do you think about those "changes"?

xentac:
Both of these things I chose not to do to make the system more flexible and
secure. I don't use docker-compose so that I can start my containers as
systemd services and run other web services on the same machine. I chose
the 3 container solution for letsencrypt to keep the docker socket out of a
container that is publicly accessible. The nginx container exists only to
serve web traffic, not to do any management.

I'm not opposed to these changes, I just can't test them. Perhaps we could
offer both solutions?

xentac:
Who is responsible for merging this one?

aumayr:
Thanks again, @xentac!

PR #737, opened 2018-04-04
handle user budget period formatting error

elohmeier:
If the user enters a wrong period for a budget, e.g.
`2018-01-01 custom "budget" Expenses:Food "M" 300 EUR`
instead of 
`2018-01-01 custom "budget" Expenses:Food "monthly" 300 EUR`
fava returns a HTTP 500 and shows a NotImplementedError on the console, originating in fava/util/date.py", line 220, in number_of_days_in_period.

This PR adds a check in the budget parser to check for a valid period name and creates a Error object which assists the user with correcting the issue.


yagebu:
Hey, thanks for the PR! Could you add a test for this?

fava/core/budgets.py
```diff
@@ -61,6 +61,19 @@ def parse_budgets(custom_entries):
                                 entry.values[1].value,
                                 entry.values[2].value.number,
                                 entry.values[2].value.currency)
+                if budget.period not in [
+                    "daily", "weekly", "monthly", "quarterly", "yearly"
+                ]:
+                    errors.append(
+                        BudgetError(
+                            entry.meta,
+                            'invalid budget period "%s", please use "daily", \
```
aumayr: Can you make the message translatable?

fava/core/budgets.py
```diff
@@ -61,6 +61,19 @@ def parse_budgets(custom_entries):
...
+                        "weekly", "monthly", "quarterly" or "yearly"'
```
aumayr: Should we put the possible values in the error message?

fava/core/budgets.py
yagebu: @aumayr: none of the other error messages are translateable - I wouldn't change that here

aumayr: I think we should start to translate the error messages, at least the ones from Fava.

fava/core/budgets.py
elohmeier: I think it's the most user friendly option, while being clearly redundant.

yagebu:
I've refactored the budget code to also use fava.util.date.Interval and added a similar check in #746. Thanks again

PR #747, opened 2018-04-19
Add a query string to static urls to force cache updates when they change

xentac:
Now no one will ask questions about why their page doesn't render correctly. Modified files will always be forced to be downloaded.

Mostly based on https://gist.github.com/Ostrovski/f16779933ceee3a9d181.


yagebu:
Thanks. Please add a test, also, please remove all the (for Fava) unnecessary codepaths.

yagebu:
Also note that this only works for the JS, not the CSS

xentac:
It looks like it does work for style.css, but not for the inner @imports.

Can we use the npm build step to also bundle css the same way we bundle app.js?

It looks like the recommended way to do this sort of thing is with something like cssmin (https://stackoverflow.com/questions/10036977/best-way-to-include-css-why-use-import). If you're not opposed, I could work on that in a separate PR once this is merged.

xentac:
Unnecessary codepaths removed and test written.

fava/application.py
```diff
@@ -124,6 +125,21 @@ def _inject_filters(endpoint, values):
...
+    if endpoint != 'static':
+        return
+
+    filename = values.get('filename', None)
+    if not filename:
+        return
+
+    file_path = os.path.join(app.static_folder, filename)
+    if os.path.exists(file_path):
+        values['_'] = int(os.stat(file_path).st_mtime)
```
yagebu: maybe replace `_` by `mtime` - no need to be cryptic here

fava/application.py
```diff
@@ -124,6 +125,21 @@ def _inject_filters(endpoint, values):
             values[filter_name] = g.filters[filter_name]
 
 
+@app.url_defaults
```
yagebu: Since we only require this in two places (or really only one right now), I think it would be better to just to this explicitly in the two places instead of having this code that runs everytime that `url_for` is called anywhere. It could be done two ways: 1) add a function for it 2) add a template filter for mtime and then do it in the template

yagebu: I think I'd prefer 1)

fava/application.py
xentac: Done

PR #748, opened 2018-04-19
Mention missing deps for fava in index

Akuukis:
I tested on fresh box that just got fresh Python3.6.
Had to also run this fix to be able to build wheels
on my system (Debian Buster) 
```
ln -s /usr/include/locale.h /usr/include/xlocale.h
```
I guess last fix is quite specific. Details:
https://forum.alpinelinux.org/forum/general-discussion/xlocaleh

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -17,9 +17,11 @@ You can try out an online `demo <https://fava.pythonanywhere.com>`__.
 
 If you are new to Fava and Beancount, begin with the :doc:`usage` guide.
 
-If you are already familiar with Beancount, this is enough to get you up and
-running::
+If you are already familiar with Beancount and have `Python3.6` installed,
+this is enough to get you up and running::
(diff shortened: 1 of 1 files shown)


yagebu:
I'd rather leave it as it is - we also don't tell users how to install Python or pip.

Akuukis:
Ok, let's leave it as is in first page, but document that somewhere else. We should have a page how to install the whole stack - python3.6, pip3, beancount, plugins (visible to fava), fava, perhaps some dedicated user and nginx configuration. It was a struggle to sort that out for myself, but now I could write one for others.

yagebu:
"how to install python 3" is not really something we should explain IMHO.

If you have a working Python installation, installing Fava should be a breeze. The main obstacle might be Beancount, I'll add a link to Martin's installation guide to the "Usage Guide"

PR #760, opened 2018-05-13
[WIP] Integrate flask-login for authentication

bobtwinkles:
This is a WIP branch to integrate Flask-Login and provide native authentication protection instead of requiring hacky docker proxy setups. It mostly works, though there are a few caveats:

  - I have yet to implement a proper database so right now users are hardcoded in to `user.py` which is pretty ungood.
  - This is a breaking change right now, as users will have to have a `fava.environments.production` module in their python path (and we don't ship with one currently since I'm not sure how we want to handle `SECRET_KEY` for production users). 
   - Deployment instructions need to be updated, since even if we ship `fava.environments.production` users will need to supply a custom `SECRET_KEY` somehow


bobtwinkles:
After my latest series of commits, `fava.environments.production` won't cause crashes if it's missing (or at least it shouldn't)

yagebu:
I don't think that this should be provided by Fava. Password-protecting Fava makes little sense without doing SSL too (which we currently don't), and the most servers seem to provide some like basic HTTP auth out of the box AFAIK.

If you want to do this on the Python side and not on your HTTP server, I think using a WSGI middleware would be best. Have you had a look at the available authentication HTTP WSGI middlewares?

PR #765, opened 2018-05-17
Add autocompletion for links in editor and filter

tbm:
beancount 2.1 introduced getters.get_all_links() so we can add autocompletion for links now.

### DIFF
--- a/fava/core/attributes.py
+++ b/fava/core/attributes.py
@@ -15,11 +15,13 @@ class AttributesModule(FavaModule):
         self.accounts = None
         self.currencies = None
         self.payees = None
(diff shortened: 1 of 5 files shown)


tbm:
I'm not sure why AppVeyor installs beancount==2.0.0 since I changed setup.py

yagebu:
> I'm not sure why AppVeyor installs beancount==2.0.0 since I changed setup.py

Apparently tox doesn't check or update the dependencies in the cached virtual environment (which I find pretty stupid).

yagebu:
Thanks!

PR #776, opened 2018-06-24
use cheroot wsgi server

yagebu:


aumayr:
Why the change to `cheroot`?

yagebu:
> WARNING: Do not use the development server in a production environment.
> Use a production WSGI server instead.

PR #777, opened 2018-06-24
generalize link_statements plugin to use "document" metadata keys

### LINKED ISSUE #740: support for generic documents (not only statements) associated to transactions
I'm looking into how Fava support documents referenced from the journal, and it is not clear to me if there is a desire to support generic documents, as opposed to only bank-style statements.

In particular, I'm interested in the ability to accept "receipts", "invoices", etc., that will generally be associated to Expenses transactions, or Assets:Receivable, etc. Is that something that Fava wants to support?

If so, I think it would be nice to rename the link_statement plugin into something more generic, and not use `#statement` as a tag, but maybe something like `#document`. I suspect that the name comes from the `document` directive of Beancount itself which, on the other hand, really seems to be statement-specific, as documents are conceptually inserted by Beancount *in between* transactions, rather than *on* them.

zacchiro:
Close #740

### DIFF
--- a/fava/plugins/link_documents.py
+++ b/fava/plugins/link_documents.py
@@ -1,8 +1,9 @@
-"""Beancount plugin to link statements to documents.
+"""Beancount plugin to link transactions to documents.
 
-It goes through all transactions with a `statement` metadata-key, and tries to
-match them to Document entries. It then adds a link to the transaction and the
-document, and the tag "#statement" to the document.
+It goes through all transactions with a `document` metadata-key, and tries to
+associate them to Document entries. It then adds a link from transactions to
+documents, as well as the "#linked" tag.
+
 """
 
 import collections
@@ -11,13 +12,13 @@ from os.path import join, dirname, normpath, basename
 from beancount.core import data
 from beancount.core.compare import hash_entry
 
-StatementDocumentError = collections.namedtuple('StatementDocumentError',
-                                                'source message entry')
+DocumentError = collections.namedtuple('DocumentError',
+                                       'source message entry')
 
-__plugins__ = ['link_statements']
+__plugins__ = ['link_documents']
 
 
-def link_statements(entries, _):
+def link_documents(entries, _):
     errors = []
 
     all_documents = [(index, entry) for index, entry in enumerate(entries)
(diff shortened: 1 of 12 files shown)


zacchiro:
OK, here's the PR addressing what we discussed in #740.

Using "document" as metadata key prefix causes some naming clashes in the code (and documentation), so I've picked "disk_doc" to differentiate documents expected to be on disk and document (as Beancount entries).

Also, I've renamed the plugin to link_documents.py, the idea being that people might easily keep around the old link_statements.py file if they want to transition more gracefully. YMMV.

zacchiro:
Oh, also, I'm not entirely sure you want to have the #document tag for documents linked by this plugin (a document tagged document sounds weird...), but I haven't found convincing ideas for a better name. Maybe "linked_document"? Ideas welcome.

fava/plugins/link_documents.py
```diff
@@ -1,8 +1,8 @@
-"""Beancount plugin to link statements to documents.
+"""Beancount plugin to link documents on disk to document entries.
```
yagebu: It rather links transactions to documents, doesn't it?

yagebu:
The code in the journal template and some other places also needs to be changed.

> Maybe "linked_document"? Ideas welcome.

Maybe just "linked"?

yagebu:
@zacchiro: ping

zacchiro:
@yagebu thanks for the ping, I was on VAC and this PR slipped my mind.

I've just updated it, fixing the description, using `#linked` as you suggested, and fixing all other places where I could find mention of the old `#statement` tag. If you can spot other places where adaptation is needed, please shout.

yagebu:
Thanks. Merged without the changes to the translations, as they are maintained on POEditor.com

vloris:
I'm having somee problems with the current fava release (1.9):

- The help-page still shows link_statements as plugin-name
- If you drag-and-drop documents on a transaction, it creates metadata named 'statement', but the new link_documents plugin doesn't look at 'statement', but at 'document' metadata.

So, I think there are still two changes necessary: update the documentation and change the metadata-key used by the statement-upload process. Does that sound correct?

zacchiro:
@vloris thanks for the info. The first issue you mention should be fixed by #858 .
I'm not sure about the second, as I don't use the drag&drop UI, but it does sound right. I'm not the fava maintainer, but I suspect a PR fixing that would be welcome too.

PR #782, opened 2018-06-29
Update regex for account names

### LINKED ISSUE #775: Transaction entry form disallows valid account name
According to the Beancount docs:  "Each component of the account names begin with a capital letter or a number and are followed letters, numbers or dash (-) characters."

However the transaction data entry form does not allow a component to begin with a number.

I'd suggest making the following change in fava/templates/_entry_forms.html:
`pattern="(?:[A-Z0-9][A-Za-z0-9\-]+)(?::[A-Z0-9][A-Za-z0-9\-]*)+">`

Cheers.

tbm:
Fixes #775

### DIFF
--- a/fava/static/javascript/codemirror/mode-beancount.js
+++ b/fava/static/javascript/codemirror/mode-beancount.js
@@ -73,7 +73,7 @@ CodeMirror.defineSimpleMode('beancount', {
       token: 'date',
     },
     {
-      regex: /(?:[A-Z][A-Za-z0-9\-]+)(?::[A-Z][A-Za-z0-9\-]*)+/,
+      regex: /(?:[\p{Lu}][\p{L}\p{Nd}\-]*)(?::[\p{Lu}\p{Nd}][\p{L}\p{Nd}\-]*)+/,
       token: 'account',
     },
     {
--- a/fava/templates/_entry_forms.html
+++ b/fava/templates/_entry_forms.html
@@ -1,5 +1,5 @@
 {% macro input_account(account) %}
-<input type="text" class="account" placeholder="{{ _('Account') }}" list="accounts" value="{{ account or '' }}" required pattern="(?:[A-Z][A-Za-z0-9\-]+)(?::[A-Z][A-Za-z0-9\-]*)+">
+<input type="text" class="account" placeholder="{{ _('Account') }}" list="accounts" value="{{ account or '' }}" required pattern="(?:[\p{Lu}][\p{L}\p{Nd}\-]*)(?::[\p{Lu}\p{Nd}][\p{L}\p{Nd}\-]*)+">
 {% endmacro %}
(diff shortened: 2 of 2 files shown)


yagebu:
Are these character classes supported in most (modern) browsers?

tbm:
> Are these character classes supported in most (modern) browsers?

I'd think so. Pretty much everything is UTF-8 compatible now.

Where do you see potential problems? If so, I can do some tests.

yagebu:
> I'd think so. Pretty much everything is UTF-8 compatible now.

Supporting unicode at some basic level is quite unrelated to supporting Unicode character categories in regular expressions. At the earliest it was added in ES6 and I didn't directly find any information on how well it is supported, hence my question.

As far as I can tell, the editor mode expression needs a unicode flag to work.

tbm:
Ok, I don't really know. I did limited testing with an umlaut but that's about it.

yagebu:
Can you do some more testing? ;)

tbm:
It seems the autocompletion assumes a specific width of the character.  See this screenshot. The yellow/orange background doesn't cover the full character.

autocomplete

yagebu:
And it works in the editor too?

tbm:
> And it works in the editor too?

I can enter text and save it and it looks ok.

There's a minor problem with the alignment because of different size of characters.
alignment

yagebu:
That doesn't look like it's working - the account names aren't coloured

tbm:
> That doesn't look like it's working - the account names aren't coloured

Coloured how?  The other (non-UTF) accounts aren't coloured either as far as I can tell.  And tab completion works:
tab-completion

yagebu:
> Coloured how? The other (non-UTF) accounts aren't coloured either as far as I can tell.

They aren't coloured because you broke the syntax ;)

yagebu:
I've merged the change to the template, thank you! It would be cool if you could make the change to the editor syntax work as well :)

tbm:
I see you merged the template but not `fava/static/javascript/codemirror/mode-beancount.js`. Is that not needed?

tbm:
> It would be cool if you could make the change to the editor syntax work as well :)

Any pointers on where to look?

yagebu:
> I see you merged the template but not fava/static/javascript/codemirror/mode-beancount.js. Is that not needed?

It is needed, but the change you made breaks the syntax and doesn't work at all.

> Any pointers on where to look?

Not really - trial-and-erroring on how to provide a unicode RE to codemirror, maybe?

yagebu:
I'll close this, if you ever want to fix the editor RE, just reopen or open a new PR

yagebu:
I'll have to revert this, the regular expression doesn't work at all

tbm:
Hmm, what problems did you see?

yagebu:
The regular expression doesn't compile

Issue #745, opened 2018-04-10
Add nicer net worth chart, showing assets, liabilities and net worth in one chart


tbm:
A net worth chart like this would be great, i.e. showing assets, liabilities and net worth in one chart:

![gl4n3vi 1](https://user-images.githubusercontent.com/1556037/38586728-3dcd3ff6-3d1f-11e8-94b0-59fd34eb9f0a.png)

Source: http://imgur.com/gL4N3vI (which was posted somewhere on reddit).

@yagebu do you have time for this?  If not, maybe time for me to learn d3 (might take a while though).

yagebu:
@tbm: probably won't have time for this soon

mariosangiorgio:
@tbm from your message it seems you could have attempted to implement this.

Have you had a chance to give it a try?

tbm:
Not yet, no.

yagebu:
I'm closing this, since it's become unfeasible with the net worth chart possibly displaying multiple lines.

PR #789, opened 2018-07-06
Improved net worth visualisation

mariosangiorgio:
This PR has been inspired by https://github.com/beancount/fava/issues/745, where @tbm suggested to visualisation net worth, assets and liabilities as it's done by YNAB.

I don't know [d3js](https://d3js.org/) but I figured out that we could rely on what `LineChart` already does to display different lines for different currencies.

This sample file
```
option "operating_currency" "GBP"
2018-01-01 open Liabilities:Mortgage GBP
2018-01-01 open Equity:OpeningBalances GBP
2018-01-01 open Assets:Current GBP
2018-01-01 open Income:Salary GBP
2018-01-01 open Expenses:Interests GBP
2018-01-01 * "Opening balances" ""
  Assets:Current 11000 GBP
  Liabilities:Mortgage -10000 GBP
  Equity:OpeningBalances

2018-02-01 * "Salary" ""
  Income:Salary
  Assets:Current 1500 GBP
2018-02-01 * "Interests" ""
  Liabilities:Mortgage
  Expenses:Interests 500 GBP
2018-02-10 * "Mortgage" ""
  Liabilities:Mortgage
  Assets:Current -1000 GBP

2018-03-01 * "Salary" ""
  Income:Salary
  Assets:Current 1500 GBP
2018-03-01 * "Interests" ""
  Liabilities:Mortgage
  Expenses:Interests 500 GBP
2018-03-10 * "Mortgage" ""
  Liabilities:Mortgage
  Assets:Current -1000 GBP
```
results in this chart being drawn
screen shot 2018-07-07 at 00 24 48

@yagebu do you think an approach like this could make sense?
I am not completely convinced this is the best solution, but now knowing d3js it was the easiest way I could get to the results we wanted.

### DIFF
--- a/fava/core/charts.py
+++ b/fava/core/charts.py
@@ -155,6 +155,7 @@ class ChartModule(FavaModule):
         inventory = CounterInventory()
         assets_inventory = CounterInventory()
         liabilities_inventory = CounterInventory()
+
         def reduce_inventory(date, inventory):
             return {
                 currency: inventory.reduce(convert.convert_position,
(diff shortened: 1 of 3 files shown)


yagebu:
Sorry for not having a look at this any earlier.

I'm not convinced by the look of it, for me the two new lines rather add a bit of confusion (and for my data they're not that helpful really). What do others think?

Implementation-wise this adds a bit much duplicate code - sticking to the `balances` chart type would be better IMHO.

mariosangiorgio:
Yeah, I think this works well only when the lines are well separated, e.g. for people who have a mortgage. I agree that in the general case this is going to be confusing.

PR #797, opened 2018-08-05
Fixup in the journal table for link_statements generalisation

jimmyxu:
Following up #777. Document entries were not being hidden in the journal, and metadata links were broken.

Not sure if the `/<bfile>/statement/` routing needs a new name. Ping @zacchiro :)


zacchiro:
I *guess* it should, but I don't feel confident enough to make that call myself, as I've never looked into that code. If others feel like it, by all means go ahead!

(And thanks for spotting this regression. I'm using the newly named plugin in production, but I didn't notice it unfortunately.)

PR #798, opened 2018-08-07
Fix for variable substitution in time filter

tf198:
Version: master
Bug: currently fails to parse `quarter` and `week` in time filter.

`parse_date()` requires lowercase but `substitute()` returns uppercase 'Q' and 'W'


aumayr:
Thanks @tf198 for catching and fixing that bug!

PR #799, opened 2018-08-07
RFC - fiscal year support

tf198:
Have had a go at implementing time filter support for arbitrary accounting periods. 

Here in Australia we use 30th June as end of the tax year so by adding

    2018-01-01 custom "fava-option" "fiscal-year-end" "06-30"

to the top of my beanfile I can then use time filters like `FY2018` for the tax year to 2018-06-30 and `FY2017-Q4` or `fiscal_quarter-1` for preparing my April - June BAS statements.

Seems to work well for me at the moment but involves quite a few changes so needs review and comment...


aumayr:
Thanks for the contribution, @tf198! This case is quite common IMHO, because in some countries companies can choose arbitrary `fiscal-year-end`s.

Will wait for @yagebu to chip in, but to me the PR looks good, and special thanks for the test-converage and documentation!

fava/util/date.py
```diff
@@ -217,9 +243,76 @@ def parse_date(string):  # pylint: disable=too-many-return-statements
...
+        quarter: one of [None, 1, 2, 3 or 4]
+        fye: End date for period in "%m-%d" format
+
+    Returns:
+        A tuple (start, end) of dates.
+
+    """
+    if fye is None:
+        from flask import g
+        fye = g.ledger.fava_options['fiscal-year-end']
```
yagebu: I'd prefer it if `util.date` didn't depend on the active request context - better pass the fiscal year end as an explicit argument to the functions in this module

tf198: I agree, but from groking the code you'd need to inject the arguments in the `FavaLedger.filter()` method and adjust the signatures on all the `Filter` classes - I didn't want to make that call!

Am happy to refactor if you can give me some guidance on what context you want passed down to the filters - am guessing the `fava_options` object for future expansion?

tf198: I tried moving the context access to the `TimeFilter` class but then the filter tests started failing as there is no valid context.  
Have just created another pull request at #807 which makes the Filters stateful and would allow for generic configurable filters - will wait to hear your input on that before continuing here...

tf198:
Merged #807 and hooked it up.  All seems to be working as expected.
Probably needs a squash before merging...

yagebu:
Thank you very much for the contribution!

PeterDavidson:
Thanks - I really like this feature (I'm from Australia too). I was wondering how difficult it would be to extend the support to include the chart intervals. It would be really useful for me if we could show charts by fiscal year. Fiscal quarter would not be needed in Australia, but in those places that don't align neatly to standard quarters then fiscal quarters would be useful too. I've had a quick look but I'm not familiar enough with the code yet to know where to start. The control has an id of 'chart-interval' if that helps.

tf198:
I agree @PeterDavidson, this needs to be implemented.  Will try and have a look into this, though the display code is not really my forte.  Will open a separate issue it it needs to be bumped to someone else..

PR #800, opened 2018-08-11
Wrap text in the source row

alexiri:
Source lines may be long and they get cut off inside the pre, as text is not wrapped and scrolling is not possible. This change wraps text so the full line is displayed.

### DIFF
--- a/fava/static/css/ingest.css
+++ b/fava/static/css/ingest.css
@@ -18,6 +18,11 @@
 
 .ingest-row .source pre {
   font-size: .9em;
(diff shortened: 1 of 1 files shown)


fava/static/css/ingest.css
```diff
@@ -18,6 +18,11 @@
 
 .ingest-row .source pre {
   font-size: .9em;
+  white-space: pre-wrap;       /* Since CSS 2.1 */
```
yagebu: Only add this line - we don't support crap browsers

yagebu:
Thanks!

PR #807, opened 2018-08-26
Configurable filters

tf198:
Possible solution for #799.

The `Filter` objects may need access to the options from the beancount file - currently the beancount options are being passed to the `apply()` method.  #799 requires access to the `fava_options` during the `set()` call, which is all a bit messy.

In my mind it is better to configure the filter instances once the files have been read and the filters can then use that however they need which avoids any flask globals black magic.

I think file reloads should be handled gracefully but need one of the maintainers to confirm.


yagebu:
You're quite right - can you merge this into #799?

PR #819, opened 2018-09-05
Transaction form: Make narration wider than payee

aumayr:
This PR makes the narration slightly wider than the payee field in the transaction form, and uses the full width for the form in the extract view when source is hidden.

### DIFF
--- a/fava/static/css/entry-forms.css
+++ b/fava/static/css/entry-forms.css
@@ -22,10 +22,16 @@
 }
 
 .entry-form input.account,
-.entry-form input[name='narration'],
+.entry-form input[name='narration'] {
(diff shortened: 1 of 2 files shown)


yagebu:
> uses the full width for the form in the extract view when source is hidden.

I tried this at some point and decided against it on purpose - I found the entry forms to be a bit unwieldy at full width. If it's not wide enough for you currently I'm fine with increasing the width a bit, but I think it should be bounded.

aumayr:
> If it's not wide enough for you currently I'm fine with increasing the width a bit, but I think it should be bounded.

Makes sense - the max width is now 1000px.

PR #823, opened 2018-09-20
Guard lazy _load_file with threading.Lock.

### LINKED ISSUE #821: Segfault at fava initialization.
Steps to reproduce:
1. Start fava
2. While fava initializes make many simultaneous requests to any page (even `/beancount/help/`)
3. After about 5-6 requests fava either hangs (couldn't kill with C-c) or segfaults.

```
#0  0x00007f6ffd68896d in build_grammar_error_from_exception () at beancount/parser/grammar.y:61
#1  0x00007f6ffd6897e6 in yyparse () at beancount/parser/grammar.y:338
#2  0x00007f6ffd68b98c in parse_file (self=<optimized out>, args=<optimized out>,
    kwds=<optimized out>) at beancount/parser/parser.c:97
#3  0x000000000056ed53 in PyCFunction_Call ()
...
#64 0x0000000000599b21 in ?? ()
#65 0x00000000005a6dbe in PyObject_Call ()
#66 0x00000000005e3db2 in ?? ()
#67 0x00007f70014bdf2a in start_thread () from /lib/x86_64-linux-gnu/libpthread.so.0
#68 0x00007f7000ff5edf in clone () from /lib/x86_64-linux-gnu/libc.so.6
```

sarg:
Fixes #821

### DIFF
--- a/fava/application.py
+++ b/fava/application.py
@@ -14,6 +14,7 @@ Attributes:
 import datetime
 import functools
 import inspect
+import threading
 import os
 import os.path
 from io import BytesIO
@@ -74,6 +75,9 @@ REPORTS = [
     'statistics',
     'trial_balance',
 ]
+
+
+LOAD_FILE_LOCK = threading.Lock()
 
 
 def _load_file():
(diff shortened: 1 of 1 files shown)


yagebu:
I think the problem could still occur when the files are reloaded - the lock should also be more local around the place where it actually matters I think, namely the call to Beancount's loader.

sarg:
If I understood you correctly you suggest to put lock into `FavaLedger.load_file` (`core/__init__.py`). If this is the case then I have tried it and it doesn't work correctly. The reason for that is the fact that many FavaLedger instances for the same beancount file are created because of race condition.

yagebu:
Ah I see - there still needs to be some lock for the potential reloads of the Beancount file. I'll rework the watcher a bit in the coming week and should be able to address that then.

It would be ideal if you could write a test that would fail on this race condition if that's not too hard to do.

yagebu:
Thanks for this!

PR #826, opened 2018-09-21
Reference insert-entry option in help/import.md.

### LINKED ISSUE #822: Add option 'import-destination'.
My ledger is split between few beancount files. My current setup presumes that main file is used only to declare options, open/close accounts and include other files. It would be nice to have `import-destination` option which points to beancount file where new entries should be written.

sarg:
Fixes #822


yagebu:
Thanks!

PR #827, opened 2018-09-26
entry-forms: set type of amount input to text, fix #824

### LINKED ISSUE #824: Adding transactions broken on mobile
#818 removed the currency field from the new entry form. The field where now both amount and currency shall be entered (which is not obvious) has type "tel" which means on mobile devices you get keyboard where you can type only numbers and a few special characters typically used in phone numbers which is not enough to enter the currency. If you omit the currency, you only get an unhelpful internal server error when submitting the form (it works though if I enter the currency in the field via copy&paste, so it is not generally broken).

michitux:
As the amount field is not just for numbers but also the currency, the "tel" type is not sufficient for that and in particular means on-screen keyboards will not offer any way to enter the currency.


yagebu:
Thanks!

PR #833, opened 2018-10-12
Fix saving of balance statements

michitux:
ba73750a2850b46dd2c387ce377de0100aedcb0a erroneously removed the name attribute of the number input for balance statements which broke the import of balance statements.


aumayr:
Thanks @michitux !

Issue #840, opened 2018-10-23
set chart y axis range


alexmat:
Is there a way to set the net-worth chart y axis range? It's very hard to see changes in net worth with the y axis starting at zero. Would be great if it was +/- 10% or so of the lowest and highest net-worth for the chart period.

yagebu:
Sounds good! Keen to implement this?

alexmat:
Looks like the change would have to happen in charts.js and I'll need to look into D3. Does that sound like I'm on the right track? I can give it a shot but might need some hand holding :)

yagebu:
Thanks for giving it a shot! :)

That sounds like you're already on the right track. I guess the relevant part will be related to the scale (or maybe the axis) of the chart, so part of d3-scale and d3-axis, both of which are well-documented on their respective Github pages. I guess increasing the domain of the scale should do it. If you need any further help, just ask :)

alexmat:
Took me a bit to get up to speed, but I think I have a first draft. Look forward to hearing your feedback.

PR #842, opened 2018-10-27
line charts now span y-axis from min to max value

alexmat:
This is a draft for enhancement request #840 

I chose a 3% margin after trying a few values. This seemed to me to use up the available space without feeling cramped.

### DIFF
--- a/fava/static/javascript/charts.js
+++ b/fava/static/javascript/charts.js
@@ -540,13 +540,13 @@ class LineChart extends BaseChart {
       max(this.data, s => s.values[s.values.length - 1].date),
     ]);
 
-    // Span y-axis min to max value plus margin as percent of either value
-    this.yMarginPercent = 0.03;
-    this.minDataValue = min(this.data, d => min(d.values, x => x.value));
-    this.maxDataValue = max(this.data, d => max(d.values, x => x.value));
+    // Span y-axis as max minus min value plus 3 percent margin
+    const minDataValue = min(this.data, d => min(d.values, x => x.value));
+    const maxDataValue = max(this.data, d => max(d.values, x => x.value));
+    const yMarginValue = (maxDataValue - minDataValue) * 0.03;
     this.y.domain([
-      this.minDataValue - (this.minDataValue * this.yMarginPercent),
-      this.maxDataValue + (this.maxDataValue * this.yMarginPercent)
+      minDataValue - yMarginValue,
+      maxDataValue + yMarginValue,
     ]);
 
     this.selections.lines = this.canvas.selectAll('.line')
--- a/fava/static/javascript/charts.js
+++ b/fava/static/javascript/charts.js
@@ -540,13 +540,12 @@ class LineChart extends BaseChart {
       max(this.data, s => s.values[s.values.length - 1].date),
     ]);
 
-    // Span y-axis as max minus min value plus 3 percent margin
+    // Span y-axis as max minus min value plus 5 percent margin
     const minDataValue = min(this.data, d => min(d.values, x => x.value));
     const maxDataValue = max(this.data, d => max(d.values, x => x.value));
-    const yMarginValue = (maxDataValue - minDataValue) * 0.03;
     this.y.domain([
-      minDataValue - yMarginValue,
-      maxDataValue + yMarginValue,
+      minDataValue - (maxDataValue - minDataValue) * 0.05,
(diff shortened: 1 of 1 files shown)


fava/static/javascript/charts.js
```diff
@@ -539,9 +539,14 @@ class LineChart extends BaseChart {
       min(this.data, s => s.values[0].date),
       max(this.data, s => s.values[s.values.length - 1].date),
     ]);
+
+    // Span y-axis min to max value plus margin as percent of either value
+    this.yMarginPercent = 0.03;
```
yagebu: Don't think we need a variable for this, just use it inline

fava/static/javascript/charts.js
```diff
@@ -539,9 +539,14 @@ class LineChart extends BaseChart {
...
+    this.minDataValue = min(this.data, d => min(d.values, x => x.value));
```
yagebu:
```suggestion
    const minDataValue = min(this.data, d => min(d.values, x => x.value));
```

fava/static/javascript/charts.js
```diff
@@ -539,9 +539,14 @@ class LineChart extends BaseChart {
...
+    this.maxDataValue = max(this.data, d => max(d.values, x => x.value));
     this.y.domain([
-      Math.min(0, min(this.data, d => min(d.values, x => x.value))),
-      Math.max(0, max(this.data, d => max(d.values, x => x.value))),
+      this.minDataValue - (this.minDataValue * this.yMarginPercent),
+      this.maxDataValue + (this.maxDataValue * this.yMarginPercent)
```
yagebu: Our linter complains, you should add a comma at the end of the line

fava/static/javascript/charts.js
```diff
@@ -539,9 +539,14 @@ class LineChart extends BaseChart {
...
     ]);
+
+    // Span y-axis min to max value plus margin as percent of either value
+    this.yMarginPercent = 0.03;
+    this.minDataValue = min(this.data, d => min(d.values, x => x.value));
+    this.maxDataValue = max(this.data, d => max(d.values, x => x.value));
     this.y.domain([
-      Math.min(0, min(this.data, d => min(d.values, x => x.value))),
-      Math.max(0, max(this.data, d => max(d.values, x => x.value))),
+      this.minDataValue - (this.minDataValue * this.yMarginPercent),
```
yagebu: I think the margin should scale with the difference of min and max, not just the min and max. Let's say the min is 0, this wouldn't do anything.

yagebu:
Thanks! I've added some comments

alexmat:
Thanks for the great feedback! I applied the changes you mentioned but I still used a yMargin constant. My argument in favor of it is that the value is used more than once and involves a "magic value". I am happy to inline it if you still feel it should go that route.

fava/static/javascript/charts.js
alexmat: My argument in favor of it is that the value is used more than once and involves a "magic value". I am happy to inline it if you still feel it should go that route.

fava/static/javascript/charts.js
alexmat: How do I run the linter so I can check before commiting?

fava/static/javascript/charts.js
alexmat: Good catch. Fixed.

fava/static/javascript/charts.js
alexmat: Makes sense. Fixed.

fava/static/javascript/charts.js
alexmat: It also saves an extra calculation haha

fava/static/javascript/charts.js
aumayr: In the root folder of the project: `make lint`

alexmat:
```
  lint: commands succeeded
  congratulations :)
```
 Looks like it's happy. Thanks.

fava/static/javascript/charts.js
alexmat: After playing with the values a bit, 5% looks better. Code also looks nicer with the value inline. I'd advocate for the variable more if it was used somewhere outside the local scope, as it is, it seems clear enough.

yagebu:
Thanks again :)

PR #849, opened 2018-12-13
Reduce docker image size and fix libpng CVEs

adamgibbins:
Upgrades python to pull in newer alpine which fixes a load of  libpng CVEs (https://snyk.io/vuln/SNYK-LINUX-LIBPNG-174098 and https://snyk.io/vuln/SNYK-LINUX-LIBPNG-174252).

Additionally removes apk cache to remove need to clean it up later and purges pyc files to reduce the size.

### DIFF
--- a/contrib/docker/Dockerfile
+++ b/contrib/docker/Dockerfile
@@ -1,13 +1,13 @@
-FROM python:3.6.3-alpine
+FROM python:3.6.7-alpine
 
 ENV BEANCOUNT_INPUT_FILE ""
 ENV FAVA_OPTIONS "-H 0.0.0.0"
 ENV FINGERPRINT "4e:65:3e:76:0f:81:59:85:5b:50:06:0c:c2:4d:3c:56:53:8b:83:3e:9b:fa:55:26:98:9a:ca:e2:25:03:92:47"
-ENV BUILDDEPS "libxml2-dev libxslt-dev gcc musl-dev mercurial git nodejs make g++"
+ENV BUILDDEPS "libxml2-dev libxslt-dev gcc musl-dev mercurial git npm make g++"
 ENV RUNDEPS "libxml2 libxslt"
 
 RUN cd /root \
-        && apk add --update $BUILDDEPS $RUNDEPS \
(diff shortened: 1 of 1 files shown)


contrib/docker/Dockerfile
```diff
@@ -1,13 +1,13 @@
-FROM python:3.6.3-alpine
+FROM python:3.6.7-alpine
```
yagebu: Why not use a more up-to-date version? And why fix it in the first place (that will probably just lead to similar security problems in the future)?

adamgibbins: To go higher would mean moving to 3.7, I'm not familiar enough with Python to make the call of doing that. My interest was rather the newer alpine, not python -- but the hub only had newer alpines if we upgraded python too.

yagebu: The build's not deterministic in any case - go with python:alpine

contrib/docker/Dockerfile
```diff
@@ -19,10 +19,10 @@ RUN cd /root \
         && rm fava/docs/changelog.rst \
         && python3 -mpip install ./fava \
         && python3 -mpip uninstall --yes pip \
-        && find /usr/local/lib/python3.6/site-packages -name *.so -print0|xargs -0 strip -v \
+        && find /usr/local/lib/python3.6/site-packages -name '*.so' -print0|xargs -0 strip -v \
+        && find /usr/local/lib/python3.6/site-packages -name '*.pyc' -delete \
```
yagebu: This was intentionally removed - don't delete Python bytecode

yagebu:
@adamgibbins: ping

contrib/docker/Dockerfile
adamgibbins: Sorry I didn't notice it was previously removed. Can I ask why? The bytecode is regenerated on first execution, deleting upfront reduces the size of the image thats hanging around and duplicated multiple places by a fair bit at cost of slightly slower startup.

yagebu:
I'll replace the Dockerfile with a simpler one to make it more easily maintainable

PR #859, opened 2019-01-02
setup.cfg: ship tests in distribution tarball

zacchiro:
rationale: 3rd party distributors of Fava (e.g., GNU/Linux distributions) might
want to run tests at package build time, for quality assurance purposes

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -58,7 +58,3 @@ excel =
     pyexcel-ods3>=0.1.1
     pyexcel-xls>=0.1.0
     pyexcel-xlsx>=0.1.0
-
-[options.packages.find]
-exclude =
-    tests


yagebu:
We shouldn't ship a Python package named `tests`, they need to be included differently

zacchiro:
Right. It should be shipped as something like `fava.tests`.

yagebu:
No it shouldn't be shipped as a package at all

zacchiro:
Well, you can distribute tests as part of the fava package, but OK, you don't want to :)
Do you want to distribute them as setuptools package_data/data_files then? If so, which one?

yagebu:
If we'd include them in the package, we'd have to include all test dependencies as well. I don't see any reason why the tests should be part of the fava package.

yagebu:
Since we started using `setuptools_scm` we (more or less accidentally) already include all test files, so this PR is not really necessary. I should add some excludes to the manifest file though.

I'm still not sure how I feel about this - it seems a bit weird to ship all tests and data files for the tests to all users.

zacchiro:
@yagebu feel free to close the PR if it's unnecessary.
Note that the fava 1.9 tarball doesn't include the tests though, so unless the change you mention happened after that release, something will need to change before the next release to fix the underlying issue.

Regarding your mixed feelings: I'm not sure how else 3rd party distributors of fava could get the tests. Note that tests will not be *installed* in, say, .deb packages that end on user machines; they will just be testable at package build-time thanks to availability in the distribution tarball.

yagebu:
The correct files should be included now - can you see if this works?

zacchiro:
On Mon, Jan 07, 2019 at 11:09:28AM -0800, Jakob Schnitzer wrote:
> The correct files should be included now - can you see if this works?

It works great now, thank you !

PR #861, opened 2019-01-06
Fix generated metadata uploading statements

### LINKED ISSUE #860: Remnants of statement metadata
With #740, the `statement:` metadata key was renamed to the more generic `document:` key, but there are some remnants of the old key in the code (tested on 1.9):

* When uploading a document through drag-and-dropping on a transaction, fava adds a `statement` metadata instead of `document`.
* The "Fava's features" help page describes the old metadata and inexistent `fava.plugin.link_statements` plugin.

vianney:
The `statement` metadata key is not recognized as special by fava
anymore, since it has been renamed to `document`.  However, fava was
still inserting the now-obsolete `statement` metadata when uploading a
document by dropping it on a transaction in the journal view.

This patch makes fava add a `document` metadata instead of a `statement`
metadata.

Fixes #860

### DIFF
--- a/fava/help/features.md
+++ b/fava/help/features.md
@@ -135,7 +135,7 @@ be tagged with `#discovered` and can be filtered in the Journal:
 ### Uploading statements
 
 When dropping a file on a transaction (or one of its postings) in the Journal,
-the file will be uploaded as described above, and a `statement`-metadata-entry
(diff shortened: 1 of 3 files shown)


yagebu:
Thanks!

Issue #759, opened 2018-05-13
extensions: support user reports


seltzered:
I'm trying to see if there's a need among others for an option for fava to look for additional templates in a separate directory (outside the fava codebase) - the idea is that this would facilitate 'custom reports'.

I'm pretty new to the fava code but from looking at it so far, implementing this would entail:
- A 'custom' directive so that one's beancount file points to the directory of templates
   - e.g. `2010-01-01 custom "fava-option" "custom-templates-dir" "$LEDGERDIR/fava_user_templates/"`
- app.routes dynamically created for the detected user templates. 
- Modifying things so 'navigation_bar' has an insert for 'User' templates.

Is there interest in such a thing? Am I missing something fundamental in fava/beancount that would already do this? 
 
Motivation: I'm finding areas where I want to have mini reports that are specific to my own needs (e.g. watching balance among work expenses vs. reimbursement income over time, wanting to see both a 'raw' and a 'filtered' version of my income statement, showing percent profit/loss on non-commodity investments, etc.). I could write/execute my own separate python scripts to do this, but ideally I want the presentation of these things to be within fava. 

I've considered using beancount 'plugins' to do filtering but there's times where I still want to see the unfiltered reports.

I'm also trying to understand fava's 'extensions' ([issue](https://github.com/beancount/fava/pull/477), [commit](https://github.com/beancount/fava/pull/477)) - one thought is that custom reports could instead be implemented via extensions if some extra hooks are added within application.py.

aumayr:
Have you seen the [Custom links in the sidebar](https://fava.pythonanywhere.com/example-with-budgets/help/features/)-feature in the help pages? They are pretty powerful, as you can add a link to a  BQL-query to the sidebar as well. 

If this does not go far enough for you, I think we could extend the Extension mechanism to allow for custom views to be added to the sidebar. (I would rather go the Extension way than the autodetect-templates-way, as this is already a "known" mechanism to extend the functionality of Fava and does not add yet another way to do that.)

seltzered:
Thanks, I hadn't seen that feature - custom sidebar links to filtered statements will cover some of my needs. Will look into the adding to the extension mechanism if that doesn't go far enough.

yagebu:
This sort of thing was always something that I thought the extension system should support at some point, so a PR adding this to extensions would be very welcome! :)

seltzered:
I'm currently playing around with this idea. Code is too ugly to share at the moment, the approach I've been trying to take is:
[...]

PR #866, opened 2019-01-17
Use an application factory instead of global app.

seltzered:
Use an application factory instead of using a global app instance. The intent is to make it easier to write tests for fava that use other sample beancount files, which I expect when working on https://github.com/beancount/fava/issues/759 

I did consider doing a cleaner implementation that involved moving route methods to a separate `routes.py` file, changing their decorator association to a blueprint / registering a blueprint in `create_app`, but this seemed to cause issues where the app.url_map get prepended with 'fava.'. Could be done, though from my understanding that means tests would then have to change from e.g. `flask.url_for("report"...` to `flask.url_for("fava.report"...`

### DIFF
--- a/fava/application.py
+++ b/fava/application.py
@@ -30,6 +30,7 @@ from flask import (
     render_template_string,
 )
 import flask
+from flask import current_app
 from flask_babel import Babel
 import markdown2
 import werkzeug.urls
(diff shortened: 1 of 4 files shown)


yagebu:
Hey, thanks for pushing for this feature! I'm not sure if such a big refactor is really necessary. Wouldn't something like the following work (as a fixture to be given to tests)?

```python
import app, _load_file from fava.application

def create_app(beancount_file):
    app.config["BEANCOUNT_FILES"] = [beancount_file]
    _load_file()
```

seltzered:
Huh, good point! Lemme try that out another branch where I was playing with extensions/other samples. In my first attempt of extension reports I was changing the app.config's `BEANCOUNT_FILES` but was probably forgetting to call `_load_file` afterward.

seltzered:
Ok, making a fixture for the other beancount example (e.g. `extension_report_file`) then doing

`
    app.config["BEANCOUNT_FILES"] = [extension_report_file]
    _load_file()
`

in the test seems to work and is simpler for now. Closing this, thanks.

PR #870, opened 2019-01-21
Fix documents metadata with relative paths

vianney:
Documents specified with the `document` metadata may contain relative
paths.  Such paths were resolved against the directory containing the
top beancount file.  For consistency with the beancount `document`
directive and the link_documents plugin, this patch resolves relative
paths against the directory containing the beancount file containing the
transaction.


yagebu:
Thanks!

PR #872, opened 2019-01-25
Extension reports

yagebu:


yagebu:
@seltzered: what's the state of this?

seltzered:
Squashed / cleaned up commits last weekend. Can't think of anything else to add at the moment beyond perhaps polishing the sample extension further.

The one thing I'd like some feedback on is whether my use of a semaphore'd `active_ext` interface between the extension's python and associated template is a good idea or not. See https://github.com/beancount/fava/blob/479c5602cd072976477dbf85d00242718ad69a60/fava/core/extensions.py#L55 , https://github.com/beancount/fava/blob/479c5602cd072976477dbf85d00242718ad69a60/fava/ext/portfolio_list/templates/portfolio_list.html#L9

fava/ext/portfolio_list/__init__.py
```diff
@@ -18,15 +18,11 @@ class PortfolioList(FavaExtensionBase):  # pragma: no cover
     """Sample Extension Report that just prints out an Portfolio List.
     """
 
-    def __init__(self, ledger):
-        super().__init__(ledger)
+    # pylint: disable=eval-used
+    def __init__(self, ledger, extension_config):
+        super().__init__(ledger, extension_config)
+        self.config = eval(extension_config, {}, {})
```
yagebu: Don't use `eval` and rather `ast.literal_eval`. Also I think we should do this for the extension

fava/ext/portfolio_list/templates/portfolio_list.html
```diff
@@ -0,0 +1,13 @@
+{% extends "_layout.html" %}
```
yagebu: This should be handled by the extension system, the extension should only provide the contents for the content block

fava/help/extensions.md
```diff
@@ -29,3 +51,15 @@ Called after metadata (`key: value`) has been added to an `entry`.
 ### `after_insert_entry(entry)`
 
 Called after an `entry` has been inserted.
+
+---
+
+### `report_template()`
```
yagebu: Remove, i.e., update with documentation on where the template should be placed.

fava/ext/portfolio_list/__init__.py
```diff
@@ -0,0 +1,154 @@
...
+    """Sample Extension Report that just prints out an Portfolio List.
+    """
+
+    # pylint: disable=eval-used
+    def __init__(self, ledger, extension_config):
+        super().__init__(ledger, extension_config)
+        self.config = eval(extension_config, {}, {})
+        self.report_title = "Portfolio List"
+
+    def report_template(self):
```
yagebu: This code should be moved to the extensions core module

fava/ext/portfolio_list/templates/portfolio_list.html
```diff
@@ -0,0 +1,13 @@
...
+{% import "_query_table.html" as querytable with context %}
+{% set active_page = 'PortfolioList' %}
```
yagebu: Likewise, the extension should really only provide what's in content block

fava/ext/portfolio_list/templates/portfolio_list.html
yagebu: Actually, I want to make a similar change for the reports that we ship with, so you can wait for that change

seltzered:
Thanks again for the review! Will try to come back to this late this week/weekend.

fava/ext/portfolio_list/__init__.py
seltzered: Done - https://github.com/beancount/fava/pull/872/commits/32c4557ad79b5820e2a32bd8365025516f8f76bb

> Instead of the `active_ext` hack, I'd just pass the active extension to the template.
Done in https://github.com/beancount/fava/pull/872/commits/da80a9241f1080a1d1fd107f6ed156afd23a70df#diff-6d90976a6c500672baef2abdbc814f8f

> Please remove the `extensions-dir` option, if you want to use extensions from different paths, just use a python module...
Done, just need to squash.

fava/ext/portfolio_list/templates/portfolio_list.html
seltzered: Done, created a loader function in order to handle template includes of extensions  - https://github.com/beancount/fava/pull/872/commits/da80a9241f1080a1d1fd107f6ed156afd23a70df#diff-6d90976a6c500672baef2abdbc814f8f

fava/help/extensions.md
seltzered: Done, documentation updated at https://github.com/beancount/fava/pull/872/commits/1b282eadaa5778e92d9caef27003b9ed443e0112#diff-a254dd640147f6c3f0aa361a8c736821

fava/ext/portfolio_list/__init__.py
seltzered: Done, report_title attr is still there but it's now optional.

fava/ext/portfolio_list/templates/portfolio_list.html
seltzered: Done, though i still import query_table for the sample extension

seltzered:
Note: I have not been able to figure out why lint (black) is complaining about application.py.
Also, I undid your removal of 'with context' in _aside / _layout in https://github.com/beancount/fava/pull/872/commits/6943530fb65fec5888a6c78073b3b47edf0d2673 because I reference ledger in _globals.html ( https://github.com/beancount/fava/pull/872/commits/e1b147fbc667d1f7dad2865210969ae7ad246c98#diff-a300e908d18c44b8971354a702f02d17 ) - will see if i can think of an alternative.

yagebu:
> Note: I have not been able to figure out why lint (black) is complaining about application.py.

Run `make lint`

seltzered:
> Run `make lint`
D'oh, thanks again! Fixed and squashed.

a-bali:
So how is this supposed to work? I have checked out this branch and added the following to my beancount file. No error shows up but the portfolio report is not available either.

`1800-01-01 custom "fava-extension" "fava.ext.portfolio_list" "[('account_name_pattern', '.*')]"`

seltzered:
@a-bali can you try copying the `portfolio_list` plugin directory (located in fava's ext/) to the same directory as your beancount file? The other thing you can do is add the extensions directory to your $PYTHONPATH (i.e. adding the python module).

(I did have an `extensions-dir` option earlier in the PR but removed it per request. Perhaps I should add more to the extension documentation on adding modules, since I originally didn't understand Jakob's advice).

seltzered:
> @a-bali can you try copying the `portfolio_list` plugin directory (located in fava's ext/) to the same directory as your beancount file? The other thing you can do is add the extensions directory to your $PYTHONPATH (i.e. adding the python module).
> 
> (I did have an `extensions-dir` option earlier in the PR but removed it per request. Perhaps I should add more to the extension documentation on adding modules, since I originally didn't understand Jakob's advice).

EDIT: nevermind, noticed you're writing it as `fava.ext.portfolio_list` and no errors getting reported, looking into it.

seltzered:
@a-bali thanks for testing & finding a bug. The issue was that the template path was getting interpreted as `portfolio_list/templates/fava.ext.portfolio_list.html`. Changed convention for extension templates to use qualname in the latest commit.

You can now just run the example directly too: 
`$fava /myPathToFava/tests/data/extension-report-example.beancount`

a-bali:
Thanks, I confirm that it works now.

yagebu:
@seltzered: Is this ready?

seltzered:
Yes, haven’t had any more changes planned for now.

yagebu:
Will continue and merge this in #924

Issue #867, opened 2019-01-19
Show net profit on income statement


zacchiro:
In the Net Profit view currently there is the timeline plus two columns, for Income and Expenses. It would be nice to show somewhere the cumulative total of Expenses - Income (i.e., the cumulative net profit) matching the current view and filters.

Right now one can easily obtain the net profit only for each bar in the chart (by hovering with the mouse cursor) or by mentally doing the difference between the topmost entries in the Expenses and Income chart, which is kind of annoying.

yagebu:
I don't understand what you mean by "Net Profit view". If you're referring to the net profit chart, that one already does exactly what you seem to request: show the sum of income and expenses (per interval).

zacchiro:
I'm talking about the Net Profit "tab" of the Income Statement view. It shows the profit chart and the sum of income and expenses *per interval* indeed. But it does not show the sum of income and expenses over the entire time period being observed. Or, alternatively, the grand total of all the bars you're seeing in the chart.

So, say, if you've filtered on a specific time period, e.g., 2015-2019, and you've selected yearly charting, *for each year* you can easily find your net profit, but you don't have an easy way to find out your net profit over the overall period. To do so you've to either add up the values of each bar, or make the sum of the top cells in the Expenses and Income tables.

Hope this clarifies,

zacchiro:
So, if you're against this, what is then the recommended Beancount way to see "how well" something that you've tracked with accounting went in terms of expenses+income?

What I currently do is:
- set up my filters using the top bar (time period, tags, metadata, etc.)
- go to the net profit view
- add up (in my head or with an external calculator) the totals of the Income and Expenses tables just below the net profit chart

Of course I can do a BQL query in the Query view, but that feels weird, given I'm using a UI. If the proper way is to use BQL, then I could as well just run bean-query on the command line.

Am I missing something here?

zacchiro:
Income statement in the "net profit" tab, as per initial suggestion, or as you saying to put it in a different tab of that view?

yagebu:
There aren't any "tabs" on the income statement - it's just one page, you can only switch between the charts.

zacchiro:
OK, so you want it below the charts, in the common part. I'll give it a try.

yagebu:
> OK, so you want it below the charts, in the common part. I'll give it a try.

Yes, it would be unnecessarily complicated to implement otherwise.

I don't where net profit will usually be positioned on an income statement, but if there's some sensible convention, we should follow it.

PR #873, opened 2019-01-26
income statement: add net profit total (income + expenses)

zacchiro:
This PR still needs some work, but it is a first stab at #867.

Where to place the net profit total isn't entirely clear to me. In income
statements it's usually at the bottom (see, e.g.,
https://en.wikipedia.org/wiki/Income_statement#Usefulness_and_limitations_of_income_statement
), but I fear that placing it at the bottom would reduce the visibility in
Fava, given the long list of recursive unfolding.

It's now at the top, but it's not meant to remain there...

It's also a column right now, while it should probably be full-width, to avoid
ending up on the same row of either income/expenses. What would be the most
appropriate way to achieve that?

Comments on the overall approach welcome.


zacchiro:
After some UI fiddling, I think that actually showing the new net profit column *last* is not that bad. On wide screens it will actually fit as a 3rd column, which makes sense. On normal-size screens it will float to a new row as the sole column, which isn't too bad either.

I've updated the PR accordingly.

The only glitch I can see is that, when the 3rd column actually flows to a new row, the "Net Profit" title centers on the whole page rather than on top of the new Net Profit table. It seems to be a general issue with the CSS "column" class, though, rather than specific to this PR.

fava/templates/_tree_table.html
```diff
@@ -178,3 +178,41 @@
     </li>
 </ol>
 {% endmacro %}
+
+{% macro net_profit(income, expenses) %}
```
yagebu: This seems like a lot of duplicated code - I think it would be better to reuse the normal tree-table for this (you can create a tree containing a single "Net Profit" account for that)

yagebu:
> The only glitch I can see is that, when the 3rd column actually flows to a new row, the "Net Profit" title centers on the whole page rather than on top of the new Net Profit table. It seems to be a general issue with the CSS "column" class, though, rather than specific to this PR.

That's a rather unsolvable problem, since the account tree has a max width but the column class will grow (and we can't center the heading with respect to a sibling element)

I guess we could remove the heading altogether if we emphasise the base accounts a bit more.

fava/templates/_tree_table.html
zacchiro: Thanks for the tip about tree/treenode, I've given it a try in the new version of the PR. There is no template code duplication remaining, but I had to add an extra method to the ledger object, in order to have access to the tree constructor (and given I had to do that, I've put there the logic of computing the net profit).

zacchiro:
> I guess we could remove the heading altogether if we emphasise the base accounts a bit more.

I've dropped the heading in the most recent version of the PR. I agree it's not bad.

yagebu:
Moved the function for this to `fava.core.tree` and merged.

PR #874, opened 2019-02-01
ingest: Fix duplicate detection with recent Beancount versions

jonleighton:
https://bitbucket.org/blais/beancount/commits/b00e95b9c61ceac924826cd0ac15f64f9d9cd8f0
changed Beancount such that the extract.extract_from_file function no
longer calls extract.find_duplicate_entries. This means that when using
Fava with a recent vesion of Beancount, no duplicate detection would be
performed.

This commit fixes the behaviour and adds some assertions to ensure it is
tested.

### DIFF
--- a/fava/core/ingest.py
+++ b/fava/core/ingest.py
@@ -119,6 +119,10 @@ class IngestModule(FavaModule):
 
         if isinstance(new_entries, tuple):
             new_entries, _ = new_entries
+        else:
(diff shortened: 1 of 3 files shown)


yagebu:
Thanks! (sorry for the delay)

jonleighton:
No probs, thanks for merging :)

PR #879, opened 2019-03-01
Add 'collapse-below-level' option

mbudde:
Setting the 'collapse-below-level' option will collapse accounts at or
below this depth in the account tree. The 'fava-collapse-account'
metadata on an account will override this setting.

### DIFF
--- a/fava/help/features.md
+++ b/fava/help/features.md
@@ -56,6 +56,7 @@ number or a deep hierarchy of accounts, Fava offers the following options:
 -   `show-closed-accounts`
 -   `show-accounts-with-zero-balance`
 -   `show-accounts-with-zero-transactions`
+-   `collapse-below-level`
 
 Additionally, accounts that have the metadata `fava-collapse-account` set to
 `TRUE` on their Open directive will be collapsed in the account trees.
--- a/fava/template_filters.py
+++ b/fava/template_filters.py
@@ -182,3 +182,13 @@ def format_errormsg(message):
         .replace("for '", "for ")
         .replace("': ", ": ")
     )
+
+def collapse_account_at_level(account_name, level):
+    """Return true if account should be collapsed at the given tree depth."""
+    collapse_account = g.ledger.accounts[account_name].meta.get('fava-collapse-account')
+    if collapse_account is not None:
+        return collapse_account
+    collapse_level = g.ledger.fava_options['collapse-below-level']
+    if collapse_level is not None:
+        return level >= collapse_level
+    return False
--- a/fava/templates/_tree_table.html
+++ b/fava/templates/_tree_table.html
@@ -35,7 +35,7 @@ Hold Ctrl or Cmd while clicking to expand one level.') %}
   {% set balance_children = account.balance_children|cost_or_value(ledger.end_date) %}
   {% set cost = account.balance|cost if g.conversion == 'at_value' else {} %}
   {% set cost_children = account.balance_children|cost if g.conversion == 'at_value' else {} %}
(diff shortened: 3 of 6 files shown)


fava/template_filters.py
```diff
@@ -182,3 +182,13 @@ def format_errormsg(message):
         .replace("for '", "for ")
         .replace("': ", ": ")
     )
+
+def collapse_account_at_level(account_name, level):
```
yagebu: Could you add a test for this function

fava/help/features.md
```diff
@@ -56,9 +56,13 @@ number or a deep hierarchy of accounts, Fava offers the following options:
 -   `show-closed-accounts`
 -   `show-accounts-with-zero-balance`
 -   `show-accounts-with-zero-transactions`
+-   `collapse-below-level`
```
yagebu: The option should also be documented in `options.md`

fava/help/features.md
```diff
@@ -56,9 +56,13 @@ number or a deep hierarchy of accounts, Fava offers the following options:
...
 
-Additionally, accounts that have the metadata `fava-collapse-account` set to
-`TRUE` on their Open directive will be collapsed in the account trees.
+The last option will collapse all accounts at or below a given level in the
```
yagebu: Don't think we need to be so verbose here, this detailed explanation rather belongs in `options.md`

fava/templates/_tree_table.html
```diff
@@ -35,7 +35,8 @@
   {% set balance_children = account.balance_children|cost_or_value(ledger.end_date) %}
   {% set cost = account.balance|cost if g.conversion == 'at_value' else {} %}
   {% set cost_children = account.balance_children|cost if g.conversion == 'at_value' else {} %}
-  <li{{ ' class=toggled' if ledger.accounts[account.name].meta.get('fava-collapse-account') else '' }}>
+  {% set collapse = account.name|collapse_account_at_level(loop.depth0) %}
```
yagebu: Since this is used only once, maybe it's better to just use it in-place

mbudde: I'm not sure how would go about testing the feature in that case, so I have kept it as a filter.

fava/help/features.md
mbudde: I have moved the explanation to `options.md`.

fava/templates/_tree_table.html
yagebu: I meant to not declare a variable here and use the value directly below

mbudde: Ahh, right :smile:

yagebu:
What seems a bit odd to me about the current implementation is that it uses the current level of the account in the tree and not the level of the account (as can be deduced from its name). Since this change doesn't currently apply to the tree tables on the account pages, that doesn't really make a difference since all the tree tables start at the same level of the account tree, but I think for consistency it should also apply on the account pages. I think it's somewhat preferable to have it only depend on the level of the account, or do you disagree?

mbudde:
> I think it's somewhat preferable to have it only depend on the level of the account, or do you disagree?

Hmm, I'm a bit conflicted. Personally I think I would prefer the current behavior. The reason I want to collapse the accounts in the account tree is to reduce the amount of information so it's easier to get an overview. On the account pages I don't have a problem with information overload so there I would prefer to have the accounts (partially) expanded when viewing a subtree. But I guess I could live it either way, so I'll update the PR if you think having it depend only on the account is the way to go.

I agree that it should be enabled on the account pages for consistency.

yagebu:
Sorry for not getting back any earlier, was a bit unsure of how to proceed with this - let's just stick with the behaviour you implemented for now then. I'll apply it on the account pages and merge.

PR #884, opened 2019-03-09
Text files always open in UTF-8

### LINKED ISSUE #883: UnicodeDecodeError: 'charmap' codec can't decode byte when adding new transaction
I'm using 
* fava 1.10
* beancount 2.2.1
* Python 3.6.5 in virtualenv
* Windows 10
* Hungarian locale

When I try to enter a new transaction on the web interface I can type in all details, but even though there is no special characters in the transaction, clicking on `Save` causes an `INTERNAL SERVER ERROR` and the following trace:

```
D:\Documents\beancount>c:\Users\belidzs\PycharmProjects\beancount_virtualenv\Scripts\fava --debug fokonyv.beancount
 * Serving Flask app "fava.application" (lazy loading)
 * Environment: production
   WARNING: Do not use the development server in a production environment.
   Use a production WSGI server instead.
 * Debug mode: on
 * Restarting with stat
 * Debugger is active!
 * Debugger PIN: 546-083-666
 * Running on http://localhost:5000/ (Press CTRL+C to quit)
127.0.0.1 - - [09/Mar/2019 10:48:22] "PUT /f%C5%91k%C3%B6nyv/api/add-entries/ HTTP/1.1" 500 -
  File "c:\users\belidzs\pycharmprojects\beancount_virtualenv\lib\site-packages\flask\app.py", line 2309, in __call__
    return self.wsgi_app(environ, start_response)
  File "c:\users\belidzs\pycharmprojects\beancount_virtualenv\lib\site-packages\flask\app.py", line 2295, in wsgi_app
    response = self.handle_exception(e)
  File "c:\users\belidzs\pycharmprojects\beancount_virtualenv\lib\site-packages\flask\app.py", line 1741, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "c:\users\belidzs\pycharmprojects\beancount_virtualenv\lib\site-packages\flask\_compat.py", line 35, in reraise
    raise value
  File "c:\users\belidzs\pycharmprojects\beancount_virtualenv\lib\site-packages\flask\app.py", line 2292, in wsgi_app
    response = self.full_dispatch_request()
  File "c:\users\belidzs\pycharmprojects\beancount_virtualenv\lib\site-packages\flask\app.py", line 1815, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "c:\users\belidzs\pycharmprojects\beancount_virtualenv\lib\site-packages\flask\app.py", line 1718, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "c:\users\belidzs\pycharmprojects\beancount_virtualenv\lib\site-packages\flask\_compat.py", line 35, in reraise
    raise value
  File "c:\users\belidzs\pycharmprojects\beancount_virtualenv\lib\site-packages\flask\app.py", line 1813, in full_dispatch_request
    rv = self.dispatch_request()
  File "c:\users\belidzs\pycharmprojects\beancount_virtualenv\lib\site-packages\flask\app.py", line 1799, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "c:\users\belidzs\pycharmprojects\beancount_virtualenv\lib\site-packages\fava\json_api.py", line 42, in _wrapper
    return func(request_data)
  File "c:\users\belidzs\pycharmprojects\beancount_virtualenv\lib\site-packages\fava\json_api.py", line 26, in _wrapper
    json_data = func(*args, **kwargs)
  File "c:\users\belidzs\pycharmprojects\beancount_virtualenv\lib\site-packages\fava\json_api.py", line 179, in add_entries
    g.ledger.file.insert_entries(entries)
  File "c:\users\belidzs\pycharmprojects\beancount_virtualenv\lib\site-packages\fava\core\file.py", line 114, in insert_entries
    insert_entry(entry, self.list_sources(), self.ledger.fava_options)
  File "c:\users\belidzs\pycharmprojects\beancount_virtualenv\lib\site-packages\fava\core\file.py", line 286, in insert_entry
    accounts, entry.date, insert_options, filenames
  File "c:\users\belidzs\pycharmprojects\beancount_virtualenv\lib\site-packages\fava\core\file.py", line 328, in find_insert_position
    return (filenames[0], len(open(filenames[0]).readlines()) + 1)
  File "C:\Program Files\Python36\lib\encodings\cp1250.py", line 23, in decode
    return codecs.charmap_decode(input,self.errors,decoding_table)[0]
UnicodeDecodeError: 'charmap' codec can't decode byte 0x81 in position 6806: character maps to <undefined>
```

I've tried to change console locale to utf-8 (`chcp 65001`) and 1250 (`chcp 1250`) from 852 but I always get the same results.

The transaction:
image

belidzs:
If encoding isn't explicitly specified when opening a text file, Python defaults to locale.getpreferredencoding()

For example, on a Windows 10 with Hungarian locale Python would assume that a text file is encoded in code page 852 instead of the standard UTF-8.

Since detecting the encoding of a text file is very difficult we should at least make the decoder constant so there is no need to guess which file encoding was expected by fava.

This fixes #883 on my setup.

### DIFF
--- a/contrib/scripts.py
+++ b/contrib/scripts.py
@@ -46,7 +46,7 @@ def generate_bql_grammar_json():
         os.path.dirname(__file__),
         "../fava/static/javascript/codemirror/bql-grammar.json",
     )
-    with open(path, "w") as json_file:
+    with open(path, "w", encoding="utf-8") as json_file:
         json.dump(data, json_file)
 
 
--- a/fava/core/file.py
+++ b/fava/core/file.py
@@ -177,7 +177,7 @@ def leading_space(line):
 def insert_metadata_in_file(filename, lineno, key, value):
     """Inserts the specified metadata in the file below lineno, taking into
     account the whitespace in front of the line that lineno."""
-    with open(filename, "r") as file:
+    with open(filename, "r", encoding="utf-8") as file:
         contents = file.readlines()
 
     # use the whitespace of the following line, else use double the whitespace
@@ -185,7 +185,7 @@ def insert_metadata_in_file(filename, lineno, key, value):
 
     contents.insert(lineno + 1, '{}{}: "{}"\n'.format(indention, key, value))
 
-    with open(filename, "w") as file:
+    with open(filename, "w", encoding="utf-8") as file:
         contents = "".join(contents)
         file.write(contents)
 
@@ -220,7 +220,7 @@ def get_entry_slice(entry):
             source files.
 
     """
-    with open(entry.meta["filename"], mode="r") as file:
+    with open(entry.meta["filename"], mode="r", encoding="utf-8") as file:
         lines = file.readlines()
 
     entry_lines = find_entry_lines(lines, entry.meta["lineno"] - 1)
@@ -247,7 +247,7 @@ def save_entry_slice(entry, source_slice, sha256sum):
 
     """
 
-    with open(entry.meta["filename"], "r") as file:
+    with open(entry.meta["filename"], "r", encoding="utf-8") as file:
         lines = file.readlines()
 
     first_entry_line = entry.meta["lineno"] - 1
@@ -262,7 +262,7 @@ def save_entry_slice(entry, source_slice, sha256sum):
         + [source_slice + "\n"]
         + lines[first_entry_line + len(entry_lines) :]
     )
-    with open(entry.meta["filename"], "w") as file:
+    with open(entry.meta["filename"], "w", encoding="utf-8") as file:
         file.writelines(lines)
 
     return sha256(codecs.encode(source_slice)).hexdigest()
@@ -287,12 +287,12 @@ def insert_entry(entry, filenames, fava_options):
     )
     content = _format_entry(entry, fava_options) + "\n"
 
-    with open(filename, "r") as file:
+    with open(filename, "r", encoding="utf-8") as file:
         contents = file.readlines()
 
     contents.insert(lineno, content)
 
-    with open(filename, "w") as file:
+    with open(filename, "w", encoding="utf-8") as file:
         file.writelines(contents)
 
     for index, option in enumerate(insert_options):
@@ -325,4 +325,4 @@ def find_insert_position(accounts, date, insert_options, filenames):
             if insert_option.re.match(account):
                 return (insert_option.filename, insert_option.lineno - 1)
 
-    return (filenames[0], len(open(filenames[0]).readlines()) + 1)
+    return (filenames[0], len(open(filenames[0], encoding="utf-8").readlines()) + 1)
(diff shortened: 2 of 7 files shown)


belidzs:
I have ran the new tests on a vanilla appveyor instance and it failed as it should. It seems that the character encoding issue is global to the Windows platform and not specific to any culture. 

While I have included a new beancount file with utf-8 encoded contents to trigger this exception I recommend merging its content with `long-example.beancount` so every test case will have to crunch through a transaction with utf-8 encoded contents too. 

I haven't done this myself as the header of that file said `THIS FILE HAS BEEN AUTO-GENERATED` which mean it might be overwritten later.

After this has been done the new test case `test_utf8_content()` is no longer needed.

yagebu:
Thanks for discovering the bug and providing the fix!

> While I have included a new beancount file with utf-8 encoded contents to trigger this exception I recommend merging its content to long-example.beancount so in the future every test case will have to crunch through a transaction with utf-8 encoded contents as well. After this has been done the new test case test_utf8_content() is no longer needed.
>
>I haven't done this myself because the header of that file states THIS FILE HAS BEEN AUTO-GENERATED so I'm not really sure that this change would persist if someone would auto-generate that file again.
Merge state



Sounds good. Just remove the auto-generated line at the top, this is left since the file was initially generated by bean-example.

belidzs:
You're very welcome, I'm happy to provide a fix for such a great project like fava.

In the latest commit I have merged the new test data with the main ledger file and removed the newly created test case `test_utf8_content()`. 

After running the tests I had to fix some of them as they also used `open()` without the `encoding` attribute which fails on Windows platform except if `locale.getpreferredencoding()` returns `utf-8` somehow.

Before merging I recommend making sure that beancount files indeed assumed to be UTF-8 encoded so no previously working ledger files will stop working after this change.

belidzs:
I have done some research and it seems that beancount already has [chardet](https://github.com/chardet/chardet) listed as a [dependency](https://bitbucket.org/blais/beancount/src/561894834ea5e9dacc0f1c355248324c88a786bc/setup.py?at=default&fileviewer=file-view-default) so we could use that to detect encoding instead of hardcoding UTF-8. 

Funnily enough its only known limitation is that it cannot detect ISO-8859-2 and CP1250 character sets which are exactly the two code pages used to encode Hungarian characters besides UTF-8. However since UTF-8 is supported this shouldn't be a problem for the average Joe.

Anyway, if you're interested I can modify this PR to take advantage of that library instead of hardcoding UTF-8.

yagebu:
Beancount has the `chardet` dependency for the import functionality only and only supports utf-8 input files. So forcing utf-8 is preferable IMHO.

PR #894, opened 2019-03-31
added biweekly budget interval

cbrunet:
There are many cases where my expenses and incomes are one every two weeks: salary, mortgage, etc. Adding a biweekly interval for budget will help me entering those numbers in a more meaningful way.


yagebu:
Thanks for the PR. The code looks good :)

I'm not quite sure this is worth the complexity though (in particular I don't like that it introduces a split between the intervals available in time filters etc. and the ones that can be used for budgets). Are you aware that Beancount allows you to do simple arithmetic? You could achieve a biweekly budget as follows:

```
2013-01-01 custom "budget" Expenses:Books        "weekly"       20.00/2 EUR
```

cbrunet:
You're right. I didn't realized that. Thank you!

Issue #895, opened 2019-04-01
Loading entry context failed


kiney:
In the current HEAD when I click on a transaction in journal view it fails with
`Loading entry context failed.`
in the gui.
The version (1.10) i installed with pip works just fine.
I think it was introduced with commit c5ad85f but I still don't understand the exact reason...

exact traceback from console:
```
Exception on /vtth-kassenbuch/_context/ [GET]
Traceback (most recent call last):
  File "/home/kiney/tmp/fava/env/lib/python3.6/site-packages/flask/app.py", line 2292, in wsgi_app
    response = self.full_dispatch_request()
  File "/home/kiney/tmp/fava/env/lib/python3.6/site-packages/flask/app.py", line 1815, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/home/kiney/tmp/fava/env/lib/python3.6/site-packages/flask/app.py", line 1718, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/home/kiney/tmp/fava/env/lib/python3.6/site-packages/flask/_compat.py", line 35, in reraise
    raise value
  File "/home/kiney/tmp/fava/env/lib/python3.6/site-packages/flask/app.py", line 1813, in full_dispatch_request
    rv = self.dispatch_request()
  File "/home/kiney/tmp/fava/env/lib/python3.6/site-packages/flask/app.py", line 1799, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/home/kiney/tmp/fava/fava/application.py", line 308, in report
    return render_template("_layout.html", active_page=report_name)
  File "/home/kiney/tmp/fava/env/lib/python3.6/site-packages/flask/templating.py", line 135, in render_template
    context, ctx.app)
  File "/home/kiney/tmp/fava/env/lib/python3.6/site-packages/flask/templating.py", line 117, in _render
    rv = template.render(context)
  File "/home/kiney/tmp/fava/env/lib/python3.6/site-packages/jinja2/asyncsupport.py", line 76, in render
    return original_render(self, *args, **kwargs)
  File "/home/kiney/tmp/fava/env/lib/python3.6/site-packages/jinja2/environment.py", line 1008, in render
    return self.environment.handle_exception(exc_info, True)
  File "/home/kiney/tmp/fava/env/lib/python3.6/site-packages/jinja2/environment.py", line 780, in handle_exception
    reraise(exc_type, exc_value, tb)
[...]

PR #896, opened 2019-04-02
dont set page_title for popups #895

kiney:
Hi,
i tried debugging #895. It seems that the template tries to set a page_title when opening the transaction view (_context.html) which is just a popup and has no entry in _globals.html.
I guess it makes no sense giving it a title in _globals, so I changed the condition for setting page_title.

I'm not familiar with the codebase so please review if this is a proper fix.

Also the layout is a bit broken when the popup is active: the close button does not work (but clicking the backgroud works fine) and the search bars move from top right to top left. I guess this is not directly related?
Should i open another bug for that?

### DIFF
--- a/fava/templates/_layout.html
+++ b/fava/templates/_layout.html
@@ -1,6 +1,8 @@
 {% import "_charts.html" as charts with context %}
 {% import "_globals.html" as globals %}
-{% set page_title = globals.all_pages[active_page].0 if not page_title else page_title %}
+{% if active_page in globals.all_pages and not page_title %}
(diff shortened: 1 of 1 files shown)


yagebu:
We don't want to render the layout at all for the popups, so this isn't the right fix ;)

Fixed in https://github.com/beancount/fava/commit/9f6bfb4995b0601b17249851c31a1720e26afe9f instead

kiney:
Ok, makes sense. Thanks for the feedback, thats why I asked ;-)

PR #907, opened 2019-04-21
Make URLs in tag values into clickable links

tbm:


tbm:
The idea is to make URLs in tag values into clickable links.  I refer to internal web sites for receipts in some cases (rather than filenames) so it would be nice of I could click on them.

BTW, I added `class="url"` in a second iteration of the patch and for some reason it doesn't show up in my browser.  But maybe that's a cache issue. or does something remove that?

fava/static/css/journal-table.css
```diff
@@ -211,6 +211,11 @@
   margin-left: 1em;
 }
 
+.journal .url{
```
yagebu: merge this with .filename

PR #912, opened 2019-04-30
Consider leap years for FYE 02-28

### LINKED ISSUE #911: fiscal-year-end 02-28 and leap year: ValueError: day is out of range for month
```
2000-01-01 custom "fava-option" "fiscal-year-end" "02-28"
```

can lead to:

```
  File "/home/tbm/.local/lib/python3.7/site-packages/fava/util/date.py", line 262, in parse_date
    return get_fiscal_period(year, fye)
  File "/home/tbm/.local/lib/python3.7/site-packages/fava/util/date.py", line 314, in get_fiscal_period
    return start_date, start_date.replace(year=start_date.year + 1)
ValueError: day is out of range for month
```

start date is set to 2016-02-29 (instead of 2016-03-01) but 2017-02-29 is not valid.

The fix I tried caused other problems, but I'll try to submit a different patch.

tbm:
A FYE of 02-28 would lead to the following error for certain years:

        return start_date, start_date.replace(year=start_date.year + 1)
    ValueError: day is out of range for month

This is because the start_date was incorrectly calculated as 2016-02-29
(instead of 2016-03-01) but 2017-02-29 (start_date plus a year) doesn't
exist.

Fixes #911

### DIFF
--- a/fava/util/date.py
+++ b/fava/util/date.py
@@ -307,6 +307,9 @@ def get_fiscal_period(year, fye, quarter=None):
                 )
                 + datetime.timedelta(days=1)
             ).date()
+            # Special case 02-28 because of leap years
+            if fye == "02-28":
+                start_date = start_date.replace(month=3, day=1)
         except ValueError:
             return None, None
 
--- a/tests/test_util_date.py
+++ b/tests/test_util_date.py
@@ -249,6 +249,9 @@ def test_month_offset(date_input, offset, expected):
         # 5th Apr - UK [FYE=04-05]
         (2018, None, "04-05", "2017-04-06", "2018-04-06"),
         (2018, 1, "04-05", "None", "None"),
+        # 28th February - consider leap years [FYE=02-28]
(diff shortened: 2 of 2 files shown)


fava/util/date.py
```diff
@@ -307,6 +308,8 @@ def get_fiscal_period(year, fye, quarter=None):
                 )
                 + datetime.timedelta(days=1)
             ).date()
+            if fye == "02-28" and calendar.isleap(year - 1):
```
yagebu: This is a bit confusing I find since independently of whether the year is a leap year, we set the date to x-03-01. So let's make that more explicit (and simplify the code a bit) by only making the first check and then setting the date directly

tbm: That's what I did originally: add a month to the date and then set day to 1.

Unfortunately, that doesn't work because it breaks FYE like 05-06.

tbm: Oh, I see what you mean.  Sorry, I misunderstood.  Yeah, ok, I can do that

yagebu:
Thanks!

PR #914, opened 2019-05-01
Set default years to fiscal years if FYE is specified

### LINKED ISSUE #906: Show FY in time dropdown menu when fiscal-year-end is set
Fava shows all years that are mentioned in the beancount file in the `time` box.

Some organizations and countries don't follow the calendar year. For example, Software Freedom Conservancy's financial year starts on March 1.  In the UK, the tax year runs from April 6 to April 5.

It would be nice if fava allowed users to configure some time periods.

(I'm not sure if this should be done in fava or beancount. But afaik beancount doesn't have any mechanism for predefined time periods. Any opinion?)

tbm:
Instead of showing calendar years for `time`, let's add fiscal
years if the `fiscal-year-end` option is set.

Fixes #906

### DIFF
--- a/fava/core/attributes.py
+++ b/fava/core/attributes.py
@@ -3,8 +3,10 @@
 from beancount.core import getters
 from beancount.core.data import Transaction
 
+from fava.core.fava_options import DEFAULTS
 from fava.core.helpers import FavaModule
 from fava.util.ranking import ExponentialDecayRanker
+import fava.util.date
 
 
 class AttributesModule(FavaModule):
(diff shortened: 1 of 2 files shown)


tbm:
This doesn't work at the moment because core/attributes doesn't have a self.fava_options.

Otherwise it does what I want

@yagebu what's the most elegant way to solve this?  (Or maybe you can fix up my WIP patch)

yagebu:
@tbm: It has self.ledger.fava_options ;)

tbm:
> @tbm: It has self.ledger.fava_options ;)

Ok, that works.

tbm:
I can add a test but it will have to wait about a week.

I was actually going to add a test but couldn't quite figure out why because I'll have to define some beancount transactions to get the years from.  I'll have to look into that after some travel I have coming up.

yagebu:
@tbm: ping

tbm:
sorry, I'm behind with things.  I'll try to do it next week

tbm:
sorry, I'm still behind with things, but trying to work on my todo list...

yagebu:
Use of `get_fiscal_period` is rather inefficient. Reimplemented with tests in #1079

PR #924, opened 2019-05-11
Extension reports

### LINKED ISSUE #759: extensions: support user reports

### LINKED ISSUE #878: extensions: support an options map?
While thinking of an example for extension reports https://github.com/beancount/fava/issues/759 , I'm starting to wonder if extensions themselves should allow for an options_map similar to how beancount's own `plugin` statements allow. Doing this likely means changing the `extensions` fava-option from being in `LIST_OPS` to a non-listed `extension` perhaps marked under a new list called `MAP_OPTS` or `OPT_OPTS`.

yagebu:
Continues #872 

Closes #878 
Closes #759


yagebu:
Thanks @seltzered!

PR #930, opened 2019-05-15
fix #918: add collapse-patterns to collapse based on regex list

### LINKED ISSUE #918: fava-collapse-account and "unreal" accounts
I'd love being able to use the collapse option on "unreal" parent accounts as my account hierarchy is quite deep. It would have to be a fava specific option since the idea is to not have to open an account in beancount. As an example:

```
1901-01-01 custom "fava-option" "collapse-account-below-level" "Assets:Banks" "2"
```

This would also take care of being able to collapse selected root accounts (for instance, I want "Equity" collapsed fully always, but not Assets or Liabilities).

Also see:
> The option you want is "fava-collapse-account" as metadata on the account.

That's a nice option! But it only works if the parent account you want to collapse is a 'real' account. In my case usually it doesn't, so I can't attach metadata.
I don't want to have to create a 'real' parent account, to enforce amounts can only be posted to sub-accounts, not to the parent account itself.

_Originally posted by @vloris in https://github.com/beancount/fava/issues/876#issuecomment-466007149_

redstreet:
#918: add 'collapse-patterns' that collapses accounts based on regex list. Removes fava-collapse-account option.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -5,9 +5,9 @@ Unreleased
 ----------
 
 The import page of Fava has been reworked - it now supports moving files to the
-documents folder and the import process should be a bit more interactive.
-This release also contains various fixes and a new `collapse-patterns`
-option to collapse accounts in account trees.
+documents folder and the import process should be a bit more interactive. This release
+also contains various fixes and a new `collapse-pattern` option to collapse accounts in
+account trees based on regular expressions.
 
 v1.10 (2019-01-31)
 ------------------
--- a/CHANGES
+++ b/CHANGES
@@ -6,14 +6,14 @@ Unreleased
 
 The import page of Fava has been reworked - it now supports moving files to the
 documents folder and the import process should be a bit more interactive.
-This release also contains various fixes and a new `collapse-below-level`
+This release also contains various fixes and a new `collapse-patterns`
 option to collapse accounts in account trees.
 
 v1.10 (2019-01-31)
 ------------------
 
 This release contains mostly smaller changes and fixes. In particular, the net
-worth chart will now follow the seleected conversion.
+worth chart will now follow the selected conversion.
 
 v1.9 (2018-10-08)
 -----------------
--- a/fava/core/fava_options.py
+++ b/fava/core/fava_options.py
@@ -16,7 +16,7 @@ InsertEntryOption = namedtuple("InsertEntryOption", "date re filename lineno")
 DEFAULTS = {
     "account-journal-include-children": True,
     "currency-column": 61,
-    "collapse-patterns": [],
+    "collapse-pattern": [],
     "auto-reload": False,
     "default-file": None,
     "fiscal-year-end": "12-31",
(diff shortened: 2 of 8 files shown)


yagebu:
Let's make this a string option instead giving one regex

yagebu:
Also, I think this should also replace `collapse-below-level` - we can add a note to the documentation giving an example how to replicate it. Otherwise the interaction will be hard to understand (or elaborate to explain).

redstreet:
> Let's make this a string option instead giving one regex

I'm not seeing what you're saying. It is a string option. The string is a space separated list of regexes.

yagebu:
One string -> one regex

redstreet:
Gotcha, similar to insert-entry. Will do.

redstreet:
> Also, I think this should also replace collapse-below-level - we can add a note to the documentation giving an example how to replicate it. Otherwise the interaction will be hard to understand (or elaborate to explain).

See my comment in #918:
Agreed that in theory it could replace collapse-below-level, but collapse-below-level is much more intuitive, and is cheaper than a regex as well, so I too would vote for retaining it.

I like the elegance of a single option that covers it all. However, in this case, IMHO, collapse-below-level offers a much more intuitive and human-readable way to express the config.

How about I add to the doc;
collapse-below-level is a convenient shortcut, making the following options equivalent:

"fava-option" "collapse-below-level" 3
"fava-option" "collapse-patterns" "[^:]+:[^:]+:.+"

Does this take care of your concerns about the interaction being hard to explain?

yagebu:
The cost of this is really rather irrelevant, running a regex a hundred times won't impact performance.

> collapse-below-level is a convenient shortcut, making the following options equivalent:

But they're not (currently)...

Also, the regex can be shortened to `.*:.*:`, which we can give as an example - I don't think that regex warrants its own option...

Another problem with collapse-below-level is that it's not much more intuitive as it's unclear what the "level" of the account and it would not be totally obvious what happens if both options are set.

redstreet:
> Also, the regex can be shortened to .*:.*:, which we can give as an example - I don't think that regex warrants its own option...

You convinced me. I'll update the PR.

>> collapse-below-level is a convenient shortcut, making the following options equivalent:
> But they're not (currently)...

This is moot now, but out of curiosity, I'm not sure I'm seeing why those options are not equivalent. Mind explaining?

yagebu:
> Gotcha, similar to insert-entry. Will do.

Actually, I meant to only have one regex (which is as powerful as multiple regexes) - but your solution is better :)

> This is moot now, but out of curiosity, I'm not sure I'm seeing why those options are not equivalent. Mind explaining? I see that there's a 'level' argument to collapse_account_at_level(), but haven't looked to see why it's necessary or different from the account_name argument.

`collapse-below-level` would only consider the current level of the displayed tree. So on the income statement for example the two options would be equivalent, but on the, say, the `Assets` account page it would show one "account level" more.

redstreet:
Thanks for the explanation!

Cool, let me know if the PR needs anything else.

yagebu:
Merged, thanks!

PR #934, opened 2019-05-18
Autocompletion (fix #933)

### LINKED ISSUE #933: Autocompletion fails on non-transaction directives
Autocompletion throws a TypeError every time it's not used in a transaction.

How to reproduce: type `1970-01-01 open A`. It should hint `Assets` accounts, but instead it throws the following error:

    Uncaught TypeError: Cannot read property 'filter' of undefined
        at fuzzyMatch (app.js?mtime=1558219766:22610)
        at app.js?mtime=1558219766:22716
        at fetchHints (app.js?mtime=1558219766:22160)
        at run (app.js?mtime=1558219766:22173)
        at resolved (app.js?mtime=1558219766:22178)
        at fetchHints (app.js?mtime=1558219766:22158)
        at Completion.update (app.js?mtime=1558219766:21899)
        at CodeMirror.showHint (app.js?mtime=1558219766:21828)
        at Object.CodeMirror.showHint [as autocomplete] (app.js?mtime=1558219766:21802)
        at app.js?mtime=1558219766:23228

I have been taking a look at the code and the problem seems to be in [hint-beancount.js:105](https://github.com/beancount/fava/blob/master/fava/static/javascript/codemirror/hint-beancount.js#L105). There, we are passing an undefined, e.g. `completionSources["accounts"]` (?) into `fuzzyMatch`, and instead we should be passing the list of accounts or commodities.

cruizh:
Fixes #933 

Give `fuzzyMatch` the actual list to filter instead of an undefined string.

### DIFF
--- a/fava/static/javascript/codemirror/hint-beancount.js
+++ b/fava/static/javascript/codemirror/hint-beancount.js
@@ -19,14 +19,14 @@ const completionSources = {
 };
 
 const directiveCompletions = {
-  open: ["accounts", "commodities"],
+  open: ["accounts", "currencies"],
   close: ["accounts"],
-  commodity: ["commodities"],
-  balance: ["accounts", null, "commodities"],
+  commodity: ["currencies"],
+  balance: ["accounts", null, "currencies"],
   pad: ["accounts", "accounts"],
   note: ["accounts"],
   document: ["accounts"],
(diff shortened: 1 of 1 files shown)


yagebu:
Merged, thanks!

PR #938, opened 2019-06-01
Feat(chart): display the percentage in chart

### LINKED ISSUE #839: How to display the percentage of expenses in sunburst chart
I would like the sunburst chart of expenses not only show the balance of each subaccount but also show the percentage.

xwartz:
Feat(chart): display the percentage in chart

close #839


yagebu:
Thanks! Merged with some minor fixes

PR #939, opened 2019-06-02
change to show table header text-align to 'center'

c0b:
feel the header text is better to be `'center'` aligned than inherited `'left'` ? especially when you have multi-currency accounts, or when travelling in many countries, spending when >=5 different currencies

image


aumayr:
Or align them `'right'`?

yagebu:
I think right-aligned works better.

See also https://ux.stackexchange.com/questions/9833/best-ui-approach-to-align-the-column-titles-of-reports

PR #945, opened 2019-06-19
Allow specify command line flags from environ

yegle:
With this option, command line flags can be specified via environment variables like `FAVA_HOST`.


yagebu:
Needs to be documented

yegle:
Document added.

yegle:
Gentle ping :-)

yagebu:
Merged, thanks! I've moved the docs to the command line help and removed the `show_envvar` stuff because I found it rather ugly (and I think it should be rather obvious from the one example what the env vars are named)

PR #946, opened 2019-06-21
insert-entry: Fix a regression introduced by 740aed6

pmarciniak:
The bug is that it would use the oldest matching insert-entry option, instead of the most recent one (as documented by options.md).

Also improved test coverage.


yagebu:
Thanks!

PR #949, opened 2019-07-02
documents upload: retain the filename entered in the UI form when saving the file

stefanomasini:
It's important to preserve the Beancount convention that documents should start with `YYYY-MM-DD`. In fact, the UI already prepares a filename with such a prefix, but somehow it is disregarded in the web form.

Was it intended? Or did it just get lost along the way?

### DIFF
--- a/fava/json_api.py
+++ b/fava/json_api.py
@@ -205,7 +205,9 @@ def add_document():
         raise FavaAPIException("No file uploaded.")
 
     filepath = filepath_in_document_folder(
-        request.form["folder"], request.form["account"], request.form["filename"]
+        request.form["folder"],
+        request.form["account"],
+        request.form["filename"],
     )
     directory, filename = path.split(filepath)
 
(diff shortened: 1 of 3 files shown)


yagebu:
> Or did it just get lost along the way?

I guess it currently get's lost along the way - the file name should be passed as the name of the uploaded file. So this needs fixing in the frontend if it doesn't currently work, not in the backend.

stefanomasini:
Actually, the (small) problem was both in the frontend, that was missing the name in the html form field, and the backend, that didn't get the information about the filename from the aforementioned field.

By the way, my fix works. My branch is what I'm currently using. If you like the code, you can go ahead and merge it.

yagebu:
> By the way, my fix works

It can't really - for multiple documents, only one filename will be passed. As I said, the backend should work as is, this needs only fixing in the frontend...

I think in L19 and L21 of `DocumentUpload.svelte`, `filename` needs to be changed to `name`. And in L87 `value=` -> `bind:value=`

yagebu:
Fixed in ae6516c as detailed in my above comment

Issue #964, opened 2019-09-02
balance imported are not rounded.


grostim:
Hi!

I have an issue with my importers:
I am not sure if this is related to fava, beancount or my importer.

My importer parse pdf statement, looking for transaction and balance information.

Let's take a specific example.
In my pdf statement, i have the following balance value:
`154185.12`
My regexp find this value and store it in `chunk[7]`

I am then converting this string into an amount and creating a balance entry:

```
from beancount.core.number import Decimal, D

ope["CRD"] = amount.Amount(D("-" + chunk[7].replace(",", ".")), "EUR")

entries.append(
    data.Balance(
        meta,
        ope["date"] + datetime.timedelta(1),
        self.accountList[compte],
        ope["CRD"],
        None,
        None,
    )
)
return entries
```

This work perfecly well in CLI with bean-extract:

```
root@5029447231bb:/myData# bean-extract Famille.import Passif/Boursorama/EmpruntRP/
[...]

2016-03-27 balance Passif:Boursorama:EmpruntRP                     -154185.12 EUR
  source: "pdfbourso"

[...]
```
When i import this file from fava, the balance entry is properly shown as well in the preview form. (with 2 digits after the decimal dot).

But after confirming the import by clicking the "save" button, the entry added in the beancount file is not correct:

```
2016-03-27 balance Passif:Boursorama:EmpruntRP  -154185.119999999995343387126922607421875 EUR
  source: "pdfbourso"
```

Of course, because of theses values, the balance fail.

This happens only for balances. Not for the transaction, although my transaction entries are built exactly the same way by my importer.
Also, this rounding issue does not happen 100% of the time. Sometimes (maybe 20%), the entry is correct with 2 digits after decimal dot.

Any idea about what may happen ?

grostim:
Thanks. I'll give it a try tonight.

PR #967, opened 2019-09-15
Fix Serialization of Balance.amount.number field

6f:
The current serialization of a Amount field in a entry like "...... balance .... 204.2 CHF" results in a
in a json object {"amount": 202.4, "currency": CHF"). While this is technically correct most
json parsers treat json numbers as floating point numbers which can cause problems later
on (see #964).

We replace amount with a dictionary to force serialization of the amount field as a string instead
(i.e json object will be .{"amount": "202.4", "currency": CHF"}.

### DIFF
--- a/fava/serialisation.py
+++ b/fava/serialisation.py
@@ -67,6 +67,9 @@ def serialise(entry):
         del ret["links"]
         del ret["tags"]
         ret["postings"] = [serialise(pos) for pos in entry.postings]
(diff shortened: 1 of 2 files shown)


yagebu:
Thanks! Can you add a test?

6f:
> Thanks! Can you add a test?

Done.

yagebu:
Perfect - thanks again :)

PR #973, opened 2019-09-26
Perform currency conversion on the last day of an interval, not one day later

pmarciniak:
This is mostly done for consistency with beancount's net worth reporting logic.
E.g. when showing net worth for year 2018, beancount includes transactions up to 2018-12-31 and performs currency conversion using prices available on 2018-12-31. Fava would also include transactions up to 2018-12-31, but use prices from 2019-01-01, causing an inconsistent report results between beancount and fava.

Also updated the test data file so that an existing test_net_worth catches the problem.


yagebu:
Nice, thanks :)

PR #976, opened 2019-09-29
A hack to avoid dropping postings that specify total price (i.e. the …

pmarciniak:
…@@ notation) on the floor.

When we encounter a posting with total price (@@), we convert it to unit price notation (@), and write that in the file instead.

This is not ideal, as it introduces changes behind the user's back, resulting in the output file containing something else than what the user specified in the transaction entry dialog. Yet, one can argue it's still better than throwing an error.

The inability to write the total cost to the output file stems from the fact that fava uses beancount.core.data.Posting together with beancount.parser.printer.format_entry to format the final contents of beancount files. It's been an explicit design decision to not store total cost internally in Beancount (see discussion in https://bitbucket.org/blais/beancount/issues/194/use-of-syntax-was-use-of-float-numbers-in).

This commit makes fava effectively mirror Beancount logic (https://bitbucket.org/blais/beancount/src/621cec5ed38bcd128a3502a3b5c367f283deffe2/beancount/parser/grammar.py?at=default&fileviewer=file-view-default#grammar.py-806).

### DIFF
--- a/fava/serialisation.py
+++ b/fava/serialisation.py
@@ -90,11 +90,15 @@ def _serialise_posting(posting):
 def deserialise_posting(posting):
     """Parse JSON to a Beancount Posting."""
     amount = posting.get("amount")
-    price = None
+    unit_price = None
+    total_price = None
     if amount:
-        if "@" in amount:
-            amount, raw_price = amount.split("@")
-            price = A(raw_price)
+        if "@@" in amount:
+            amount, raw_total_price = amount.split("@@")
+            total_price = A(raw_total_price)
+        elif "@" in amount:
+            amount, raw_unit_price = amount.split("@")
+            unit_price = A(raw_unit_price)
         pos = position.from_string(amount)
         units = pos.units
         if re.search(r"{\s*}", amount):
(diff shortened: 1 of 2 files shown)


pmarciniak:
Actually, I just found out that the topic of @@ has been discussed at large in the past (e.g. https://bitbucket.org/blais/beancount/issues/194/use-of-syntax-was-use-of-float-numbers-in)

With this new knowledge, I conclude that my PR is not a hack, but the right thing to do given the current Beancount design choice, as it effectively mirrors Beancount logic. (https://bitbucket.org/blais/beancount/src/621cec5ed38bcd128a3502a3b5c367f283deffe2/beancount/parser/grammar.py?at=default&fileviewer=file-view-default#grammar.py-806)

I'm still not happy that the exact total price gets lost in the translation, but this is an orthogonal issue now.

fava/serialisation.py
```diff
@@ -103,7 +107,17 @@ def deserialise_posting(posting):
             cost = pos.cost
     else:
         units, cost = None, None
-    return data.Posting(posting["account"], units, cost, price, None, None)
+    if total_price is not None:
+        if units.number == 0:
+            unit_price = 0
```
yagebu: This should be an Amount - I think it's better to fail here and throw the error that the division by zero will throw since this is quite the edge case.

tests/test_serialisation.py
```diff
@@ -174,6 +175,45 @@ def test_deserialise():
         deserialise({"type": "NoEntry"})
 
 
+def test_deserialise_total_price():
```
yagebu: It would be a bit more concise to add these cases to the existing test for the deserialise_posting function. Will do this in the merge.

yagebu:
Merged, thanks again :)

PR #977, opened 2019-10-03
Fixed file naming for the move function in the import menu.

a-bali:
Fixes:

1. Do not skip adding date prefix to the filename in case the filename already contains a date but not in the beginning

1. Separate date and original filename by a dot (.) as bean-file does


yagebu:
Thanks - I'll merge the first part. The second is actually intentional - I prefer to separate by a space there (and we don't try to mirror bean-file's behaviour anyway, but we mostly keep the filename as is with spaces and other characters, that bean-file would replace by underscores).

PR #978, opened 2019-10-04
Add margin to custom links header

skrajewski:
When you add a custom link to fava, it appear at the top of sidebar, preceded by "Custom Links" header which has no proper margin. This PR adds the margin property.

My another proposal is to remove this header entirely to be consistent with the rest sections of sidebar.

Screenshot 2019-10-04 at 09 24 10

### DIFF
--- a/fava/static/css/base.css
+++ b/fava/static/css/base.css
@@ -351,11 +351,6 @@ aside {
   border-right: 1px solid var(--color-sidebar-border);
 }
 
-aside .header-custom-links {
-  margin-bottom: 0.5rem;
-  margin-left: 1rem;
-}
-
 .navigation {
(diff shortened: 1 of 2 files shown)


yagebu:
Hey, thanks for the PR :) I don't use this feature myself, which can explain why it get's out of sync with the rest of the UI when that changes.

> My another proposal is to remove this header entirely to be consistent with the rest sections of sidebar.

I think I'd slightly prefer that :+1:

skrajewski:
Great!

Should I remove the translations from `.po` files manually?

yagebu:
Thanks again :)

> Should I remove translations from .po files manually?

There's no need, there's a script to update them which I run periodically.

PR #981, opened 2019-10-06
Handle simple arithmetic expressions typed in amount fields.

pmarciniak:
This is useful when manually entering transactions from receipts that have multiple entries belonging in a single Beancount posting. One can simply write "5 * 3.95 CHF" (multiple copies of the same item) or "15.95 - 1.60 CHF" (10% discount applied in a separate line of the receipt), without having to perform simple albeit error-prone mental calculations.

### DIFF
--- a/fava/serialisation.py
+++ b/fava/serialisation.py
@@ -9,8 +9,12 @@ This is not intended to work well enough for full roundtrips yet.
 """
 
 import decimal
+import ast
 import functools
+import operator
 import re
+from io import BytesIO
+from tokenize import tokenize, untokenize, NAME, NUMBER, OP, STRING
 
 from beancount.core import data, position
 from beancount.core.amount import A, Amount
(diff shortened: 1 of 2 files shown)


fava/serialisation.py
```diff
@@ -9,8 +9,12 @@
 """
 
 import decimal
+import ast
 import functools
+import operator
 import re
+from io import StringIO
+from tokenize import generate_tokens, untokenize, NAME, NUMBER, OP, STRING
```
yagebu: I'd prefer to not depend on the undocumented generate_tokens function, could you change the implementation to use tokenize instead?

fava/serialisation.py
```diff
@@ -43,14 +47,66 @@ def extract_tags_links(string):
...
+
+def _decimalize_statement(statement):
+    """Wraps every float number in the statement string with a call to D().
+
+    Example: "3.14" -> "D(3.14)"
+    """
+    result = []
+    tokens = generate_tokens(StringIO(statement).readline)
+    for toknum, tokval, _, _, _ in tokens:
+        if toknum == NUMBER and _is_float_literal(tokval):
```
yagebu: Is the second part really needed? Can't we just assume that we can wrap any number (that is sensible input here) in a call to D()?

fava/serialisation.py
```diff
@@ -43,14 +47,66 @@ def extract_tags_links(string):
     return new_string, frozenset(tags), frozenset(links)
 
 
-def parse_number(num):
-    """Parse a number as entered in an entry form, supporting division."""
-    if not num:
+def _is_float_literal(literal):
+    return re.search(r"^[0-9.][0-9.]*$", literal)
```
yagebu: this regex looks wrong (and is probably not needed, see below)

fava/serialisation.py
yagebu: Wouldn't it be simpler overall to replace this function with one that just does a simple regular expression replace from number to D('number')?

pmarciniak: Changed to any number. However, I'd prefer to leave the discovery of numbers to Python, rather than trying to craft a custom regexp.

yagebu:
This looks good now. However I think I've now found a way how this can be done simpler and more completely (the current implementation only allows for calculations in the number of units, not the price or cost for example). We can embed the string input by the user in a mock transaction and let Beancount parse that - then we don't have to duplicate functionality from Beancount and we support the full syntax without any further effort :). I'll see if I can implement that in the coming days.

yagebu:
Implemented in #988

PR #992, opened 2019-10-28
Use font packages from @openfonts collection

xuhcc:
Fonts from `typeface` collection contain only `latin` subset of characters. I replaced Fira Sans and Source Code Pro font packages with full versions from `@openfonts` collection (which is a [fork](https://github.com/bedlaj/openfonts) of `typeface` project).

**Before**:

Screenshot from 2019-10-28 17-29-16

**After**:

Screenshot from 2019-10-28 17-29-20

---

I didn't touch Fira Mono and Source Serif Pro packages because these fonts are seemingly used only to display English text (BQL queries and help pages).

### DIFF
--- a/fava/static/css/fonts.css
+++ b/fava/static/css/fonts.css
@@ -3,14 +3,14 @@
   font-family: "Source Code Pro";
   font-style: normal;
   font-weight: normal;
-  src: url("./source-code-pro-latin-400.woff2") format("woff2");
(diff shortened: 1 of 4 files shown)


yagebu:
Hey, thanks for the PR! Could you switch over the other fonts to use `@openfonts` packages as well for consistency?

xuhcc:
Done. 

The Source Serif Pro package from `@openfonts` collection comes only in `latin` variant.

PR #994, opened 2019-11-03
Fix text overflow in balance table

xuhcc:
Fixed text overflow in balance table when amount has many decimal places.

**Before**:

Screen Shot 2019-11-03 at 15 25 06

**After**:

Screen Shot 2019-11-03 at 15 24 50

### DIFF
--- a/fava/static/css/journal-table.css
+++ b/fava/static/css/journal-table.css
@@ -22,7 +22,6 @@
   font-family: var(--font-family-monospaced);
   color: var(--color-text);
   text-align: right;
-  white-space: nowrap;
 }
 
 .flex-table .number {


yagebu:
That `nowrap` line is intentional, simply removing might break the intended layout in the journal tables According to the git blame, that line was introduced to prohibit breaking of the numbers themselves (and also the dates originally). However, the browser shouldn't break the numbers themselves in any case - still, have you checked that this doesn't lead to unwanted line breaks in the journal?

xuhcc:
Yes, I checked the journal as well and I didn't see any breakings there. On the contrary, the appearance on small screens has been somewhat improved because long numbers are no longer truncated:

Before

Screen Shot 2019-11-03 at 23 06 18

After

Screen Shot 2019-11-03 at 23 06 16

yagebu:
Thanks!

Issue #968, opened 2019-09-20
Adding jinja loopcontrols for benefit of extension reports


seltzered:
Hi, I recently wrote an [extension report](https://github.com/seltzered/fava-classy-portfolio) that makes use of the jinja extension jinja2.ext.loopcontrols due to how I render table data. 

I tried playing around with a few ways to enable the jinja option via my extension itself but wasn't having luck, it seemed like the only way to do this was to add `app.jinja_options["extensions"].append("jinja2.ext.loopcontrols")` to Fava's application.py. 

One workaround is for me to submit a PR to add this line officially into application.py, the other workaround is for me to refactor things in my extension so I don't need loopcontrols (which is arguably a better practice).

Is there any other workaround I'm missing that doesn't involve direct modification to application.py?

yagebu:
Don't have a strong opinion on this. If it makes it much easier to write extension, sure, go ahead and add it. If your extension would be more cleanly implemented without it, then don't ;)

yagebu:
@seltzered: ping

seltzered:
Apologies for the delay. Was debating about it and still think it helps in prototyping certain extension reports, so opened a PR.

yagebu:
@seltzered: If it's out of the prototype stage and you want to make your extension more easily available in Fava, I'll happily accept a PR to merge it into `fava.ext`

seltzered:
Honestly the portfolio extension I'm working on is still in a prototype state and think it's going to stay that way so I can think more about math / presentation before ensuring clean code. The other thing is that there may be more 'opinionated' things added on later in terms of required beancount metadata.

That said, with the PR in mind I've gone ahead and updated where I am right now: https://github.com/seltzered/fava-classy-portfolio . Perhaps we could add something to Fava's readme listing out user extensions?

yagebu:
> Perhaps we could add something to Fava's readme listing out user extensions?

#451 - someone just needs to add that page :)

PR #1017, opened 2019-11-23
Add Jinja loopcontrols option...

seltzered:
...while Fava standalone doesn't need it, I've found it helps in prototyping/developing extension reports.
Discussed in https://github.com/beancount/fava/issues/968


yagebu:
Thanks!

PR #1023, opened 2019-11-30
Use draggable button to move postings

### LINKED ISSUE #1015: Text selection is not working in "Add transaction" dialog
After a recent change (ad3b3435eef0df71d0224ba2676926f5aa548854) to "Add transaction" dialog, it is no longer possible to select text in postings with a mouse. The whole posting is entering a drag-n-drop mode instead.

As an alternative to making the whole posting draggable, I would suggest adding a small 'drag' button (e.g. [↕](https://unicode-table.com/en/sets/arrows-symbols/)) on the left side of the posting, near the 'remove' button:

Screen Shot 2019-11-20 at 19 11 06

xuhcc:
Resolves #1015

### DIFF
--- a/fava/static/css/entry-forms.css
+++ b/fava/static/css/entry-forms.css
@@ -67,11 +67,14 @@
   width: 10em;
 }
 
-.entry-form .remove-fieldset {
+.entry-form .remove-fieldset,
+.entry-form .drag-fieldset {
+  font-family: var(--font-family-monospaced);
   opacity: 0;
 }
 
-.entry-form .fieldset:hover .remove-fieldset {
+.entry-form .fieldset:hover .remove-fieldset,
+.entry-form .fieldset:hover .drag-fieldset {
   opacity: 1;
 }
 
--- a/fava/static/css/entry-forms.css
+++ b/fava/static/css/entry-forms.css
@@ -67,14 +67,12 @@
   width: 10em;
 }
 
-.entry-form .remove-fieldset,
-.entry-form .drag-fieldset {
+.entry-form .fieldset-action {
   font-family: var(--font-family-monospaced);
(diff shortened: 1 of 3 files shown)


fava/static/javascript/entry-forms/Posting.svelte
```diff
@@ -39,20 +39,26 @@
...
-  on:dragenter={dragenter}
-  on:dragover={dragenter}
-  on:dragleave={dragleave}
-  on:drop|preventDefault={drop}>
+<div class="fieldset posting" class:drag>
+  <button
+    class="muted round drag-fieldset"
+    draggable="true"
+    on:dragstart={dragstart}
+    on:dragenter={dragenter}
```
yagebu: This and all the following handlers should still apply to the whole `div`

fava/static/css/entry-forms.css
```diff
@@ -67,11 +67,14 @@
   width: 10em;
 }
 
-.entry-form .remove-fieldset {
+.entry-form .remove-fieldset,
+.entry-form .drag-fieldset {
```
yagebu: These two classes can be merged

fava/static/javascript/entry-forms/Posting.svelte
xuhcc: Fixed.

yagebu:
Mhh, actually this approach has a couple of problems: 1) I found it much harder to drag the posting since the draggable area is much smaller - I think the whole div apart from its children should be draggable. We can indicate that it's draggable by setting the cursor. 2) the preview image was generated just for the button not the whole posting.

I'll solve this by checking in the dragstart event handler whether we dragged the `<div>` or one of its children.

PR #1025, opened 2019-12-07
Use local time instead of UTC to fill transaction date

xuhcc:
The `toISOString()` method returns UTC datetime string (at least on my system: Linux, FF/Chrome). Because of this, the auto-filled date in transaction entry form could be inaccurate during certain hours (depending on user's time zone). This PR fixes it by using different method of converting date to string.

### DIFF
--- a/fava/static/javascript/entries.ts
+++ b/fava/static/javascript/entries.ts
@@ -1,6 +1,6 @@
 import router from "./router";
 import { notify } from "./notifications";
-import { dateToString } from "./format";
+import { todayAsString } from "./format";
 import { putAPI } from "./helpers";
 
 interface Posting {
@@ -25,7 +25,7 @@ abstract class Entry {
   constructor(type: string) {
     this.type = type;
     this.meta = {};
-    this.date = dateToString(new Date());
+    this.date = todayAsString();
   }
 }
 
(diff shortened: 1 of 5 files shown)


fava/static/javascript/entries.ts
```diff
@@ -24,7 +25,7 @@ abstract class Entry {
   constructor(type: string) {
     this.type = type;
     this.meta = {};
-    this.date = new Date().toISOString().slice(0, 10);
+    this.date = dateToString(new Date());
```
yagebu: There's two other places where this kind of logic is used, could you update those too? And maybe wrap the whole thing up as a `todayAsString()` function

fava/static/javascript/format.ts
```diff
@@ -69,3 +69,10 @@ export let currentTimeFilterDateFormat = timeFilterDateFormat.month;
 interval.subscribe(intervalValue => {
   currentTimeFilterDateFormat = timeFilterDateFormat[intervalValue];
 });
+
+export function dateToString(date: Date): string {
```
yagebu: This can use a formatter from d3-time-format, like dateFormat.day (but using `timeFormat` instead of `utcFormat` I guess)

xuhcc: You're right, `timeFormat` gives the same result as my implementation.

yagebu:
Thanks again :)

Issue #451, opened 2017-01-03
Add contributions to Beancount to Fava help pages


aumayr:
In light of new tools in the Beancount ecosystem like gnucash-to-beancount, add a dedicated section to the help pages that mirrors http://furius.ca/beancount/doc/contrib.

Fava should also contain a better guide on how to use Beancount.

blais:
After I'm done with the intro doc I'll work on automatically replicating
all the GDocs into your beautiful static HTML thingamajig.
Then I'll fix #149.
Then I'll do some cleanups.
Then I'll start a normal release process and christen 2.0.

aumayr:
All the Beancount documentation (what is now in GDocs) naturally lends itself to be integrated into Fava, so all the necessary documentation comes (at least linked-to) with the UI. 

> After I'm done with the intro doc I'll work on automatically replicating
all the GDocs into your beautiful static HTML thingamajig.

I recently came across another documentation based on Sphinx, which had a pretty neat front page. We could use this for the beancount documentation as well: 

![bildschirmfoto_2016-12-12_um_10 19 071](https://cloud.githubusercontent.com/assets/5135450/21671289/ac541e7e-d31a-11e6-870b-bc433b5b33b7.png)

aumayr:
@blais Any updates the 2017-new-year's-resolutions from above? 😇 

> Then I'll fix #149.
> Then I'll do some cleanups.
> Then I'll start a normal release process and christen 2.0.

If this is not relevant anymore I think we should close this issue.

yagebu:
@aumayr: Still interested in adding this help page?

aumayr:
Question is if it should me a mere mirror of the beancount contrib page or out own, maintained version. What do you think?

yagebu:
I think we can start with a short and opinionated list with just plugins that are pip-installable and seem to have some working tests (and then have a link to Martin's document).

yagebu:
This has gotten quite stale. I don't see much benefit in this, probably contributions are better centrally listed in Beancount's document.

PR #1032, opened 2020-01-12
Add a contributions help page (#451)

seltzered:
Made a contrib page to bring some closure to https://github.com/beancount/fava/issues/451 . 

 - Used 'gdocs2md' to migrate content from beancount contributions
 - Rather than trying to get too opinionated and remove contributions, I instead tried to improve how they were categorized.
 - A number of minor formatting details were also done, notably trying to have some better convention around links & naming.
 - Generally opted for writing out individual plugins rather than referencing generic collections of plugins. Could be further improved.

### DIFF
--- a/fava/help/__init__.py
+++ b/fava/help/__init__.py
@@ -11,4 +11,5 @@ HELP_PAGES = {
     "features": "Fava's features",
     "filters": "Filtering entries",
     "extensions": "Extensions",
+    "related_tools": "Related Tools",
 }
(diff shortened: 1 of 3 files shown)


yagebu:
Thanks for adding the page. However, I don't want to duplicate Beancount's Contributions page and I'd stick with what I said in #451:

> I think we can start with a short and opinionated list with just plugins that are pip-installable and seem to have some working tests (and then have a link to Martin's document).

If that limits it too much, I'd be against adding the contributions page to Fava - I don't want to encourage people to download random untested unmaintained plugins from some snippet, but rather encourage them to use and extend the official plugins and those at https://github.com/beancount/fava-plugins

seltzered:
Yeah, the "short and opinionated list" was on my mind, though when writing it I wasn't sure how to approach the curation process as someone who hasn't used too many plugins yet beyond the ones I made for myself, or used single-use tools that may not have been robustly written but were still highly useful.

An alternative thought might be to have some badging to mark tools that have tests or are known stable. Arguably could suggest such status markings (along with the cleanups & category improvements) could just be improvements to @blais 's beancount contributions document itself. In which case it may be better just to close this for now then.

PR #1035, opened 2020-01-13
fixes error when running pyinstaller build

johannesjh:
problem: the following code from `fava/__init.py__` does not work if running as pyinstaller bundle: 

```python
__version__ = get_distribution(__name__).version
```

proposed solution: this pull request provides a fallback for when fava is running as pyinstaller bundle. in this case, fava will read the version number from a bundled file version.txt

note: the file version.txt will written as part of `fava-electron`'s build process


yagebu:
I think you can also just copy the metadata with a PyInstaller hook, see for example this repository: https://github.com/cfclrk/python-demo

Issue #1034, opened 2020-01-13
Account graph broken when balance reaches zero


mickname:
The account graph renders incorrectly if the account's balance reaches exactly zero in the selected time range:
![Untitled](https://user-images.githubusercontent.com/3900234/72271059-dfd69900-362e-11ea-8856-ba0258e6e979.png)
This seems to be a regression when upgrading from fava 1.11 to 1.12. Related issue [https://github.com/beancount/fava/issues/880](https://github.com/beancount/fava/issues/880), except now the graph doesn't render at all?

mickname:
I did some digging and this bug is introduced in commit https://github.com/beancount/fava/commit/8159554205ff0bd831fc53032de3af5d29145eee.

PR #1037, opened 2020-01-14
Fix linechart y-domain when the minimum value in the range is zero.

mickname:
Fixes issue https://github.com/beancount/fava/issues/1034. If the minimum value in the line chart happens to be 0, the condition to set the chart domain is skipped.


yagebu:
Thanks for finding, debugging and fixing this issue :)

Issue #1041, opened 2020-01-21
Improve currency conversion filtering


chrishas35:
The fix for #854 (35bcfeab76d748965a86298ec8ad899d040419c8) involved limiting the list of commodities to convert to based upon their length.  Unfortunately, there are plenty of stocks and ETFs that meet this criteria as well. I was very confused as to why one, and only one, of my non-cash commodities would show up in the list until I dug into the source code to understand.

As I'm happy to provide a fix, my initial suggestion was going to be to leverage beancount's [operating currency option](https://docs.google.com/document/d/1wAMVrKIA2qtRGmoVDSUBJGmYZSygUaR0uOMW1GV3YE0/edit#heading=h.zetg7bya646o). Per documentation, that option allows for multiple currencies to be defined, and seems designed for reporting purposes such as this. But it also seems like the code is using this, and supplementing it with the 3 character list. If there's a compelling reason not to use the operating currencies list alone, I'd like to understand a bit more so a better conversion list can be presented to the user.

yagebu:
> If there's a compelling reason not to use the operating currencies list alone, I'd like to understand a bit more so a better conversion list can be presented to the user.

I think it's valuable to be able to see your reports converted to any given currency. Let's say your operating currency is EUR, but you'd like to see all expenses of a recent trip to the UK in GBP.

I think a preferable fix that should exclude your ticker symbols would be to check against the ISO 4217 currency codes.

chrishas35:
> I think a preferable fix that should exclude your ticker symbols would be to check against the ISO 4217 currency codes.

That makes sense. For dev reference: https://www.currency-iso.org/en/home/tables/table-a1.html. Changing to a simple in list check should suffice.

Is there a compelling reason to keep the operating currency piece separate? We could adjust the logic to be the sorted union of operating currencies and used commodities that are in the ISO list. Keeping operating currencies separate would have the benefit of putting operating currencies at the top of the selection list, if desired, though not sure how long typical lists would be.

yagebu:
> Is there a compelling reason to keep the operating currency piece separate? [...] Keeping operating currencies separate would have the benefit of putting operating currencies at the top of the selection list, if desired, though not sure how long typical lists would be.

I have 1 operating currency and 14 other currencies, so having the operating currency at the top might have it more easily accessible. Don't have a strong opinion on this though.

PR #1046, opened 2020-02-02
WIP: Filter conversion options to valid currencies

chrishas35:
Took a stab at this based on discussion in #1041.  Initial testing looks good, but was not exhaustive. I don't have a lot of experience with this app architecture, so wanted to throw this out there for feedback to make sure the approach is in line before testing further.


yagebu:
Do we need an option for this?

If we keep the option, I'd change the behaviour a bit: Have the option be empty by default (which implies checking against ISO4217) and if a list is given, just offer that complete list as options.

yagebu:
Closing this stale PR. Will reimplement this without the fava-option.

PR #1047, opened 2020-02-02
Add default-conversion option

OptikSage:
Use this option to specify a default conversion type. This option takes a string representing the same values available in the standard conversion dropdown.

Examples are:

-   `at_value` - At Market Value
-   `units`    - Units
-   `USD`      - Converted to USD
-   `EUR`      - Converted to EUR

If this option is not specified or the specificed value is not valid (i.e., a missing commodity), Fava defaults to "At Cost"


OptikSage:
I think I just realized there is a better way to do this. I'll submit an update soon.

OptikSage:
I modeled what I did after the interval option, but it was a more substantial change than I thought. The `test_linechart` test failed for some unknown reason - or more specifically it was passing prior even though it seemed to have only one snapshot (there are two function calls, but I think one overwrote the other). So I broke those apart and added some new ones to cover some more cases. Everything looks correct when I manually tested the different scenarios and the tests are now updated to match my changes.

James-E-A:
Was there any reason this was removed? &#x1f972;  
https://github.com/beancount/fava/commit/543357abb6468612ea8168017197c389c83cda94

yagebu:
See #1306:

> This option was removed as this behaviour can now be achieved with the default-page option.

PR #1050, opened 2020-02-09
Fava is not compatible with werkzeug 1.0

mbudde:


yagebu:
We want to be compatible with 1.0 though - from a quick glance it seems that only one import needs to be updated.

yagebu:
See #1051

mbudde:
That's another way to fix it 😅 Thanks!

PR #1065, opened 2020-02-21
Add missing trailing slash

### LINKED ISSUE #1064: Latest fava tries to use http for api requests when served over https with a reverse proxy
My setup uses a reverse proxy which does the https and fava itself is running on simply http (I use kubernetes for doing all of that).

This uses to work fine with 1.12 but since I upgraded to 1.14 it tries to access the api through http instead of https urls. (for things like imports or api/changed)

Is there a way I can force to use https for the api somehow?

tarioch:
fixes #1064


yagebu:
See #1066 instead

PR #1072, opened 2020-03-10
Escape filter when clicking on payee in the journal

wylfen:
As filters use regular expression matching, we need to escape certain
characters in the payee string when adding a filter for it after
clicking on it in the journal.

This commit adds an "escape" parameter to addFilter() in case any future
filters need escaping also. As "escape" defaults to false, there is no
impact to old code.

----

An example: Suppose a payee "PayPal (Europe)". Without escaping the payee string, clicking on the payee in the journal will not match that payee correctly. Instead, as the filter is processed as a regular expression, a capture group of "Europe" is created, matching "PayPal Europe". To match against the payee literally, a filter of "PayPay \\(Europe\\)" is needed.

This PR adds code to escape all special characters taken from [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions#Escaping)

### DIFF
--- a/fava/static/javascript/journal.ts
+++ b/fava/static/javascript/journal.ts
@@ -3,7 +3,11 @@ import e from "./events";
 import router from "./router";
 import { filters } from "./stores";
 
-function addFilter(value: string): void {
+function addFilter(value: string, escape: boolean = false): void {
+  if (escape) {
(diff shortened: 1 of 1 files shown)


yagebu:
Thank you! :)

Issue #1073, opened 2020-03-10
Improve help page on filtering entries: Clarify how filters match


wylfen:
Whilst working on https://github.com/beancount/fava/pull/1072, I came across a few issues in how filters are explained on the help page.

I've added the paragraphs in question below for reference:

> Filter by payee `payee:".*restaurant.*"`, or narration `narration:'Dinner with Joe'`, or any other entry attribute. The argument needs to be quoted (with `'` or `"`) if it contains spaces and can either be simply a string or a regular expression that matches the attribute. The regular expression matching is case insensitive.

> Search payee and narration by supplying a simple string, which again is matched as a regular expression. For Note directives, the comment will be searched.

#### 1. Drop the terminology of "simple" strings
In two places there is a reference to a "simple string" (or, "simply a string"). There is no indication of what a "simple" string is; no definition is given in the text. A user familiar with regular expressions might assume that a "simple" string is translated to a filter matching literally, but that is not the case.

On the other hand, a user unfamiliar with regular expressions will be confused once they enter a string containing special characters, as it might not match the term literally as they would expect.

#### 2. Inconsistent usage of the "simple" strings terminology
Compare:
> [..] can either be simply a string or a regular expression [..]

> [..] a simple string, which again is matched as a regular expression.

The first sentence implies that there are two types of strings, but the second one contradicts this. In reality (like the second sentence suggests), all quoted strings are matched as regular expressions.

#### 3. Clarify that matches are anchored to the beginning of a string
The help text does not clearly state that matches are only applied to the very beginning of a string. If familiar with regular expressions, this might be clear from a couple of examples (such as `payee:".*restaurant.*"`), but it could lead to confusion otherwise.

#### 4. Perhaps add a link to the documentation on the regex engine used
For people familiar with regular expressions (and others who are curious), it might be helpful adding a link to the Python [Regular Expression Syntax](https://docs.python.org/3/library/re.html?highlight=match#regular-expression-syntax).

----

I'd be quite happy to open a pull request for this, but would appreciate comments & feedback on the points mentioned above first.

yagebu:
> I'd be quite happy to open a pull request for this, but would appreciate comments & feedback on the points mentioned above first.

:+1: for improving these help texts! 1 and 4 sound good :)

Regarding 2:

There's actually one edge case where we look for an exact match: If the given string is not a valid regex. That behaviour is probably not ideal though, as one single character that makes an invalid regex changes the filter quite drastically without any feedback for the user why that happened.

[...]

PR #1076, opened 2020-03-12
Stop filters from being anchored to the beginning & filter help page improvements

wylfen:
See #1073.

### DIFF
--- a/fava/help/filters.md
+++ b/fava/help/filters.md
@@ -30,23 +30,22 @@ the account name, or a regular expression matching the account name, e.g.
 This final filter allows you to filter entries by various attributes.
 
 -   Filter by `#tag` or `^link`.
--   Filter by payee `payee:".*restaurant.*"`, or narration `narration:'Dinner with Joe'`,
-    or any other entry attribute. The argument needs to be quoted (with `'` or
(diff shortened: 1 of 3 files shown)


fava/help/filters.md
```diff
@@ -30,14 +30,13 @@ the account name, or a regular expression matching the account name, e.g.
...
 -   Filter by `#tag` or `^link`.
--   Filter by payee `payee:".*restaurant.*"`, or narration `narration:'Dinner with Joe'`,
-    or any other entry attribute. The argument needs to be quoted (with `'` or
-    `"`) if it contains spaces and can either be simply a string or a regular
-    expression that matches the attribute. The regular expression matching is
-    case insensitive.
--   Search payee and narration by supplying a simple string, which again is
-    matched as a regular expression. For Note directives, the comment will be
-    searched.
+-   Filter by any entry attribute, such as payee `payee:".*restaurant.*"` or
```
yagebu: The `.*`s can be removed. Maybe you could add an example with an anchored regex?

wylfen:
Removed the `.*` from all examples where they could be dropped or replaced with anchors. Also noticed that the help page still referred to `statement:`, which I changed to `document:`.

Also now has examples with both `$` and `^`.

Rebased on current master.

PR #1091, opened 2020-04-19
Fix Pattern typing in InsertEntryOption

xentac:
Apparently in python 3.6 you can't directly refer to an re.Pattern. Instead for typing purposes, it's an typing.Pattern.


xentac:
> Odd that mypy didn't complain on 3.6. `Pattern[str]` would be more precise.

Done.

yagebu:
Thanks!

PR #1092, opened 2020-04-22
remove unused args from extension

redstreet:


redstreet:
@seltzered: any objections?

seltzered:
I think the original intent was to handle Time filtering since nearly all standard views in Fava seem support Time Filtering. Feel like this is more of a bug/incomplete feature since I never passed the time arguments in.

redstreet:
Makes sense. Are you okay with cleaning it up for now to minimize confusion? You can add time filtering in the future.

seltzered:
Fine with removing it for now, let's create another issue so we don't forget to add it later.

redstreet:
Sounds good. I'll let you create the issue if that's cool, since you have a better sense of what you'd like added :).

@yagebu good to merge.

PR #1114, opened 2020-05-25
Minor metadata filter improvements

0:
Summary of changes:

* Escaped and anchored the filters that are created by clicking on the metadata. This ensures that only exact matches are shown.
* Made metadata keys clickable to filter on only the key. This creates filters of the form `key:""`, which allow anything for the value.
* Added `any()` to filters that are created by clicking posting metadata. This applies the filter to postings rather than entries.
* Added indicators for posting metadata. These make it visible at a glance that collapsed postings have metadata.

### DIFF
--- a/frontend/src/journal.ts
+++ b/frontend/src/journal.ts
@@ -84,8 +84,13 @@ export class FavaJournal extends SortableJournal {
       // Note: any special characters in the payee string are escaped so the
       // filter matches against the payee literally.
       addFilter(`payee:"^${escape(target.innerText)}$"`);
+    } else if (target.tagName === "DT") {
+      // Filter for metadata key when clicking on the key. The key tag text
+      // includes the colon.
+      addFilter(`${target.innerText}""`);
     } else if (target.tagName === "DD") {
-      // Filter for metadata when clicking on the value.
+      // Filter for metadata key and value when clicking on the value. The key
+      // tag text includes the colon.
       addFilter(
         `${(target.previousElementSibling as HTMLElement).innerText}"^${escape(
           target.innerText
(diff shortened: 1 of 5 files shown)


frontend/src/journal.ts
```diff
@@ -4,13 +4,17 @@ import { SortableJournal } from "./sort";
...
 
 /**
- * Add filter, possibly escaping it to produce a valid regex.
+ * Escape the value to produce a valid regex.
  */
-function addFilter(value: string, escape = false): void {
-  if (escape) {
-    value = value.replace(/[.*+\-?^${}()|[\]\\]/g, "\\$&");
-  }
+function escape(value: string): string {
```
yagebu, line 9: :+1: for extracting this as a function

frontend/src/journal.ts
```diff
@@ -79,14 +83,27 @@ export class FavaJournal extends SortableJournal {
       // Filter for payees when clicking on them.
       // Note: any special characters in the payee string are escaped so the
       // filter matches against the payee literally.
-      addFilter(`payee:"${target.innerText}"`, true);
+      addFilter(`payee:"^${escape(target.innerText)}$"`);
+    } else if (target.tagName === "DT") {
+      // Filter for metadata key when clicking on the key. The key tag text
+      // includes the colon.
+      const expr = `${target.innerText}""`;
+      if (target.className === "metadata-entry") {
```
yagebu, line 91: I think we can do without the extra classes by using the following check here:

    if (!target.closest('.postings')) {

0: That's clever! I undid the change to the template and updated these checks.

0:
Thank you for the feedback! I've made the requested change.

0:
As an aside, I was really impressed by how easy it was to set up a development environment for Fava, and to find and modify the relevant parts of the code for what I wanted to accomplish.

yagebu:
Thanks again!

> As an aside, I was really impressed by how easy it was to set up a development environment for Fava, and to find and modify the relevant parts of the code for what I wanted to accomplish.

That's great to hear :)

PR #1118, opened 2020-06-06
Support for fava option in accounts page

### LINKED ISSUE #1116: When using Filter, Balance/Changes screen should not show accounts which have zero value because of the filter
See this URL for example:
https://fava.pythonanywhere.com/huge-example-file/account/Expenses/balances/?filter=payee%3A%22Uncle+Boons%22

Uncle Boons filter is showing values for only Food and Restaurant, If we skip showing other accounts, it would be better.

prafullat:
Now account page supports fava options such as

show-accounts-with-zero-balance
show-accounts-with-zero-transactions

Fixes #1116

### DIFF
--- a/src/fava/templates/_tree_table.html
+++ b/src/fava/templates/_tree_table.html
@@ -96,7 +96,7 @@ Hold Ctrl or Cmd while clicking to expand one level.') %}
         {% endfor %}
         </p>
     </li>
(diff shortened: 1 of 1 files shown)


yagebu:
Closing this stale PR. Just reopen it if you ever want to continue this :)

PR #1123, opened 2020-06-29
Fix drag and drop events not being processed

### LINKED ISSUE #1122: Documents can't be imported via drag and drop
I recently updated to 1.15 and noticed today that I cannot import/link documents by dragging them onto transactions in the journal. This is the case for the current tip as well.

I bisected this issue back to https://github.com/beancount/fava/pull/1106/commits/a5eec6d6dc1e1d0b3e297107f5638492b591dcbc, which [moved](https://github.com/beancount/fava/pull/1106/commits/a5eec6d6dc1e1d0b3e297107f5638492b591dcbc#diff-a5599c85d6c6828d1fdef4448a8dbee5R17) validation of `event.dataTransfer.files.length` from `drop()` to `dragover()`. The relevant section now looks like this:

https://github.com/beancount/fava/blob/6951133380e7de6efce7b5d7b18079be83192b51/frontend/src/document-upload.ts#L15-L19

Back in `drop()` we still had read access to `event.dataTransfer.files`, but this is [no longer the case](https://html.spec.whatwg.org/multipage/dnd.html#dom-datatransfer-files) in `dragover()` because the data store is in [protected mode](https://html.spec.whatwg.org/multipage/dnd.html#concept-dnd-p):

> The files attribute must return a live FileList sequence consisting of File objects representing the files found by the following steps.
> [...]
> 3. If the drag data store's mode is the protected mode, Return the empty list L.

To fix this, we could instead use the [DataTransferItemList](https://html.spec.whatwg.org/multipage/dnd.html#the-datatransferitemlist-interface) interface, but it seems that this is [not implemented yet](https://developer.mozilla.org/en-US/docs/Web/API/DataTransfer/items#Browser_compatibility) in Safari.

Instead, I propose we look up the "Files" entry in `event.dataTransfer.types` which is [defined by the standard](https://html.spec.whatwg.org/multipage/dnd.html#concept-datatransfer-types):

>If there are any items in the DataTransfer object's drag data store item list whose kind is File, then add an entry to L consisting of the string "Files". (This value can be distinguished from the other values because it is not lowercase.)

This seems to be implemented across the board and is already used by the check for "text/uri-list".

wylfen:
When uploading documents via drag and drop, we want to make sure that
the DragEvent is only processed if it references dragged files. This is
done by checking the length of the 'event.dataTransfer.files' list.

In a5eec6d6 (support attaching (already uploaded) documents to txns,
2020-05-16), a check for dragged URLs was added and the validation of
'event.dataTransfer.files' was moved from 'drop()' to 'dragover()'.

The HTML Living Standard defines a 'drag data store mode' [1], which
governs read and write access to the drag data store. Specifically, if
the drag data store is in protected mode, the 'files' attribute must
return the empty list [2].

Before commit a5eec6d6, 'event.dataTransfer.files' was accessed in the
'drop' event, with the drag data store in read-only mode. Now, however,
we access it in the 'dragover' event, with the drag data store in
protected mode. The validation in 'dragover()' therefore fails because
'event.dataTransfer.files' is empty. For browsers that are compliant
with the HTML Living Standard, Fava does not process any drag and drop
events.

The standard also defines a 'DataTransferItemList' interface [3], whose
'length' attribute [4] can be used to get the number of items in the
drag data store. This is possible even in protected mode, but some
browsers (notably Safari) do not implement this yet [5].

Instead, similar to checking for URLs, solve this problem by checking
whether the 'types' list includes entries of type "Files". [6]

Closes #1122.

[1] https://html.spec.whatwg.org/multipage/dnd.html#drag-data-store-mode
[2] https://html.spec.whatwg.org/multipage/dnd.html#dom-datatransfer-files
[3] https://html.spec.whatwg.org/multipage/dnd.html#the-datatransferitemlist-interface
[4] https://html.spec.whatwg.org/multipage/dnd.html#dom-datatransferitemlist-length
[5] https://developer.mozilla.org/en-US/docs/Web/API/DataTransfer/items#Browser_compatibility
[6] https://html.spec.whatwg.org/multipage/dnd.html#concept-datatransfer-types

### DIFF
--- a/frontend/src/document-upload.ts
+++ b/frontend/src/document-upload.ts
@@ -14,7 +14,7 @@ import { put } from "./api";
 function dragover(event: DragEvent, closestTarget: HTMLElement): void {
   if (
     event.dataTransfer &&
-    (event.dataTransfer.files.length ||
+    (event.dataTransfer.types.includes("Files") ||
       event.dataTransfer.types.includes("text/uri-list"))
   ) {
     closestTarget.classList.add("dragover");


wylfen:
Should probably also mention that I tested this successfully in the following browsers: 
 - Firefox 77.0.1
 - Microsoft Edge 83.0.478.56
 - Chromium 83.0.4103.116

I sadly do not have easy access to a machine running Safari.

yagebu:
Thanks for debugging and fixing this issue.

The thorough explanation is great :)

PR #1124, opened 2020-07-16
Fixed tick format

baderj:
The y-Axis of line charts is formatted with ``.2s``, which sets the precision to two digits. This can lead to a problem, if the values use three digits. For instance, here is the chart of USD to CHF rates, which are just below 1. 

problem

The formatter uses the SI prefix "m", which is fine, but it also truncates the number to the two most significant digits. The issue is, that a few ticks now have the same label.

This pull request proposes to use the format ``.3s`` instead. With SI prefixes, all numbers should correctly display with 3 digits. Here is the same data as above with the fix applied:

fix

### DIFF
--- a/frontend/src/format.ts
+++ b/frontend/src/format.ts
@@ -30,7 +30,7 @@ export function formatPercentage(number: number): string {
   return `${formatterPer(Math.abs(number) * 100)}%`;
 }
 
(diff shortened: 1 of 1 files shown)


yagebu:
Thanks for this fix! :)

PR #1125, opened 2020-07-17
Added an optional area chart visualization wherever line charts can be drawn

baderj:
This pull request adds an option to all line charts (for example in "Commodities" or "Balance Sheet") to switch to an area chart. The Line chart is still the default, as it works much better for multiple value series, however, I prefer an area chart for single data series like net worth or stock prices.

The line chart looks as before and hasn't been modified:

default

Switching to "area" shows this new chart:

area

Hovering over data points works as for the line chart, but I took the liberty to make some style changes for the area chart:

- I removed the ``domain`` bounding box around the chart
- I changed the horizontal lines from dashed to solid, but decreased the opacity
- I removed the dots of the line

### DIFF
--- a/frontend/src/charts/LineChart.svelte
+++ b/frontend/src/charts/LineChart.svelte
@@ -3,9 +3,10 @@
   import { axisLeft, axisBottom } from "d3-axis";
   import { scaleLinear, scaleUtc } from "d3-scale";
   import { quadtree } from "d3-quadtree";
-  import { line } from "d3-shape";
+  import { line, area } from "d3-shape";
   import { getContext } from "svelte";
 
+  import { lineChartMode } from "../stores/chart";
   import { currenciesScale } from "./helpers";
   import { axis } from "./axis";
   import { formatCurrencyShort } from "../format";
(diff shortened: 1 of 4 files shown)


frontend/src/charts/LineChart.svelte
```diff
@@ -69,28 +75,54 @@
...
-        <path d={lineShape(d.values)} stroke={$currenciesScale(d.name)} />
-      {/each}
+      use:positionedTooltip={tooltipInfo}
+      transform={`translate(${margin.left},${margin.top})`}>
+      <g
+        class="x axis"
+        use:axis={xAxis}
+        transform={`translate(0,${innerHeight})`} />
+      <g class="y axis" use:axis={yAxis} />
+      <g class="lines">
```
yagebu, line 88: Since the axes are unchanged, the `{#if...` could start here

baderj:
Good points. I will try to keep the charts more consistent.

I noticed a problem with the label "Line": You are already using the term 'Line' in a different context, which means the button will be translated wrong in some languages. For example, in German to "Zeile" or in Dutch to "Regel". These are the correct translations for the lines of a program, but not the lines of a chart. 

Should I just name the buttons "Line Chart" and "Area Chart"?

yagebu:
> Should I just name the buttons "Line Chart" and "Area Chart"?

Sounds good, I'll make that change in the merge. In classic `gettext` there's `pgettext`, which allows you to specify a context. But since the translations part in JS is hand-built, it doesn't support that yet and it's probably not worth the effort.

yagebu:
Merged, thanks again!

I've changed the domain to be displayed again (with lower opacity), which fixed the showing of tooltips for me - and also looks slightly better IMHO. I've also unified the axes in the bar chart to look the same :)

Issue #1136, opened 2020-08-17
Show files list in a tree view in editor.


xiaoniu-578fa6bff964d005:
Current design have a draw back that the lone path is printed repeatedly for each file.
Is it possible we transform the drop down list of all included files in editor into a tree based on the path of these files?

yagebu:
Since the corresponding stale PR was closed, I'm also closing this issue. I don't plan to implement this but a PR would still be welcome.

PR #1137, opened 2020-08-18
add tree view for source files in editor.

xiaoniu-578fa6bff964d005:
Implement #1136 .

Some Previews:
s1
s2

Some possible improvements:
1. Tweak the style.
2. Decide whether we still want `sources` and `list_sources`.

### DIFF
--- a/frontend/src/editor/Editor.svelte
+++ b/frontend/src/editor/Editor.svelte
@@ -11,7 +11,7 @@
   import EditorMenu from "./EditorMenu.svelte";
   import SaveButton from "./SaveButton.svelte";
 
-  /** @type {{source: string, file_path: string, sha256sum: string, sources: string[]}} */
+  /** @type {{source: string, file_path: string, sha256sum: string, sources: string[], source_tree: object[]}} */
   export let data;
   /** @type {CodeMirror.Editor | undefined} */
   let editor;
@@ -22,6 +22,8 @@
   let sha256sum = "";
   /** @type {string[]} */
   let sources = [];
+  /** @type {object[]} */
+  let source_tree = [];
 
   let saving = false;
 
@@ -63,6 +65,7 @@
     sha256sum = data.sha256sum;
     file_path = data.file_path;
     sources = data.sources;
+    source_tree = data.source_tree;
 
     // keybindings when the focus is outside the editor
     const unbind = [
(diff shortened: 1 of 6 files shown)


yagebu:
Hi @xiaoniu-578fa6bff964d005, thanks for the PR :)

I'm not willing to merge such a substantial change without any tests and I'm somewhat hesitant to merge such a big change for this feature at all. Could this maybe be implemented more succinctly by not rendering a full tree, but rather by just splitting the path into directory and filename and then just grouping by directory? I think that would already vastly reduce the visual clutter.

xiaoniu-578fa6bff964d005:
Do you have any example of "splitting the path into directory and filename and then just grouping by directory"?

yagebu:
> Do you have any example of "splitting the path into directory and filename and then just grouping by directory"?

Well, it would look like your example image, but instead of having a tree of directories it would just be a list of directories and under each directory the files in that directory would be listed.

xiaoniu-578fa6bff964d005:
Does

> 
> 
> > Do you have any example of "splitting the path into directory and filename and then just grouping by directory"?
> 
> Well, it would look like your example image, but instead of having a tree of directories it would just be a list of directories and under each directory the files in that directory would be listed.

Does each directorey have the full path? Do we need to handle nested folders?

yagebu:
Hey @xiaoniu-578fa6bff964d005, sorry for the long silence, I was on vacation and somewhat forgot about this.

>  Does each directorey show the full path? Do we need to handle nested folders?

I don't really have a strong opinion on this. I guess ignoring nested folders and showing the full path would be "good enough" without having to add about 150 lines of code.

If you really want to implement this as a tree view, it should be done completely in the frontend. The `stratify` function in `frontend/src/lib/tree` could be used to generate the tree from the array of source files (it would have to be slightly modified to obtain the "parent" for a file path instead of an account path).

yagebu:
I'm closing this stale PR. If you want to continue to work on this, feel free to reopen

xiaoniu-578fa6bff964d005:
Sorry for the absence for almost one year.
During this time, have we introduced any other way to alleviate the problem that the file list is difficult to read when there are a large number of files?
In the past year I was using my implementation in the PR. Recently I wish to merge the upstream but it appears to be hard as many refactor has happened.
If there is any "good enough" design, I think I will just abandon the tree view in this PR and adopt design in the official repository.

PR #1157, opened 2020-09-26
Add ability to filter dates across a range of fiscal years

jpluscplusm:
This PR allows the date filter box to cope with a fiscal year range.

This allows a "FYxxxx" string to be the starting date in the range, the ending date, or both.

E.g.

- FY2011 - FY2015
- 2011 - FY2015
- FY2011 - 2015

... which all mean different things from one another (unless the fiscal year is configured to start on Jan 1st!)

It also updates the in-app help text for the fiscal-year option, letting the user know they can use the forms above.


yagebu (approved):
Thanks! This looks great :)

I'll add a small test to also check that the parsing still works when `fye` is `None`

jpluscplusm:
Thanks for merging @yagebu :-D

PR #1182, opened 2020-11-21
enable manual dispatching of the ci workflow

johannesjh:
this enables the `workflow_dispatch` feature, which allows to trigger the CI workflow manually through github's GUI, see https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,9 @@
 name: Test
 
 on:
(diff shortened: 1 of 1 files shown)


yagebu (changes requested):
Sounds good. Can you remove the comment?

yagebu (approved):
Thanks  :)

PR #1183, opened 2020-11-21
Pyinstaller build fixed and added to continuous integration

johannesjh:
This pull request fixes the pyinstaller build and runs it as part of the continous integratio, so it does not break again.

fixes for the pyinstaller build:

* corrected path to fava `cli.py`
* corrected paths to fava's data files
* added fava's and beancount's version info as data files to the binary built by pyinstaller

changes for running the pyinstaller build as part of the continuous integration:

* added makefile target for building and testing the binary
* added steps in the ci workflow to call the new makefile target
* enabled manual dispatching of workflows using github's `workflow_dispatch` feature. in other words, we can now click a button in github's UI to run the continuous integration.


yagebu (changes requested):
Thanks for the PR! Can you change the pyinstaller build to use tox?

yagebu:
I think it would also be good to rename the spec file to `pyinstaller_spec.py` (with the needed `#: pylint: disable=...` at the top) so that all the Python linting tools run on this file.

johannesjh:
yes, I'll give it a try

johannesjh:
It is working now. Tox now builds the binary using pyinstaller, and runs it to see if it works. This happens as part of the continuous integration workflow.

Note: The pyinstaller spec file cannot have a `.py` extension. I therefore used the default name for this file, which is `pyinstaller_spec.spec`. I added linting for this file, as well as for all python files in the contrib folder.

There are two remaining problems: 

* The sourcehut build fails, probably because of a missing / badly configured locale in archlinux, see this build output https://builds.sr.ht/~yagebu/job/349264#task-test. 
* The pyinstaller fava binary spits out improbably version information such as `0.1.dev1+gaa96b87`. Too much python magic involved in retrieving the version from git, and in copying according package metadata into the pyinstaller build. I think it is not unlike that the mechanism may work on the master branch but not on the pyinstaller branch.

johannesjh:
The sourcehut builds are now green as well.

yagebu:
Thanks, I think I'll split off the pyinstaller build from the test target, I don' think we need to force users to always run this, will do so in the merge.

BTW, what's the use case for this? This currently seems to create a whole folder, wouldn't a single executable be preferable?

> Note: The pyinstaller spec file cannot have a .py extension.

That's annoying...

> The pyinstaller fava binary spits out improbably version information such as 0.1.dev1+gaa96b87

This seems to work locally for me.

tox.ini
```diff
@@ -36,8 +36,18 @@ deps =
...
 
 [flake8]
 exclude = node_modules
 ignore = E203,W503
+
+[testenv:pyinstaller]
+deps =
+    pyinstaller
+    pylint>2.4.1
+allowlist_externals=*
```
yagebu, line 49: Why is this needed? (doesn't seem necessary locally)

johannesjh:
> Thanks, I think I'll split off the pyinstaller build from the test target, I don' think we need to force users to always run this, will do so in the merge.

ok 

> BTW, what's the use case for this?

I wanted to get the electron app working again, I think this could be a really convenient way to use fava. The first step was to fix the pyinstaller build target, and to add it to CI to prevent it from breaking again. 

> `allowlist_externals=*`
> Why is this needed? (doesn't seem necessary locally)

this is not needed, thank you! - I used this for debug output during development and then forgot to remove it.

PR #1187, opened 2020-11-28
Pyinstaller now builds a single binary

johannesjh:
Changed pyinstaller spec file so it builds a single binary file (instead of a folder that contains many files). The binary file is named `dist/fava`.

I tested the following:

* The github CI build is green (in my fork)
* And I manually verified that running `dist/fava example.beancount` produces a usable web interface.

### DIFF
--- a/contrib/pyinstaller_spec.spec
+++ b/contrib/pyinstaller_spec.spec
@@ -12,7 +12,7 @@ hidden_imports = collect_submodules(
 # add beancount version file:
 beancount_dir = os.path.dirname(importlib.import_module("beancount").__file__)
 beancount_version_file = os.path.join(beancount_dir, "VERSION")
-data_files += [(beancount_version_file, "beancount")]
+data_files = [(beancount_version_file, "beancount")]
 
 a = Analysis(
     ["../src/fava/cli.py"],
--- a/contrib/pyinstaller_spec.spec
+++ b/contrib/pyinstaller_spec.spec
@@ -41,23 +41,17 @@ a = Analysis(
 
 pyz = PYZ(a.pure, a.zipped_data, cipher=None)
 
-exe = EXE(
-    pyz,
-    a.scripts,
-    exclude_binaries=True,
-    name="cli",
-    debug=False,
-    strip=False,
-    upx=True,
-    console=True,
-)
-
-coll = COLLECT(
-    exe,
-    a.binaries,
-    a.zipfiles,
-    a.datas,
-    strip=False,
-    upx=True,
-    name="fava",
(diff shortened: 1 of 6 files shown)


contrib/pyinstaller_spec.spec
```diff
@@ -41,23 +41,17 @@ a = Analysis(
...
-    name="fava",
-)
+exe = EXE(pyz,
+          a.scripts,
+          a.binaries,
+          a.zipfiles,
+          a.datas,
+          [],
+          name='fava',
+          debug=False,
```
yagebu, line 51: Can you remove all the kwargs that, just like this one, simply correspond to the default value?

yagebu (changes requested):

contrib/pyinstaller_spec.spec
johannesjh: done... I removed unnecessary kwargs in 5aa256e44016e774febca7cd8dd1b94105d69e2e

johannesjh:
Instead of listing fava's data files in fava's pyinstaller specfile, I provided a pyinstaller hook that lists theses files. The advantage is that other libraries or applications that may want to freeze fava now automatically benefit from this.

Note: In a similar way, it would be nicer to provide a pyinstaller hook directly within beancount, instead of patching beancount's pyinstaller build from the outside.

yagebu:
I don't really see the advantage of the hooks and will remove them. They split up PyInstaller-related logic from one file to four files and Fava is an application not a library so I don't really see many others making use of these hooks.

yagebu (approved):

johannesjh:
Hi, thank you for your consideration! 

My motivation for adding the hooks was that they may help to create apps that wrap fava with a native window. For example, I started experimenting with a Python GTK App using PyGObject. One way of building the app is to freeze it, including all the libraries that the app depends on. In this case, fava and beancount are seen as libraries, not applications, and the hooks do simplify the build process.

But I don't strictly need the hooks. In my experimental Python GTK app, I switched away from using pyinstaller because the resulting file size was too large... I think pyinstaller included WebKit into the file. So feel free to remove the hooks, or simply revert to commit 5aa256e which is a working version without the hooks.

yagebu:
I've switched the spec to use `pyinstaller.utils.hooks`, which turns the specification of data files for Fava and Beancount into one-liners (e9d166694bab60e3a9ef1cffe5c10ab4f85b3924). I think that's better than adding hooks for Fava and Beancount, since that allows to any other applications bundling them to include more/less of the data files, depending on what they actually need.

Issue #1189, opened 2020-12-08
WebKit JavaScript Error "The superclass is not a constructor" on fava's editor page


johannesjh:
I created a python + GTK application to be able to conveniently run fava as a desktop application, see attached screenshot. This works fine except for the editor page, where the following error is thrown "The superclass is not a constructor".

```
  /**
   * Read-only editors in the help pages.
   */
  class BeancountTextarea extends HTMLTextAreaElement {  // <-- The superclass is not a constructor
      constructor() {
          super();
          codemirror.fromTextArea(this, {
              mode: "beancount",
              readOnly: true,
          });
      }
  }
```

Any ideas how to fix or shim this error? (I tried loading custom-elements shims, which apparently should allow to extend HTMLElement Subclasses, but it did not help the issue).

thank you!

PS:  
Screenshot of the application... I find that fava's GUI works perfectly fine in GTK's embedded webkit browser, except for this error on the editor page.

![Screenshot from 2020-12-08 13-21-33](https://user-images.githubusercontent.com/581188/101483959-304a9980-3959-11eb-95db-22b59642a443.png)

yagebu:
It's interesting that this breaks the Editor report and not others, since that code isn't used on the editor page but only on the help page. The error itself is due to webkit not yet properly supporting custom elements (https://caniuse.com/custom-elementsv1)

johannesjh:
Thank you! - I'll try to debug this a little more. There has to be a different reason why the editor page is not working. (The superclass error appears on other pages as well, and does not prevent them from functioning in a normal way)

johannesjh:
The [ungap/custom-elements]( https://github.com/ungap/custom-elements) polyfill solves the problem. I now have a working editor page, and my overall impression is that everything is working fine inside the webkit GTK-widget. 

I used a quick and dirty fix in order to test the polyfill, i.e., I added the following line to `_layout.html` just before `app.js` is loaded:

```
<script src="//unpkg.com/@ungap/custom-elements"></script>
```

Obviously, the script should be loaded in a different way. Before we get into this, my question is:

Would it be acceptable to you to add this polyfill (or a similar one) to fava?

johannesjh:
Note: The smaller `es` variant of the same polyfill also works. This has 3.25KB if transmitted using gzip compression, 5KB without compression.

```
<script src="//unpkg.com/@ungap/custom-elements/es.js"></script>
```

yagebu:
[...]

PR #1192, opened 2020-12-22
adds @ungap/custom-elements polyfill

johannesjh:
Hi, my first attempt at adding the custom elements polyfill. I verified that the polyfill is loaded and that it is effective regarding #1189. 

Feedback welcome regarding how the polyfill should be added: 

* Should the polyfill also be loaded by modern browsers that don't need it? (I would argue for a yes, for the sake of simplicity).
* Does it make sense to create a polyfills.ts file even if we have just one polyfill?
* My version of npm created a lock file version 2, and thus a rather large diff compared to fava/master. Is this acceptable?

thank you for your feedback.

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -1,7 +1,7120 @@
 {
   "name": "fava",
+  "lockfileVersion": 2,
   "requires": true,
-  "lockfileVersion": 1,
+  "packages": {
+    "": {
+      "name": "fava",
+      "dependencies": {
+        "@openfonts/fira-mono_all": "^1.44.2",
+        "@openfonts/fira-sans_all": "^1.44.1",
+        "@openfonts/source-code-pro_all": "^1.44.2",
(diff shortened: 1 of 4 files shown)


yagebu:
> Should the polyfill also be loaded by modern browsers that don't need it? (I would argue for a yes, for the sake of simplicity).

Yes.

> Does it make sense to create a polyfills.ts file even if we have just one polyfill?

No, let's just add it to main.ts directly.

> My version of npm created a lock file version 2, and thus a rather large diff compared to fava/master. Is this acceptable?

I'd rather stick to the version produced by the stable npm for now, i.e. version 1. I can fix that when merging.

johannesjh:
Thank you for your feedback! 

I moved the import to main.ts.

yagebu:
Thanks! :)

Merged in e72ea4235185fba028de8fa78f41ad3224933aa1

PR #1196, opened 2020-12-27
Add click handler to transactions in query table

mattmattmatt:
This allows for opening and editing of transactions straight from the query table. See screenshots below.

image
image


yagebu:
Thanks for the PR, this looks like a nice improvement :)

mattmattmatt:
Glad it can be useful for others upstream as well!

PR #1199, opened 2021-01-01
Fix filter with meta key error

xinluh:
With filter like `any(some_meta_name:"^value$")` (which happens when clicking on a meta value in the entry list), it results in error:

```
  File "<...>/lib/python3.8/site-packages/fava/core/filters.py", line 279, in _key
    if key in entry.meta:
TypeError: argument of type 'NoneType' is not iterable
```

### DIFF
--- a/src/fava/core/filters.py
+++ b/src/fava/core/filters.py
@@ -276,7 +276,7 @@ class FilterSyntaxParser:
         def _key(entry):
             if hasattr(entry, key):
                 return match(str(getattr(entry, key) or ""))
-            if key in entry.meta:
+            if entry.meta is not None and key in entry.meta:
                 return match(str(entry.meta.get(key)))
             return False
 
--- a/tests/test_core_filters.py
+++ b/tests/test_core_filters.py
@@ -3,6 +3,7 @@ import datetime
 
 import pytest
 from beancount.core import account
+from beancount.loader import load_string
 
 from fava.core.filters import AccountFilter
 from fava.core.filters import AdvancedFilter
@@ -128,6 +129,22 @@ def test_advanced_filter(example_ledger, string, number):
     filtered_entries = FILTER.apply(example_ledger.all_entries)
     assert len(filtered_entries) == number
     FILTER.set("")
+
+
+def test_null_meta_posting():
+    FILTER.set('any(some_meta:"1")')
+
+    # pad entry produces Posting with null meta:
+    entries, _, _ = load_string("""
+2000-01-01 open Assets:Bank USD
+2000-01-01 open Equity:OpeningBalances USD
+
+2000-01-02 pad Assets:Bank Assets:OpeningBalances
+2000-01-03 balance Assets:Bank 1.00 USD
+    """)
+
+    filtered_entries = FILTER.apply(entries)
+    assert len(filtered_entries) == 0
 
 
 def test_account_filter(example_ledger):
--- a/tests/test_core_filters.py
+++ b/tests/test_core_filters.py
@@ -135,13 +135,15 @@ def test_null_meta_posting():
     FILTER.set('any(some_meta:"1")')
 
     # pad entry produces Posting with null meta:
-    entries, _, _ = load_string("""
+    entries, _, _ = load_string(
+        """
 2000-01-01 open Assets:Bank USD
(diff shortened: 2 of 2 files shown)


yagebu:
All Beancount entries should have a metadata dictionary, see https://github.com/beancount/beancount/blob/7a4fa841d122496ad0c2cdd501160a2ad13795eb/beancount/core/data.py#L490, so if that error is thrown, you might have some plugin that produces invalid Beancount entries?

wylfen:
I've been running into this problem as well. Decided to investigate a bit further by selectively disabling all my plugins, and you are indeed correct (in my specific case here, I don't know about @xinluh). Turns out that disabling the [split_income](https://github.com/beancount/fava-plugins/blob/master/fava_plugins/split_income.py) plugin resolves this issue.

However, the issue is a bit more complex than that. The problem arises not from `meta` being `None` in a Transaction, but in a Posting. These are [not checked](https://github.com/beancount/beancount/blob/7a4fa841d122496ad0c2cdd501160a2ad13795eb/beancount/core/data.py#L506-L512) in the `sanity_check_types` routine you linked (note that Posting is not contained in `ALL_DIRECTIVES`, so the earlier generic check for `entry.meta` ignores this).

We can also see there that most of the attributes in Posting are optional, and that `sanity_check_types` is fine with `NoneType` for most of these. See also the constraints on `meta` for Posting [here](https://github.com/beancount/beancount/blob/7a4fa841d122496ad0c2cdd501160a2ad13795eb/beancount/core/data.py#L194). So even if beancount added a sanity check for a Posting's `meta` attribute, it would probably allow `NoneType`.

Now, the story doesn't end there. If `meta` was usually empty for Postings, we should have seen this issue way earlier. Even though beancount says `meta` can be optional, its parser (the place most Postings come from) initialises `meta` using `new_metadata` [here](https://github.com/beancount/beancount/blob/7a4fa841d122496ad0c2cdd501160a2ad13795eb/beancount/parser/grammar.py#L804). The reason `meta` stays empty with split_income is that it uses [create_simple_posting](https://github.com/beancount/fava-plugins/blob/9675bf239dfb892b28d82946c2a4a5322d8014b0/fava_plugins/split_income.py#L130), which [sets it](https://github.com/beancount/beancount/blob/34440cec5667cd376d2a5d6f9cfd2a25198b1c66/beancount/core/data.py#L484) to `None`. Other plugins might be using this routine as well, and end up with an empty `meta` attribute.

Whilst reading through beancount's code, I came across [this](https://github.com/beancount/beancount/blob/7a4fa841d122496ad0c2cdd501160a2ad13795eb/beancount/core/data.py#L76-L79) comment:
```
#   meta: A dict of strings to objects, potentially attached to each of the
#     directive types. The values may be strings, account names, tags, dates,
#     numbers, amounts and currencies. There are two special attributes which
#     are always present on all directives: 'filename' and 'lineno'.
```

It seems to me that `meta` shouldn't be optional then, after all. Given that the parser takes care to initialize those two special attributes, maybe split_income should do this as well. It [does it](https://github.com/beancount/fava-plugins/blob/9675bf239dfb892b28d82946c2a4a5322d8014b0/fava_plugins/split_income.py#L124) for the Open directive already, even.

As it currently stands, `meta` is optional for Postings, so I think fava should check for `None`. However, given the previous paragraph, I am not sure it is _supposed_ to be optional.  Perhaps @blais can comment on this too.

xinluh:
In my case the null meta is indeed created by a plugin. It is not clear to me if `posting.meta` is guaranteed non-null (unlike `entry.meta`) - either way, the null check seems pretty harmless to add for robustness though.

yagebu:
I didn't consider that this might also run over postings, you're quite right that for postings meta is currently optional, so we should check for `None`.

@xinluh: Could you add a test for this?

xinluh:
@yagebu I take it back on the null posting meta being plugin related. It can happen without plugin - a standard `pad` entry also produces it. I added a test using the standard `pad` entry as example - the test would fail to run without this change.

yagebu:
Yep, since postings have optional metadata, this can happen with Beancount's entries as well. I've changed the test to use `create_simple_posting` to make the test a bit more explicit (with a Pad entry it happens in the auto-generated Transaction).

PR #1200, opened 2021-01-01
Highlight journal row on hover to make it easier to read on wide screen

xinluh:
On a wide screen, it is really hard to match the left side (date, payee, etc.) of an entry with the amount. On hover highlighting helps with that. 

The color is the same shade of gray as in the UI but with some more transparency to make it less intrusive:
image

### DIFF
--- a/frontend/css/journal-table.css
+++ b/frontend/css/journal-table.css
@@ -108,7 +108,7 @@
 }
 
 .journal .transaction:hover {
-    background: #f5f5f59c;
+  background: var(--color-journal-hover-highlight);
 }
 
 /* Metadata */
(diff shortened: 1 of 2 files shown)


frontend/css/journal-table.css
```diff
@@ -107,6 +107,10 @@
   display: block;
 }
 
+.journal .transaction:hover {
+    background: #f5f5f59c;
```
yagebu, line 111: Please use a CSS variable for this.

yagebu (changes requested):
Thanks for the PR :)

yagebu (approved):

Issue #1049, opened 2020-02-08
Allow to set/guess default page/date range etc.


ngoonee:
When first opening fava, the default is to land on the 'Income Statement' page, with no datetime restrictions.

In practice, I just save a bookmark to balance_sheet/?time=2019 which I then have to remember to update to 2020 etc.

The reason I do this is because I mainly care about balances on a day-to-day basis when checking in. Also because with historical transactions for 10 years, clicking in to active accounts just freezes my slower devices (the server isn't too hot either) if I don't have a time restriction.

Would it be possible to add fava-options for at least time range (and perhaps default page, though I can see where this gets very clunky). So basically controlling the redirect you get when you visit fava_url/

I envision something like
```
2020-01-01 custom "fava-option" "default_page" "balance_sheet"
2020-01-01 custom "fava-option" "default_time" "year-1 - year"
```
And what then happens is when I visit fava_url/ I get redirected to fava_url/my_account_name/balance_sheet/?time=year-1+-+year

grostim:
I second that feature request.
I have already entered all the future reimburement for my real estate loan until 2030... 
I would prefer to have the default page restricted to the current year. Even better if we can set it throgh fava options.

yagebu:
I think a bookmark solves this case quite well for most cases (in particular, just like you suggest using the `year` filter in the 'default_time', you can use that in your bookmark).

If you insist that a bookmark is really not enough, I'd not be terribly opposed to *one* option to control the redirect if someone provides a PR for this.

aclindsa:
@yagebu 

> I think a bookmark solves this case quite well for most cases (in particular, just like you suggest using the `year` filter in the 'default_time', you can use that in your bookmark).

I won't debate that a bookmark is one way you could solve this issue, but I'm curious why you're hesitant to add a default date range option when there are seemingly-similar 'conversion' (i.e. 'At Market Value' vs. 'At Cost', etc.) and 'interval' (monthly, weekly, yearly...) options. FWIW, one reason I would prefer an option over a bookmark is the convenience of not having to remember to use such a bookmark or sync it across different devices when it changes.
 
> If you insist that a bookmark is really not enough, I'd not be terribly opposed to _one_ option to control the redirect if someone provides a PR for this.

[...]

PR #1201, opened 2021-01-02
Add 'default-path' option

aclindsa:
This allows the user to configure where they'd like to fava to redirect
them to if they visit the default path (previously it was effectively
hard-coded to redirect you to '/income_statement/')

This is my first draft towards addressing https://github.com/beancount/fava/issues/1049

### DIFF
--- a/src/fava/application.py
+++ b/src/fava/application.py
@@ -274,7 +274,11 @@ def index():
     """Redirect to the Income Statement (of the given or first file)."""
     if not g.beancount_file_slug:
         g.beancount_file_slug = next(iter(app.config["LEDGERS"]))
-    return redirect(url_for("report", report_name="income_statement"))
+    index_url = url_for("index")
+    default_path = app.config["LEDGERS"][g.beancount_file_slug].fava_options[
+        "default-path"
+    ]
+    return redirect(f"{index_url}{default_path}")
 
 
 @app.route("/<bfile>/account/<name>/")
--- a/src/fava/core/fava_options.py
+++ b/src/fava/core/fava_options.py
@@ -47,6 +47,7 @@ DEFAULTS = {
     "auto-reload": False,
     "conversion": "at_cost",
     "default-file": None,
+    "default-path": "income_statement/",
     "fiscal-year-end": FiscalYearEnd(12, 31),
     "import-config": None,
     "import-dirs": [],
@@ -103,6 +104,7 @@ LIST_OPTS = [
 STR_OPTS = [
     "collapse-pattern",
     "conversion",
+    "default-path",
     "import-config",
     "interval",
     "language",
--- a/src/fava/help/options.md
+++ b/src/fava/help/options.md
@@ -53,6 +53,29 @@ If this option is not specified, Fava opens the main file by default.
 
 ---
 
+## `default-path`
+
+Default: `income_statement/`
+
+Use this option to specify the path to be redirected to when visiting the main
+Fava webpage. If this option is not specified, you are taken to the income
+statement. You may also use this feature to set options via the URL. For
+example, a `default-path` of `balance_sheet/?time=year-2+-+year` would result in
+you being redirected to a balance sheet reporting the current year and the two
+previous years.
+
+Note that the supplied path is relative. To be taken to your trial balance
+report at `{{url_for("report", report_name="trial_balance", account="Expenses:Restaurants", time="2021")}}` put the following in your
+beancount file:
+
+<pre><textarea is="beancount-textarea">
+2021-01-01 custom "fava-option" "default-path" "trial_balance/?account=Expenses%3ARestaurants&time=2021"</textarea></pre>
+
+It is probably easiest to navigate to the URL in your browser and copy the
+portion of the URL after the 'title' of your beancount file into this option.
+
+---
+
 ## `interval`
 
 Default: `month`
--- a/tests/test_application.py
+++ b/tests/test_application.py
@@ -60,6 +60,57 @@ def test_urls(test_client, url, return_code):
 
 
 @pytest.mark.parametrize(
+    "url,option,expect",
+    [
+        ("/", None, "/long-example/income_statement/"),
+        ("/long-example/", None, "/long-example/income_statement/"),
+        ("/", "income_statement/", "/long-example/income_statement/"),
+        (
+            "/long-example/",
+            "income_statement/",
+            "/long-example/income_statement/",
+        ),
+        (
+            "/",
+            "balance_sheet/?account=Assets:US:BofA:Checking",
+            "/long-example/balance_sheet/?account=Assets:US:BofA:Checking",
+        ),
+        (
+            "/long-example/",
+            "income_statement/?account=Assets:US:BofA:Checking",
+            "/long-example/income_statement/?account=Assets:US:BofA:Checking",
+        ),
+        (
+            "/",
+            "balance_sheet/?time=year-2+-+year",
+            "/long-example/balance_sheet/?time=year-2+-+year",
+        ),
+        (
+            "/",
+            "balance_sheet/?time=year-2 - year",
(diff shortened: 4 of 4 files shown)


src/fava/application.py
```diff
@@ -270,11 +270,19 @@ def fava_api_exception(error: FavaAPIException):
...
 @app.route("/")
 @app.route("/<bfile>/")
-def index():
+def index(bfile=None):
     """Redirect to the Income Statement (of the given or first file)."""
     if not g.beancount_file_slug:
         g.beancount_file_slug = next(iter(app.config["LEDGERS"]))
-    return redirect(url_for("report", report_name="income_statement"))
+    default_path = ""
+    if not bfile:
```
yagebu, line 278: I don't quite see why this is necessary - we need to prepend the slug in any case, no?

src/fava/application.py
```diff
@@ -270,11 +270,19 @@ def fava_api_exception(error: FavaAPIException):
...
+        default_path += g.beancount_file_slug + "/"
+    default_path += (
+        app.config["LEDGERS"][g.beancount_file_slug]
+        .fava_options["default-path"]
+        .lstrip("/")
```
yagebu, line 283: Remove this, let's make this option a bit stricter by having to specify a relative path.

src/fava/application.py
```diff
@@ -270,11 +270,19 @@ def fava_api_exception(error: FavaAPIException):
...
+    )
+    return redirect(f"/{default_path}")
```
yagebu, line 285: This can't be an absolute path at the root but rather needs to respect the path that WSGI app is mounted at.

yagebu (changes requested):
Thanks for the PR!

I've left some comments in the code. Does this work correctly with characters that need to be escaped in URLs, like spaces?

src/fava/application.py
aclindsa: I thought based on the `@app.route` lines before the definition of this function:
```
@app.route("/")
@app.route("/<bfile>/")
```
that the intended behavior was to handle redirection here even when the slug was already present. I found that if I unconditionally prepended it, I ended up with the slug in the URL twice if I visited a URL with the slug already present.

Maybe this is an artifact of how I'm doing redirection here?

src/fava/application.py
aclindsa: To be clear, you're asking me to remove only `.lstrip("/")`, correct?

src/fava/application.py
aclindsa: I think I understand what you're saying in terms of needing to be a relative path, but am not entirely sure I understand how to translate that into the implementation here. Is it as simple as not appending the leading "/" on this line?

aclindsa:
> Does this work correctly with characters that need to be escaped in URLs, like spaces?

I was primarily testing it with `2021-01-01 custom "fava-option" "default-path" "balance_sheet/?time=year-2+-+year"` and it worked as expected. Do you have other specific things you'd like me to test with?

src/fava/application.py
yagebu: Yes.

src/fava/application.py
yagebu: > I found that if I unconditionally prepended it, I ended up with the slug in the URL twice if I visited a URL with the slug already present.

That's odd, in my quick testing, `bfile` was always None (since it gets removed by the `_pull_beancount_file`) and always prepending it works.

src/fava/application.py
yagebu: I think it would be better to keep this as an absolute path, but you'll need to use Flask to figure out what the path prefix is, `url_for('index')` might work.

src/fava/application.py
aclindsa: You're right, I'll remove this unnecessary logic. I'm not 100% sure what I was seeing before, but the only way I've been able to re-create what I was seeing before is to use a relative path. I suspect I falsely attributed a change.

src/fava/application.py
aclindsa: I've done this and updated the documentation to reflect that it is required to be a relative path.

src/fava/application.py
aclindsa: Okay, `url_for('index')` does appear to be working, as you suggested. Thanks!

yagebu:
> I was primarily testing it with 2021-01-01 custom "fava-option" "default-path" "balance_sheet/?time=year-2+-+year" and it worked as expected. Do you have other specific things you'd like me to test with?

What happens if you put actual spaces in where the `+`es are? Or, say, a `:` in the query string (`income_statement/?account=Assets:SomeAccount` which should actually be `income_statement/?account=Assets%3ASomeAccount`).

Overall, this now looks good :) This should have tests though - a bit like the tests [here](https://github.com/beancount/fava/blob/8f3cea7cc8cc1ed74a54cafc95aa844eff254fed/tests/test_application.py#L62), for setting the option to test values, you can access them at `flask.g.ledger.fava_options`.

I'm totally happy with the option name yet, "path" sounds more file-system-related to me and not like something to specify the page I want to land on. I think I'd prefer `default-report`, which to me sounds more like the answer to "which option do I want to set", whereas "path" is more the "how do I set this option" detail.

aclindsa:
I'll look into the other characters, though I don't actually expect a raw space or raw colon to work as expected. What is your expectation/preferred behavior there? I was sort of assuming that folks would end up copying the URL out of the address bar in their browser of choice into this option rather than constructing it themselves. Plus, it seems difficult to attempt to auto-detect whether it's properly escaped or not.

And I'll look into adding tests, too. Might be a few days depending on workload before I get around to that.

> I'm totally happy with the option name yet, "path" sounds more file-system-related to me and not like something to specify the page I want to land on. I think I'd prefer `default-report`, which to me sounds more like the answer to "which option do I want to set", whereas "path" is more the "how do I set this option" detail.

I think I see your point. I said 'path' because I wasn't sure folks would always want to redirect to what I would consider 'reports'. For instance, if they wanted to seeing the editor or import page every morning they visited. I wouldn't expect that to be the most common case, but it seemed like something I could see people doing. Brainstorming, how would you feel about 'default-redirect' or 'homepage'?

aclindsa:
> I'll look into the other characters, though I don't actually expect a raw space or raw colon to work as expected.

I'm happy to report that I was entirely wrong. The redirection works just fine even with non-escaped spaces and colons.

aclindsa:
@yagebu I've now added tests - let me know if I did that the way you intended.

I believe the only comment I haven't addressed is about the name of the option itself... I'll wait on further direction from you before making that change.

aclindsa:
@yagebu Is there anything else you'd like for me to do before this could be merged (In particular the name of the option)?

yagebu (approved):
> Is there anything else you'd like for me to do before this could be merged (In particular the name of the option)?

Nope, this looks good :) 

I'll just make the option name change when I merge (and go with `default-page` I think)

yagebu:
Merged, thank you!

JaneSmith:
Is it possible to specify a sub-page with this option? I currently have `default-page` set to `trial_balance/?conversion=GBP`, and when I open Fava-GTK it defaults to the Expenses sub-page of the Trial Balance report. I want it to default to Assets — is that possible?

aclindsa:
> Is it possible to specify a sub-page with this option? I currently have `default-page` set to `trial_balance/?conversion=GBP`, and when I open Fava-GTK it defaults to the Expenses sub-page of the Trial Balance report. I want it to default to Assets — is that possible?

I'm not 100% sure I understand what you mean when you say "sub-page", but I'll try to answer generically. And @yagebu can correct me if I'm wrong, but in general I believe if copying and pasting the current URL in the address bar into another browser window/tab gets you the page you want, you should be able to use that same URL fragment as a `default-page` setting. So, if your desired "sub-page" is uniquely identified by a URL, you should be able to use it; if not, it won't work.

Issue #975, opened 2019-09-28
Display Commodity meta-name


grostim:
Hi,

I think, it could be useful to find a way to display the meta-name of a commodity .

When you monitor a portfolio with many ETF or Funds, the tickers are not always explicit... (and when you decide like i do to use the ISIN code as commodity name, it is not explicit at all).

I would like to discuss a way to display the complete name of the commodity (from the metatag "name").
I suggest displaying an info-bubble when a commodity is hovered by the pointer, but i will be happy to discuss any alternative solution.

yagebu:
I like it :)

> I suggest displaying an info-bubble when a commodity is hovered by the pointer

Sounds good.

PR #1207, opened 2021-01-04
Show commodity and account names on hover

mckelvin:
For the US markets, it's easy to memorize the mapping between commodity (AAPL) and the company (Apple Inc.). But it's hard for most Asian markets as tickers are just numbers. This PR is created for solving this issue.

After this PR, commodities and accounts with `name` will display the name accordingly. Commodities without `name` will not changed.

```
2021-01-01 commodity AAPL
  name: "Apple Inc."

2021-01-01 commodity CN_600519
  name: "Kweichow Moutai"

2021-01-01 open Assets:Stock:IB:Positions USD, HKD
  name: "Interactive Brokers Positions"
```

Please let me know what you guys think about it. Thanks and happy new year!

### DIFF
--- a/src/fava/core/__init__.py
+++ b/src/fava/core/__init__.py
@@ -21,6 +21,7 @@ from beancount.core.account_types import get_account_sign
 from beancount.core.compare import hash_entry
 from beancount.core.data import Balance
 from beancount.core.data import Close
+from beancount.core.data import Commodity
 from beancount.core.data import Custom
 from beancount.core.data import Directive
 from beancount.core.data import Document
@@ -128,6 +129,7 @@ class FavaLedger:
         "all_entries_by_type",
         "all_root_account",
         "beancount_file_path",
+        "commodities",
         "_date_first",
         "_date_last",
         "entries",
@@ -200,6 +202,9 @@ class FavaLedger:
         #: A dict containing information about the accounts.
         self.accounts = AccountDict()
 
+        #: A dict containing information about the commodities
+        self.commodities: Dict[str, Commodity] = {}
+
         #: A dict with all of Fava's option values.
         self.fava_options: FavaOptions = {}
 
@@ -240,6 +245,11 @@ class FavaLedger:
             self.accounts.setdefault(
                 cast(Close, entry).account
             ).close_date = entry.date
+
+        self.commodities = {}
+        for entry in entries_by_type[Commodity]:
+            commodity = cast(Commodity, entry)
+            self.commodities[commodity.currency] = commodity
 
         self.fava_options, errors = parse_options(
             cast(List[Custom], entries_by_type[Custom])
@@ -624,3 +634,16 @@ class FavaLedger:
             groups.setdefault(entry.__class__.__name__, []).append(entry)
 
         return sorted(list(groups.items()), key=itemgetter(0))
+
+    def commodity_name(self, commodity: str) -> Optional[str]:
+        """Return the 'name' field in metadata of a commodity
+        Args:
+            commodity: The commodity in string
+        Returns:
+            The 'name' field in metadata of a commodity if exists,
+            otherwise the input string is returned
+        """
+        commodity_ = self.commodities.get(commodity)
+        if commodity_:
+            return commodity_.meta.get("name")
+        return commodity
--- a/src/fava/json_api.py
+++ b/src/fava/json_api.py
@@ -119,7 +119,8 @@ def query_result():
     if contents:
         if "ERROR" in contents:
             raise FavaAPIException(contents)
-    table = table(contents, types, rows)
+    table = table(g.ledger, contents, types, rows)
+
     if types and g.ledger.charts.can_plot_query(types):
         return {
             "chart": g.ledger.charts.query(types, rows),
--- a/src/fava/templates/macros/_commodity_macros.html
+++ b/src/fava/templates/macros/_commodity_macros.html
@@ -1,0 +1,13 @@
+{% macro render_currency(ledger, currency) %}
+<span title="{{ ledger.commodity_name(currency) | default(currency) }}">{{ currency }}</span>
+{% endmacro %}
+
+{% macro render_amount(ledger, amount, class="num") %}
+{% if amount is none or amount.number is none %}
+  <span class="{{ class }}"><span>
+{% else %}
+  <span class="{{ class }}" title="{{ ledger.commodity_name(amount.currency) }}">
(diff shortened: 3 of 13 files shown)


yagebu:
Thanks for the PR and a happy new year to you too! I think having this metadata name available in Fava is a nice feature to have :) Ref #975, where this was already requested (for currencies).

However, I think just replacing the currency string with the name is not ideal as I'd expect those names to be rather long and screw up the layout. I think wrapping the currency in a `<span>` element with a title attribute to show on hovering would be preferable. Same applies to the account names IMHO.

mckelvin:
Hey @yagebu , I've updated the PR accordingly. Please review again.

src/fava/core/__init__.py
```diff
@@ -200,6 +202,9 @@ def __init__(self, path: str) -> None:
         #: A dict containing information about the accounts.
         self.accounts = AccountDict()
 
+        #: A dict containing information about the commodities
+        self.commodities: Dict[str, Commodity] = {}
```
yagebu, line 206: A test for this, i.e., that this property is correctly set, would be good.

src/fava/core/__init__.py
```diff
@@ -241,6 +246,10 @@ def load_file(self) -> None:
                 cast(Close, entry).account
             ).close_date = entry.date
 
+        for entry in entries_by_type[Commodity]:
```
yagebu, line 249: I guess we need to reset `self.commodities` to an empty dict above this line.

src/fava/templates/macros/_commodity_macros.html
```diff
@@ -0,0 +1,20 @@
+{% macro render_currency(currency) %}
```
yagebu, line 1: I think I'd prefer it if this (and the macro below) would take an explicit `ledger` first argument to make the required context a bit clearer (and allowing this file to be imported without context).

src/fava/templates/macros/_commodity_macros.html
```diff
@@ -0,0 +1,20 @@
...
+{% set alias = ledger.commodities.get(currency).meta.get('name') if ledger is defined else none %}
```
yagebu, line 2: The `if ledger is defined else none` part shouldn't really be necessary IMHO. If `ledger` is not defined, that sounds like an error we shouldn't silently swallow.

yagebu: Also, doesn't the `.meta.get('name')` part error for undefined currencies?

src/fava/templates/macros/_commodity_macros.html
```diff
@@ -0,0 +1,20 @@
...
+<span title="{{ alias | default(currency) }}">{{ currency }}</span>
+{% endmacro %}
+
+{% macro render_amount(amount, class="num") %}
+{% if amount is none or amount.number is none %}
+  <span class="{{ class }}"><span>
+{% else %}
+  {% set commodity = ledger.commodities.get(amount.currency) if ledger is defined else none %}
```
yagebu, line 10: It might make sense to extract this and and the following lines ("does this currency have a name") into a method on the FavaLedger class, say, `currency_name(cur: str) -> Optional[str]`. That would also make it more testable to ensure it works correctly in all different cases (no currency directive/currency directive without `name` metadata/currency directive with `name` metadata)

yagebu (changes requested):
Overall these changes look quite good, I've left some comments in the code.

This could definitely use some tests, not necessarily the HTML code but rather the logic behind it.

src/fava/templates/macros/_commodity_macros.html
mckelvin: The `ledger` variable has been used a lot in templates so I think it makes sense to [leave it as a global variable](https://github.com/beancount/fava/blob/a8c4f03c1a3b2a5c452858b71b84887782e91602/src/fava/application.py#L205). Otherwise I have to pass the variable as a parameter in many macro calls.

src/fava/json_api.py
```diff
@@ -114,12 +114,15 @@ def payee_accounts() -> List[str]:
 def query_result():
     """Render a query result to HTML."""
     query = request.args.get("query_string", "")
-    table = get_template_attribute("_query_table.html", "querytable")
```
mckelvin, line 117: [Global variables](https://github.com/beancount/fava/blob/a8c4f03c1a3b2a5c452858b71b84887782e91602/src/fava/application.py#L201) are not set when using `flask.get_template_attribute`, which causes a test failure in `test_api_query_result` , so I changed a bit here.



https://stackoverflow.com/questions/19501628/get-template-attribute-how-to-inject-template-globals/65758207#65758207

yagebu: As mentioned elsewhere, I'd prefer it to pass `ledger` as a macro argument, then we could just use Jinja2's cached template and keep the code here working like before.

yagebu (changes requested):
This could still do with some tests for the new functionality. If you want to stick with keeping the logic in the templates, then those should be tested. The easiest way to do that that I can see is passing `ledger` explicitly to all these macros and then adding some tests for their outputs

yagebu:
See #1222 for an example of how these macros can be tested.

tests/test_template_commodity_macros.py
```diff
@@ -0,0 +1,24 @@
...
+        macro = get_template_attribute(
+            COMMODITY_MACROS_PATH, "render_currency"
+        )
+        assert "US Dollar" in macro(example_ledger, "USD")
+
+
+def test_render_amount(app, example_ledger) -> None:
+    with app.test_request_context(""):
+        macro = get_template_attribute(COMMODITY_MACROS_PATH, "render_amount")
+        g.ledger = example_ledger
```
mckelvin, line 21: This is a bit dirty as some filters are depending on `g.ledger`. Is it OK to leave it "AS IS"?

src/fava/json_api.py
mckelvin: Sorry about the slow progress. The PR is update now.

tests/test_template_commodity_macros.py
yagebu: Using `app.preprocess_request()` here should do the trick. Will do so in the merge

yagebu (approved):
Thanks again :)

redstreet:
This is a very helpful feature!

Question for @mckelvin or @yagebu: for a column of commodities that's just a `str` (not an `Inventory` or an `Amount`), it is possible to somehow indicate to Fava to treat it as such and show the commodity name on hover?

PR #1208, opened 2021-01-04
Allow to specify the hooks in the import config

### LINKED ISSUE #1197: Allow to disable default importer hooks
It would be nice if it would be possible to disable the default importer hooks.

Unfortunately beancount uses the standard duplicate detector if you don't specify any hooks. This unfortunately marks entries as duplicates although they aren't.

I'm using the apply_hooks from smart importer, so I would not want any hooks from beancount to run.

tarioch:
Fixes #1197

### DIFF
--- a/src/fava/core/ingest.py
+++ b/src/fava/core/ingest.py
@@ -65,7 +65,7 @@ class IngestModule(FavaModule):
         super().__init__(ledger)
         self.config = []
         self.importers = {}
-        self.hooks = [extract.find_duplicate_entries]
+        self.hooks = []
         self.mtime = None
 
     @property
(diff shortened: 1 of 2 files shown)


src/fava/core/ingest.py
```diff
@@ -65,6 +65,7 @@ def __init__(self, ledger):
         super().__init__(ledger)
         self.config = []
         self.importers = {}
+        self.hooks = [extract.find_duplicate_entries]
```
yagebu, line 68: I think it would be better to initialise to an empty list here and and then set it to the default below.

src/fava/core/ingest.py
```diff
@@ -110,6 +111,8 @@ def load_file(self) -> None:
 
         self.mtime = os.stat(self.module_path).st_mtime_ns
         self.config = mod["CONFIG"]
+        if "HOOKS" in mod:
```
yagebu, line 114: Set the list to the default above this line, I think otherwise it would be incorrect after the first reload after removing `HOOKS` from an import config.

src/fava/core/ingest.py
```diff
@@ -110,6 +111,8 @@ def load_file(self) -> None:
...
+            self.hooks = mod["HOOKS"]
```
yagebu, line 115: Maybe add a basic check here that this is a list of callables and add an error message otherwise.

yagebu (changes requested):
Nice, that was quick :)

I've left some comments. Also, this should be documented on the import help page.

tarioch:
thanks a lot for your feedback, will incorporate that

tarioch:
Found something interesting when testing, even if I leave out the hooks, the duplicate detector is already triggered, will dig into beancount to see if that somehow already gets called in there.

tarioch:
Never mind, my mistake I still had my other hooks enabled

yagebu:
What's slightly annoying is that these "hooks" have a different signature than the ones in smart importer, so we should go into some more detail documenting that. Can do that in the merge.

yagebu (approved):

PR #1210, opened 2021-01-09
add extension hook after entry modification

xinluh:
There are hooks currently when new entry are inserted, file is saved, or metadata is added, but no hook when an entry is modified in the context popup.

### DIFF
--- a/src/fava/core/file.py
+++ b/src/fava/core/file.py
@@ -163,7 +163,14 @@ class FileModule(FavaModule):
         """
         with self.lock:
             entry = self.ledger.get_entry(entry_hash)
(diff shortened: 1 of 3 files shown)


yagebu:
Thanks for the PR, looks like a good addition :) This currently fails the CI, could you fix that?

xinluh:
@yagebu done

yagebu:
> done (debian build failure seems unrelated)

The debian failure is fixed in master so a merge or rebase would have fixed that as well.

yagebu:
I've merged this and clarified the types for the extension hooks (this one is called with the entry as a string, whereas the others are called with a Beancount entry)

PR #1213, opened 2021-01-10
Allow other packages to depend on fava's sources in git

johannesjh:
This pull request enables other packages to depend on fava by referencing fava's git repo.

Problem: When a project declares a dependency to fava's git repo, e.g., as in the following poetry project, then the installation of fava is broken because fava's frontend is not built as part of fava's setuptools build process.

```toml
[tool.poetry]
name = "fava-fe-build"
version = "0.1.0"

[tool.poetry.dependencies]
python = "^3.8"
fava = { git = "https://github.com/beancount/fava.git" }

[build-system]
requires = ["poetry-core>=1.0.0"]
build-backend = "poetry.core.masonry.api"
```

The problem can be experienced by running `poetry run fava example.beancount` in the above example project. As a result, fava's webserver starts successfully, but fava's GUI is broken because of missing javasscript and css.

Motivation: Other applications may want to depend on fava. For example, I am currently experimenting with a [fava-desktop](https://github.com/johannesjh/fava-desktop) application that depends on fava in order to wrap fava in a GTK application. It would be very convenient for development purposes if applications could depend on fava's sources in git. On a further note, this would also make it very easy to depend on forks and branches of fava. (Also note that alternative ways of depending on fava less convenient depending on the specific situation, e.g., having to wait for a new release of fava, or having to checkout and build fava in a vendor directory).

Suggested solution: I instructed setuptools to trigger the frontend build when building fava. 

Usage:

```bash
# building fava will first invoke the frontend build. 
# E.g., when adding a poetry dependency on fava's git repo as in the above example, 
# but also when manually running the setuptools build as usual:
python3 setup.py build

# the frontend build can also be invoked directly as follows:
python3 setup.py build_frontend
```


yagebu:
I don't think that I want to enable this way of depending on Fava. Having it run make on install just seems like opening another can of possible breakage if people don't have make, Node, npm, or some broken combination installed.

johannesjh:
The pull request does not change how the package is installed, it changes how the package is _built_. 

This is evident from how I registered the custom command in setup.py:
* `cmdclass={'install': MyCustomCommand}` would register a custom command to be run at package installation time, whereas
* `cmdclass={ "build_py": MyCustomCommand }` registers a custom command to be run at build time

yagebu:
Sorry, I meant "Having it run make on install..."

johannesjh:
The pull request does not change what tools are needed:

* Those who want to install fava can do this as usual, without make or npm
* Those who want to build fava can do this as usual, they will need make and npm, nothing new here.

johannesjh:
demo project that depends on fava via git
https://github.com/johannesjh/fava-fe-build

yagebu:
> The pull request does not change what tools are needed:
> 
>    * Those who want to install fava can do this as usual, without make or npm
> 
>    * Those who want to build fava can do this as usual, they will need make and npm, nothing new here.

I understand the change that this pull request provides.

However, as stated above, I don't want to enable other packages to depend on a development version of Fava. Because that leads to an installation of that package requiring Fava being built, which suddenly brings in the known dependencies.

johannesjh:
alright, let's agree to disagree. feel free to close this issue.

PR #1214, opened 2021-01-10
add invert-negative-accounts option to invert balances of Income, Liabilities and Equity accounts

andreasgerstmayr:
This PR adds a new option `invert-negative-accounts` which flips the signs of the Income, Liabilities and Equity accounts in the income statement (including the net profit and income chart) and balance sheet.

I didn't touch the chart or table at the individual account pages (`/beancount/account/Income/changes/` for example) or the journal, as I think it may get really confusing to deviate from the internal beancount accounting here. I was mainly concerned with the "Net Profit" chart of the income statement - it bothered me to see a negative net profit when in fact it should be positive.

It was a breeze to add, thanks for this great software and clean coding style! 👍 

Resolves: #448
Based on the work of @aumayr in 0d4681ce8e2ae2e0d532e445173cb30bb81bf518


andreasgerstmayr:
I don't know if 'negative accounts' is the correct term - is there a better term to describe these accounts, which have a negative balance in general (Income, Liabilities and Equity accounts)?

yagebu:
Hey, thanks for this PR :)

Could you add documentation for this option to the help pages? In particular the fact that it does not apply equally to all reports should be well-documented.

> I didn't touch the chart or table at the individual account pages (/beancount/account/Income/changes/ for example) or the journal, as I think it may get really confusing to deviate from the internal beancount accounting here.

I don't have a strong opinion on this and if potential users of this option agree, I think that's a solid choice from a maintenance perspective. Having it clearly constrained to a few select reports is easier than having to get all :)

Regarding the option name: `invert-negative-accounts` sounds like it would invert all negative accounts, but it does not and might actually invert an account with a positive balance. I browsed through the issues but that didn't turn up any brilliant option names either. Maybe just be explicit and call it `invert-income-liabilities-equity`? Not a much longer option name and easier to understand.

andreasgerstmayr:
Thanks for the review!

I've documented the option in the help page, and updated the option name.

yagebu (changes requested):
This looks good overall :)

It could to with some tests, at least for the functions in `fava.core.charts` and `fava.template_filters`. We don't have a test for the `render_diff_and_number` yet, but it might be a good time to add one (using the `snapshot` test fixture maybe, so you don't have to type out the generated HTML)

andreasgerstmayr:
I've squashed and rebased the commits, and added three tests for the functions/macros. Can you please take a look again?

yagebu (approved):
Thanks again!

sqybi:
I've wanted this feature for such a long time! Thanks for @andreasgerstmayr 's PR and @yagebu 's review (and continuous development for all the features :D ), can't wait for the new release!

cataclysmic:
From me, too. Thanks so much. This makes it so much more intuitive when reading. I love fava so much.

PR #1215, opened 2021-01-12
Use parent font-family in chart axes

ksesong:
This style adds a `font-family` property to `text` elements in `svg`s, so that axes use the same typeface as the surrounding body text (instead of the default `sans-serif`).


yagebu:
Cool, thanks for the PR :)

PR #1218, opened 2021-01-14
fix editor z-index position

ksesong:
The editor is positioned on top of floating components, which is especially visible on help pages, and on the query or editor page when the header overlays (file selector, filters...) are active. This style sets the editor `z-index` to `0`.

image

### DIFF
--- a/frontend/css/editor.css
+++ b/frontend/css/editor.css
@@ -10,7 +10,7 @@
 }
 
 .CodeMirror.CodeMirror {
-  z-index: 0;
+  z-index: var(--z-index-editor);
   margin-bottom: 1em;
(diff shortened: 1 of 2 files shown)


yagebu (changes requested):
Thanks for this fix!

CSS variables are used in all the other places where we set a `z-index`, which makes it easier to ensure they're correctly ranked. Could you also set and use a CSS variable for this value?

ksesong:
I named it as `--z-index-editor` if we want to stay specific to the editor.

yagebu (approved):

yagebu:
Thanks again :)

PR #1220, opened 2021-01-17
Small CSS changes

aumayr:
- Document Upload modal: More spacing between `fieldset`s and a longer filename `input`
- Journal: Avoid long numbers and their currencies to break
- Details (Context) modal: Use default browser arrow (and avoid CSS bug in Safari)

### DIFF
--- a/frontend/css/components.css
+++ b/frontend/css/components.css
@@ -10,20 +10,9 @@ details.inactive {
 }
 
 details summary {
-  display: flex;
-  align-items: center;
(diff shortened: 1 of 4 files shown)


yagebu (changes requested):
Long time no see :) 

No complaints about the first two changes. The changed CSS for <details> looks good in the context modal but doesn't seem to work well on the query page, could you try to fix that?

yagebu (approved):

Issue #1224, opened 2021-01-26
Import page shows files that can't be imported


mattr555:
I'm someone who tends to hoard a lot of files in my `~/Downloads` folder, which is where I download my monthly statements, and is the import directory for `bean-file`/`fava`. I'd prefer to only see files that have an associated importer on fava's import page (though I understand the value in having other files show up in the list if you're using a different directory to collect documents before they're filed).

I'm happy to contribute a PR that adds an option to enable this. Would this be a sensible default to enable, or should we show all files by default? Don't want to break anyone's workflow. Thanks!

edit: as an alternative, we could explore adding a toggle to do this filtering on the frontend. The only quibble I would have with that is that we get a performance win by not sending down files that the user doesn't care about.

yagebu:
This is not really something that I'd want to add an option for but doing this filtering in the frontend sounds like a worthwhile solution. Maybe we could display the list in two parts to allow those that don't have an importer to be more easily ignored and just toggled in the frontend.

> The only quibble I would have with that is that we get a performance win by not sending down files that the user doesn't care about.

The size of the data loaded on that page is really quite negligible.

mattr555:
> Maybe we could display the list in two parts to allow those that don't have an importer to be more easily ignored and just toggled in the frontend.

I like this idea! Hope to have a PR out implementing it this weekend.

> The size of the data loaded on that page is really quite negligible.

Fair enough. I was noticing loading times of a couple seconds before I purged some of my old downloads, but nothing extreme.

yagebu:
> I like this idea! Hope to have a PR out implementing it this weekend.

Cool :)

> I was noticing loading times of a couple seconds before I purged some of my old downloads, but nothing extreme.

All files in the import directory are checked by the `identify` part of Beancount's ingest mechanism, depending on your importers that might load and read all files (e.g., to check for the existence of some identifier in a CSV file). That might take some time if you have a lot of files in that directory of course and can't really be avoided, transmitting the result will only be a fraction of those seconds

PR #1228, opened 2021-01-30
Display files with importers at top of import page

mattr555:
Fixing #1224 by breaking the list of files into two: one of the files that have importers, and one of the other files found in the import directories. This second list is hidden behind a `<details>` tag that is automatically expanded if there are no files in the first list. Here's an example of what the page would look like on first load if there are importable files available:
Screen Shot 2021-01-30 at 2 52 44 PM

This is accomplished by extracting a common `FileList` component. This is my first time using Svelte, so apologies if it's not quite idiomatic on the first cut. Thanks!

### DIFF
--- a/frontend/src/import/FileList.svelte
+++ b/frontend/src/import/FileList.svelte
@@ -1,0 +1,106 @@
+<script>
+  import { createEventDispatcher } from "svelte";
+  import { _ } from "../i18n";
+
+  import AccountInput from "../entry-forms/AccountInput.svelte";
+
+  /** @type {import("./helpers").ProcessedImportableFiles} */
+  export let files;
+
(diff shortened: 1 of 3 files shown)


yagebu (changes requested):
Thanks for the PR, this looks pretty good :)

I think this could be simplified a bit (and offer better typechecking) if you just pass down the functions like `move` down to the component instead of using event listeners.

mattr555:
Thanks! Makes more sense :)

P.S. I don't think the failing checks are my fault?

yagebu (approved):
Thanks again, I'll make some minor changes and merge this :)

> P.S. I don't think the failing checks are my fault?

Nope, I guess you branched off master a while ago, when there was some breakage in CI

PR #1235, opened 2021-02-20
Update usage.rst

prabusw:
added instruction to update an existing fava installation to the latest version..


yagebu:
Thanks for this PR. I think passing the `-U` option in general is a good idea, I'll update the install instruction to include this :)

PR #1238, opened 2021-02-27
Add dark mode

### LINKED ISSUE #1084: Dark Mode
With most browsers adding the ability to have a light and dark mode. How hard would it be to add this to fava? is it just a CSS change? 

Would be nice to have  this feature

arkn98:
Resolves #1084 

I've only made a few changes so that I can create a draft PR and start a discussion. I'm still unsure of what color palette to use, where to place the theme switch option, and other relevant information. For starters, I used the colors from GitHub's dark theme, since most of us are already accustomed to it. As already mentioned in #1084, I'm using `prefers-color-scheme` to detect the theme preference of the user (you can use Dev Tools from the browser to simulate the theme preference for now). You can see a sample screenshot below.

Discussion & feedback very much welcome!

**Note:** I've never contributed to open source repositories before and also, I'm somewhat a beginner. So, please feel free to point out if I do something wrong.

fava-dark-mode-initial-scrot-1

### DIFF
--- a/frontend/css/base.css
+++ b/frontend/css/base.css
@@ -36,7 +36,7 @@ pre {
   font-family: var(--font-family-monospaced);
   white-space: pre;
   background-color: var(--color-code-background);
-  border: 1px solid var(--color-background-darker);
+  border: 1px solid var(--color-code-border);
   border-radius: 3px;
 }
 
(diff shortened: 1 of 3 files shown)


arkn98:
@yagebu Did you get a chance to go through this?

yagebu:
Thanks @arkn98 for the pull request and congratulations on your first open-source contribution :)

I think this color theme could reuse that colors that Fava normally uses a bit more (and if those are too light/dark, see if only readjusting the lightness in the hsl() color specs is enough). I'm not a user of dark mode myself, so I don't have a strong opinion on the exact looks of dark mode, but if we add one to Fava it should like Fava, as we could otherwise just recommend users to use some browser extension to enable dark mode.

yagebu:
Closing this stale PR. Feel free to reopen if you want to continue to work on this :)

Issue #1243, opened 2021-04-04
https scheme not respected in `router.navigate`


xinluh:
This is a problem if fava is hosted with SSL. 

There are a few places (account selector, for example) where `router.navigate(urlFor(...))` is used navigate to a different page, but somewhere in the navigation the url scheme `https://` is not respected, and the navigation fails in modern browser. In Chrome, here is the error message in console:

```
Mixed Content: The page at 'https://<somehost>/beancount/trial_balance/' was loaded over HTTPS, but requested an insecure resource 'http://<somehost>/beancount/account/Income:Income?partial=true'. This request has been blocked; the content must be served over HTTPS.
```

Other sidebar links works. I tried to trace where the url scheme got overwritten in the `router.navigate(urlFor(...))`  call but couldn't locate the problem...

yagebu:
As you have found out in #1245, this is an issue with your deployment setup.

PR #1245, opened 2021-04-06
Fix trailing '/' in url navigation

xinluh:
This fixes the problem in #1243 

I think what happens is
- The Flask route are like `account/<name>/` with a trailing slash
- If we request an url without trailing slash, Flask ends up doing a redirect
- I happen to have a weird deployment setup (Google App Engine with IAP) that has some trouble with making sure that https is reflected through all the redirection
- Page failed to load because browser see that a `https` request is returning back at `http` location

The redirection weirdness isn't fava's fault, but it makes sense that we don't need the redirect in the first place if we request the right URL with trailing slashes.


yagebu:
Thanks :)

PR #1246, opened 2021-04-06
make the conversion selector component available for use in plugin

xinluh:
The conversion selector can be helpful in plugin, especially if the plugin uses the querytable 
image

Also fix a bug introduced in #1207 where query table crashes when displaying the "Amount" type


yagebu:
Cool, thanks :)

PR #1248, opened 2021-04-06
fix browser back button not updating filter input

xinluh:
This is a buggy behavior currently. To reproduce:
- Go to, say, an account page
- Add a filter in any of the filter input boxes
- Page updates and filters correctly
- Use the "back" button in browser
- Page updates and filters correctly, but the filter boxes still have the filter string even though the page content doesn't reflect the filter string any more

This PR fixes this by making sure that on `popstate` event, the state in the URL is written back to the store which updates the filter input boxes correctly.


yagebu:
Thanks for this fix :)

yagebu (approved):

Issue #1181, opened 2020-11-17
Show the current version of Fava in the help pages.


grostim:
I think the title is self-explanatory.
A link to the changelog would be nice too. 
As well as the version number of beancount.

yagebu:
I think this could be part of the index page of the included help pages. A PR would be welcome.

aclindsa:
@yagebu @grostim I took a first pass at this in https://github.com/beancount/fava/pull/1250. After my change, the help index page looks like this with my local install of fava:

![2021-04-09-222613_1172x569_scrot](https://user-images.githubusercontent.com/362852/114255653-a2bc9c00-9982-11eb-9e2f-817cda7c8f87.png)

grostim:
It looks simple and straightforward. I like it.
I will give it a try over this week end.

Le sam. 10 avr. 2021 à 04:28, Aaron Lindsay ***@***.***> a
écrit :

PR #1250, opened 2021-04-10
Print fava version on help index page

aclindsa:
Closes https://github.com/beancount/fava/issues/1181

### DIFF
--- a/src/fava/help/_index.md
+++ b/src/fava/help/_index.md
@@ -1,4 +1,7 @@
-Welcome to the help pages for Fava! There are pages for the following topics:
+Welcome to the help pages for Fava! You are running Beancount version
+`{{ beancount_version }}` and Fava `{{ fava_version }}`
+([changelog](https://github.com/beancount/fava/blob/master/CHANGES)). There are
+help pages for the following topics:
 
 -   [Beancount Syntax](./beancount_syntax) - short overview of the syntax.
 -   [Budgets](./budgets) - how to use Fava's budgeting feature.
--- a/src/fava/application.py
+++ b/src/fava/application.py
@@ -23,6 +23,7 @@ from typing import Optional
 import flask
 import markdown2  # type: ignore
 import werkzeug.urls
+from beancount import __version__ as beancount_version
 from beancount.core.account import ACCOUNT_RE
 from beancount.utils.text_utils import replace_numbers  # type: ignore
 from flask import abort
@@ -36,6 +37,7 @@ from flask_babel import Babel  # type: ignore
 from flask_babel import get_translations
 from werkzeug.utils import secure_filename
 
+from fava import __version__ as fava_version
 from fava import LANGUAGES
 from fava import template_filters
 from fava.context import g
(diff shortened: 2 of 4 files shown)


aclindsa:
I am confused by the CI failures here. It says:
> src/fava/application.py:26: error: Module 'beancount' has no attribute '__version__'

But on my local Ubuntu, importing beancount's version works just fine:

```
root@c679fc96b91c:/tmp/build/fava_investor# cat /etc/lsb-release
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=20.04
DISTRIB_CODENAME=focal
DISTRIB_DESCRIPTION="Ubuntu 20.04.2 LTS"
root@c679fc96b91c:/tmp/build/fava_investor# python3
Python 3.8.5 (default, Jan 27 2021, 15:41:15)
[GCC 9.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from beancount import __version__ as beancount_version
>>> print(beancount_version)
2.3.4
```

@yagebu Is there something special going on with tox/mypy's environment or something that I'm failing to comprehend?

aclindsa:
> > src/fava/application.py:26: error: Module 'beancount' has no attribute '**version**'

To be clear, this was originally formatted as `__version__` in the error message.

aclindsa:
> @yagebu Is there something special going on with tox/mypy's environment or something that I'm failing to comprehend?

I'm still not sure I understand *why*, but importing top-level `beancount` instead of trying to selectively import `beancount.__version__` allows the 'mypy' check to succeed. I've updated this PR to use that instead for now, pending your input.

src/fava/help/_index.md
```diff
@@ -1,4 +1,7 @@
-Welcome to the help pages for Fava! There are pages for the following topics:
+Welcome to the help pages for Fava! You are running beancount version
```
yagebu, line 1: Beancount

src/fava/help/_index.md
```diff
@@ -1,4 +1,7 @@
...
+`{{ beancount_version }}` and fava `{{ fava_version }}`
```
yagebu, line 2: Fava

yagebu (changes requested):
Thanks for the PR! :)

The types for Beancount are not coming from Beancount, but rather the stub type files in the stubs directory. If you add the version to `stubs/beancount/init.pyi`, the import should work directly.

Since this adds some dynamic content to the help pages, it would be good to add a test that the help index page renders without errors and that it inserts the version numbers.

aclindsa:
> The types for Beancount are not coming from Beancount, but rather the stub type files in the stubs directory. If you add the version to `stubs/beancount/init.pyi`, the import should work directly.

Ah, doing that does allow me to use the more selective import. Thanks for the tip!

> Since this adds some dynamic content to the help pages, it would be good to add a test that the help index page renders without errors and that it inserts the version numbers.

Done! I also fixed the capitalization on 'Fava' and 'Beancount' you noted.

yagebu (approved):
Thanks again, this looks great :)

PR #1258, opened 2021-04-21
Sort all fields containing value

### LINKED ISSUE #1256: Sorting querytable not sorting by value but by string representation
Hi all!

- Noticed first in `holdings/by_currency/` (How to reproduce):
- (version `fava, version 1.19.dev83+g3ecb83a7.d20210420`)
- Click on `market_value` table header
- Then table will be sorted, though not by value but by string representation (see following example)

|...|market_value `(sorted by, ASC)`|
|---|---|
|...|100.00 EUR|
|...|12.00 EUR|
|...|2.10 EUR|
|...|2.90 EUR|
|...|3000.00 EUR|

Had a brief look in the code already: bean-query is fine: With `ORDER BY market_value DESC` in the query the values in the resulting table are sorted just fine.

So I guess querytable sorting does not work properly, because the value in the table is of type string due to the `EUR` appendix instead of int.

So the question is, how can we make the values in the table sort correctly?

- Either we could add / change the `ORDER BY` clause in the query on click on the table header (difficult integration?)
- Or we look into the use of querytable and see how this can be done
  - Encountered in `src/fava/templates/holdings.html`
  - `src/fava/templates/_query_table.html` could be a great spot to look at next

I would be happy if you would have any more remarks and ideas to this. Looking forward to working on a solution with you as I really like fava.

Thanks in advance and kind regards,

Luca

**Update**:

- Table sorting is made possible by `<table is="sortable-table" class="queryresults">`
- `sortable-table` is defined in `frontend/src/main.ts` to use class `SortableTable`
- call to `sortElements` in `frontend/src/sort.ts` which applies sorting by `getValue` function result
- We should have `data-sort` attribute `num` on table header -> `numComparator`should be used. Does parsing not work?

LucaBernstein:
Closes #1256

See linked issue. Tried to make approach as generic as possible, thus checking whether `_value` is included in field name.

A chain like `{{ sort_type[type|string] or sort_type[name|string] or 'string' }}` would also be possible, but requires explicit definition of field names that should be converted to `num` to be defined in `sort_type` dict.

### DIFF
--- a/src/fava/templates/_query_table.html
+++ b/src/fava/templates/_query_table.html
@@ -55,7 +55,7 @@
   <thead>
     <tr>
       {% for name, type in types %}
-      <th data-sort="{{ sort_type[type|string] or "string" }}">{{ name }}</th>
+      <th data-sort="{% if '_value' in name %}num{% else %}{{ sort_type[type|string] or 'string' }}{% endif %}">{{ name }}</th>
       {% endfor %}
     </tr>
   </thead>


yagebu:
Thanks for the PR, but I'm not planning to merge this change. As explained in the issue, we can't really sort inventories and I don't want to add this special case by name.

PR #1260, opened 2021-04-28
Document preview of images

orfest:
Use a simple `` tag to preview image documents. Tested locally by creating simple images in `gimp` and `inkscape`.


yagebu:
Thanks! Merged with some minor changes :)

PR #1265, opened 2021-05-12
Use steps instead of linear interpolation in line and bar charts

jeffwheeler:
Currently, line and bar charts in Fava can be confusing because values are linearly interpolated between each sample. This makes many of the charts difficult for me to read, as the line location is wrong on most dates.

With this pull request, I am suggesting to change from this:
Screen Shot 2021-05-11 at 10 20 57 PM

To this:
Screen Shot 2021-05-11 at 10 21 20 PM

I appreciate your consideration of this pull request. :)

Jeff


yagebu:
I like it, thanks for the PR :)

jeffwheeler:
Thanks!

PR #1267, opened 2021-05-12
Pin Flask, Jinja2 major versions

### LINKED ISSUE #1266: Flask, Jinja major releases break Fava
~12 hours ago, version [2.0.0 of Flask](https://github.com/pallets/flask/releases/tag/2.0.0) and [3.0.0 of Jinja](https://github.com/pallets/jinja/releases/tag/3.0.0) were released, with which Fava does not appear to be compatible:

```
$ virtualenv --python python3 venv
$ ./venv/bin/pip install fava
$ ./venv/bin/fava ledger.beancount
  File "/home/foo/./venv/bin/fava", line 5, in <module>
    from fava.cli import main
  File "/home/foo/venv/lib/python3.9/site-packages/fava/cli.py", line 11, in <module>
    from fava.application import app
  File "/home/foo/venv/lib/python3.9/site-packages/fava/application.py", line 67, in <module>
    app.jinja_options["extensions"].append("jinja2.ext.do")
KeyError: 'extensions'
$ ./venv/bin/pip freeze | grep -i fava
fava==1.18
$ ./venv/bin/pip freeze | grep -i flask
Flask==2.0.0
$ ./venv/bin/pip freeze | grep -i jinja
Jinja2==3.0.0
```

I suggest just pinning both versions as strictly-less-than the new major releases until an upgrade can be coordinated. I'll open a PR shortly.

patricklucas:
Fava is currently incompatible with recent major releases of Flask
(2.0.0) and Jinja2 (3.0.0), crashing with an error like:

```
  File "/home/foo/./venv/bin/fava", line 5, in <module>
    from fava.cli import main
  File "/home/foo/venv/lib/python3.9/site-packages/fava/cli.py", line 11, in <module>
    from fava.application import app
  File "/home/foo/venv/lib/python3.9/site-packages/fava/application.py", line 67, in <module>
    app.jinja_options["extensions"].append("jinja2.ext.do")
KeyError: 'extensions'
```

This change pins Flask and Jinja2 to the previous major versions until
upgrading can be coordinated.

Closes #1266

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -43,14 +43,15 @@ install_requires =
     Babel>=2.6.0
     cheroot
     Flask-Babel>=1.0.0
-    Flask>=0.10.1,<2
-    Jinja2>=2.10,<3
+    Flask>=0.10.1,<2 # See #1266
+    Jinja2>=2.10,<3 # See #1266
     beancount>=2.3.0
     click
     markdown2>=2.3.0
     ply
     simplejson>=3.2.0
-    Werkzeug>=0.15.0
+    Werkzeug>=0.15.0,<2 # See #1266
+    MarkupSafe<2 # See #1266
 
 [options.entry_points]
 console_scripts =
--- a/setup.cfg
+++ b/setup.cfg
@@ -43,8 +43,8 @@ install_requires =
     Babel>=2.6.0
     cheroot
     Flask-Babel>=1.0.0
-    Flask>=0.10.1
-    Jinja2>=2.10
+    Flask>=0.10.1,<2
+    Jinja2>=2.10,<3
     beancount>=2.3.0
     click
     markdown2>=2.3.0


patricklucas:
The test failures are due to [Werkzeug](https://github.com/pallets/werkzeug), part of Flask, also being upgraded to the new major version, even though its version is not set in Fava's setup.cfg.

I'll add a similar version restriction to Werkzeug and [MarkupSafe](https://github.com/pallets/markupsafe) (yet another affected dependency) with a reference back to #1266.

aronsky:
Interesting. I ran into this issue today when building a docker container with fava, and running with `pip install Jinja==2.11.3 fava` fixed it for me. So it seems to me that pinning just Jinja would be enough, as long as it's installed first...

yagebu:
Thanks for the PR, but I'll fix this in #1268 instead :)

PR #1269, opened 2021-05-14
Ignore file extension case for document preview

jeffwheeler:
Previously, .CSV files were not rendered while .csv files were. This patch removes the case-sensitivity, so both .csv and .CSV files are rendered as expected.


yagebu:
Nice, thanks :)

PR #1274, opened 2021-05-16
Support rendering qfx documents in editor

jeffwheeler:
`.qfx` files are plain-text, and therefore can be viewed in the document editor just like `.csv` and `.txt`.

### DIFF
--- a/frontend/src/documents/DocumentPreview.svelte
+++ b/frontend/src/documents/DocumentPreview.svelte
@@ -1,4 +1,5 @@
 <script lang="ts" context="module">
+  const plainTextExtensions = ["csv", "txt", "qfx"];
   const imageExtensions = [
     "gif",
     "jpg",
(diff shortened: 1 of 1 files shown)


frontend/src/documents/DocumentPreview.svelte
```diff
@@ -25,7 +25,7 @@
 
 {#if extension === "pdf"}
   <object title={filename} data={url} />
-{:else if ["csv", "txt"].includes(extension)}
+{:else if ["csv", "txt", "qfx"].includes(extension)}
```
yagebu, line 28: Now that this list gets longer, I think I would prefer to move it to the top as a `plainTextExtensions` next to `imageExtensions`

yagebu (changes requested):
Thanks for the PR, just left a minor comment.

frontend/src/documents/DocumentPreview.svelte
```diff
@@ -1,4 +1,5 @@
 <script lang="ts" context="module">
+  const plainTextExtensions = ["csv", "txt", "qfx"];
```
jeffwheeler, line 2: Note that `prettier` prefers this all on one line, instead of broken into multiple lines like `imageExtensions`.

yagebu (approved):
Thanks :)

PR #1275, opened 2021-05-16
Support for <Enter> and <Escape> in rename dialog

jeffwheeler:
This commit adds support for `<Enter>` to submit the rename modal dialog, and for `<Escape>` to exit it.

### DIFF
--- a/frontend/src/AutocompleteInput.svelte
+++ b/frontend/src/AutocompleteInput.svelte
@@ -74,6 +74,9 @@
     } else if (event.key === " " && event.ctrlKey) {
       hidden = false;
     } else if (event.key === "Escape") {
+      if (!hidden && filteredSuggestions.length > 0) {
+        event.stopPropagation();
+      }
       hidden = true;
     } else if (event.key === "ArrowUp") {
       event.preventDefault();
--- a/frontend/src/documents/Documents.svelte
+++ b/frontend/src/documents/Documents.svelte
@@ -29,7 +29,7 @@
    * Rename the selected document with <F2>.
    */
   function keyup(ev: KeyboardEvent) {
-    if (ev.key === "F2" && selected) {
+    if (ev.key === "F2" && selected && !moving) {
       moving = { ...selected, newName: basename(selected.filename) };
     }
   }
@@ -53,15 +53,15 @@
       moving = null;
     }}
   >
-    <div>
+    <form on:submit|preventDefault={move}>
       <h3>{_("Move or rename document")}</h3>
       <p><code>{moving.filename}</code></p>
       <p>
         <AccountInput bind:value={moving.account} />
         <input size={40} bind:value={moving.newName} />
-        <button type="button" on:click={move}>{"Move"}</button>
+        <button type="submit">{"Move"}</button>
       </p>
-    </div>
+    </form>
   </ModalBase>
 {/if}
 <div class="fixed-fullsize-container">
--- a/frontend/src/modals/ModalBase.svelte
+++ b/frontend/src/modals/ModalBase.svelte
@@ -25,6 +25,9 @@
           ev.preventDefault();
           attemptFocus(first);
         }
+      } else if (ev.key === "Escape") {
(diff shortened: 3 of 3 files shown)


frontend/src/documents/Documents.svelte
```diff
@@ -43,6 +34,20 @@
       router.reload();
     }
   }
+
+  /**
+   * Rename the selected document with <F2>.
+   * Submit with <Enter> and exit with <Escape>.
+   */
+  function keyup(ev: KeyboardEvent) {
```
jeffwheeler, line 42: **Note:** This function was moved to satisfy the linter, as `move()` must be defined above this function now.

frontend/src/documents/Documents.svelte
```diff
@@ -43,6 +34,20 @@
...
+    if (ev.key === "F2" && selected) {
+      moving = { ...selected, newName: basename(selected.filename) };
+    } else if (ev.key === "Enter" && moving) {
+      move();
+    } else if (ev.key === "Escape" && moving) {
```
yagebu, line 47: I think this should be solved slightly differently - the `AutocompleteInput.svelte` component that is used for the account input also has an event listener for `Escape` (and `Enter` too actually). So currently one can hit `Escape` in the account input to hide the list of suggestions. With this change, it would also close the dialog.

I think this handling of the Escape key could be done in the `ModalBase.svelte` component - then it would also benefit the "add transaction" dialog. There's already a keyboard shortcut handler in ModalBase for Escape, however that one is (like all the other "standard" keyboard shortcuts in Fava) not activated if you're currently in an input element or other editable element.

The following two changes would probably necessary: 1) add the keyboard handler in ModalBase 2) adjust the AutocompleteInput component to call `event.stopPropagation` if the suggestion list is hidden with the Escape key so that we do not inadvertently close both the list and the modal.

yagebu (changes requested):
Thanks for the PR! This is a good change of behaviour, however I think the implementation needs to be done a bit differently, see my comments.

frontend/src/documents/Documents.svelte
```diff
@@ -40,12 +40,8 @@
    * Submit with <Enter> and exit with <Escape>.
    */
   function keyup(ev: KeyboardEvent) {
-    if (ev.key === "F2" && selected) {
+    if (ev.key === "F2" && selected && !moving) {
```
jeffwheeler, line 43: I noticed that it's possible to nest the dialogs. This addition fixes that.

frontend/src/modals/ModalBase.svelte
```diff
@@ -25,6 +25,9 @@
           ev.preventDefault();
           attemptFocus(first);
         }
+      } else if (ev.key === "Escape") {
```
jeffwheeler, line 28: I'd like your feedback on the placement of this code. It's within the `keydown` handler defined within `handleFocus`. I'm not sure if that makes sense or not.

frontend/src/AutocompleteInput.svelte
```diff
@@ -74,6 +74,9 @@
     } else if (event.key === " " && event.ctrlKey) {
       hidden = false;
     } else if (event.key === "Escape") {
+      if (!hidden && filteredSuggestions.length > 0) {
```
jeffwheeler, line 77: Checking `!hidden` is insufficient, as the suggestion list is also hidden when there are no suggestions.

jeffwheeler:
Thanks for the helpful feedback, @yagebu. I've updated according to your suggestions. I've left a few comments on the change for you.

frontend/src/modals/ModalBase.svelte
yagebu: It's vaguely related to focus management, so that's fine IMHO.

yagebu:
@jeffwheeler: Can you squash the commits and rebase this PR?

jeffwheeler:
Done

yagebu (approved):
Thanks again :)

Issue #1264, opened 2021-05-12
Regression in master: Ctrl-S transaction editor save functionality


bayesianmind:
Hey @yagebu Thanks for your work on Fava!
I noticed that Ctrl-S is broken in the Transaction Editor at origin/master and bisected it to commit be308c59. Previous commit 97586909 is fine.

Repro: open journal, edit a transaction. Try to press ctrl-S to save it.
Expected behavior: Submit button pressed and transaction gets updated.
Actual behavior: key isn't bound - browser opens save dialog.

I see bindings in be308c59 for ctrl + enter on the BQL editor but it doesn't work on the transaction editor.

bayesianmind:
Thank you yagebu for fixing it!

PR #1280, opened 2021-05-20
Fix Ctrl-S keybinding in #1264

bayesianmind:
The previous fix wasn't applied to SliceEditor. I copied @yagebu's fix to SliceEditor too and it fixed the issue for me.

### DIFF
--- a/frontend/src/editor/SliceEditor.svelte
+++ b/frontend/src/editor/SliceEditor.svelte
@@ -16,7 +16,16 @@
     changed = true;
   };
 
-  const [editor, useEditor] = initBeancountEditor(slice, onDocChanges, []);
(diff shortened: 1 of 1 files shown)


bayesianmind:
Oops sorry I was working off of origin/master instead of origin/main. Perhaps consider deleting origin/master to prevent confusion?

yagebu:
> Perhaps consider deleting origin/master to prevent confusion?

@bayesianmind: It is already deleted (since right after the move to `main`). Note that git doesn't delete removed remote branches by default, you'll need to fetch with `--prune` or set the `fetch.prune` git option.

bayesianmind:
Ah, good to know, thanks!

Issue #490, opened 2017-02-01
Overlay totals on graphs


ngoonee:
Currently mouseover on graph columns brings up a tooltip showing totals.

Is it possible to overlay such totals on the graph? Currently the only way to view that information (without looking at the tables below) on mobile devices is to press on the graph. Which would be fine except that immediately loads a different page, so it's not so useful for comparing month to month (for example).

aumayr:
IMHO clicking the charts should go to the link on the second click, as does on Github for example. This way, the first click (or better: tap) would show the overlay/tooltip, and the second one follow the link.

ngoonee:
That would be fine if possible.

xentac:
I agree that this would be way helpful.  In the meantime, if you drag on the graph near where you want to see, the popup should show up.  At least it does in Chrome when I use it on Android.

yagebu:
These [Google Material design guidelines](https://material.io/guidelines/components/tooltips.html) suggest that a tooltip should show on "tap and hold" on mobile. That's also what I usually do on mobile when on a desktop I would hover. This philosophy seems to work well for buttons (where clicking is the primary behaviour) but doesn't seem to apply so well to most areas of our charts where hovering for the tooltip is the more frequent behaviour. I still think that for areas in the charts that are primarily clickable we could with the guidelines.

It seems nigh-impossible and messy to do stuff like distinguishing between taps/clicks/double-taps/etc in JS. So maybe we can individually adjust the charts to be better suited for touch interfaces. There's only three charts responding to clicks AFAIK:

- Bar Charts: Actually on my laptop as well, I've found these to be a bit too clicky, I've more often than I wanted ended up clicking a bar. I think we could move the "click to time filter" to only occur when actually clicking on the time axis. This would allow tapping on the rest of the chart to show the tooltip.
- Treemap Charts: Does anyone use/need the click-to-zoom behaviour? (I never use it) We could just remove it and then one could tap most of the area to show the tooltip (as only the account names are links).
- Sunburst Charts: I think for these requiring tap-and-hold to show the account balance is acceptable.

> IMHO clicking the charts should go to the link on the second click, as does on Github for example. 

@aumayr: Which chart behaves likes this? I tried the "Network" chart and there a single click takes me straight to the commit. I didn't find links in any of the other charts under the "Graphs" tab.

PR #1281, opened 2021-05-22
Fix minor bugs in treemap charts

jeffwheeler:
I am fixing two bugs in the treemap viewer with this commit.

The first bug is with tooltip rendering. The existing code uses the `opacity` property to hide text when it is too small to fit, but because the text is still present (just invisible to the user), the tooltip still renders over the entire invisible area of the text. Instead, it should render based on the rectangle region, not the invisible text. The fix for this is simple, change `opacity` to `visibility`.

Screen Shot 2021-05-22 at 3 52 37 PM

In the above image, the tooltip is showing the text for the thin rectangle on the right, which is an error.

The second bug is caused by the use of strokes on each rectangle. The total height of the columns sometimes varies, so that one column will be an extra pixel taller, as seen on the following image:

Screen Shot 2021-05-22 at 3 52 49 PM

To resolve this, I use `paddingInternal` to create padding, instead of a stroke. I guess the variable `--color-treemap-text` was originally meant for a different purpose (`text`), but was now being used for the white stroke color. I've removed it.

### DIFF
--- a/frontend/src/charts/Treemap.svelte
+++ b/frontend/src/charts/Treemap.svelte
@@ -17,7 +17,7 @@
 
   $: height = Math.min(width / 2.5, 400);
 
-  const tree = treemap<AccountHierarchyDatum>().paddingInner(1);
+  const tree = treemap<AccountHierarchyDatum>().paddingInner(2).round(true);
   $: root = tree.size([width, height])(data);
   $: leaves = root.leaves().filter((d) => d.value);
 
@@ -38,14 +38,14 @@
     )})<em>${d.data.account}</em>`;
   }
 
-  function setOpacity(
+  function setVisibility(
     node: SVGTextElement,
     param: HierarchyRectangularNode<AccountHierarchyDatum>
   ) {
     function update(d: HierarchyRectangularNode<AccountHierarchyDatum>) {
       const length = node.getComputedTextLength();
-      node.style.opacity =
-        d.x1 - d.x0 > length + 4 && d.y1 - d.y0 > 14 ? "1" : "0";
+      node.style.visibility =
+        d.x1 - d.x0 > length + 4 && d.y1 - d.y0 > 14 ? "visible" : "hidden";
     }
     update(param);
     return { update };
@@ -60,7 +60,7 @@
     >
       <rect fill={fill(d)} width={d.x1 - d.x0} height={d.y1 - d.y0} />
       <text
-        use:setOpacity={d}
+        use:setVisibility={d}
         on:click={() => router.navigate(urlFor(`account/${d.data.account}/`))}
         dy=".5em"
         x={(d.x1 - d.x0) / 2}
@@ -76,11 +76,6 @@
 <style>
   svg {
     shape-rendering: crispEdges;
-  }
-
-  rect {
-    stroke: var(--color-treemap-text);
-    stroke-width: 2px;
   }
 
   text {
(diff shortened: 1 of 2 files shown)


frontend/src/charts/Treemap.svelte
```diff
@@ -57,11 +57,11 @@
     <g
       transform={`translate(${d.x0},${d.y0})`}
       use:followingTooltip={() => tooltipText(d)}
+      on:click={() => router.navigate(urlFor(`account/${d.data.account}/`))}
     >
       <rect fill={fill(d)} width={d.x1 - d.x0} height={d.y1 - d.y0} />
       <text
-        use:setOpacity={d}
-        on:click={() => router.navigate(urlFor(`account/${d.data.account}/`))}
```
yagebu, line 64: I believe this was intentional to not make the whole chart clickable. Can you revert this change, which is unrelated to the bug fixes AFAICT?

yagebu (changes requested):
Thanks for these fixes :)

frontend/src/charts/Treemap.svelte
jeffwheeler: Ah, you're correct that this wasn't described in the original pull request comment.

With the original behavior, it was still possible to click on the center of small rectangles because the text was present but transparent. Changing the text to hidden means that it becomes impossible to click without this part of the change. But the original behavior was buggy, because text labels often overlapped (invisibly).

I think that this is probably the cleanest way to go. The original click targets were basically random near small rectangles.

yagebu: As I've said, the non-clickability of most of the area was intentional (and hence the use of opacity to make even hidden account names clickable), see https://github.com/beancount/fava/issues/490#issuecomment-294529035 for some context. I'd be fine if accounts whose name isn't shown just aren't clickable anymore.

frontend/src/charts/Treemap.svelte
```diff
@@ -38,14 +38,14 @@
...
+  function setVisibility(
     node: SVGTextElement,
     param: HierarchyRectangularNode<AccountHierarchyDatum>
   ) {
     function update(d: HierarchyRectangularNode<AccountHierarchyDatum>) {
       const length = node.getComputedTextLength();
-      node.style.opacity =
-        d.x1 - d.x0 > length + 4 && d.y1 - d.y0 > 14 ? "1" : "0";
+      node.style.visibility =
+        d.x1 - d.x0 > length + 4 && d.y1 - d.y0 > 14 ? "visibile" : "hidden";
```
yagebu, line 48: Type

frontend/src/charts/Treemap.svelte
jeffwheeler: OK, I've changed per your request.

I understand where you're coming from, but it seems very counter to Fitts's Law to minimize the clickable area. Anyways, you get to have what you want. :)

frontend/src/charts/Treemap.svelte
jeffwheeler: Fixed, thanks

frontend/src/charts/Treemap.svelte
yagebu: > but it seems very counter to Fitts's Law to minimize the clickable area. 

Sure, but I currently don't really see a simple alternative without making these chart rather unusable on mobile.

Issue #1094, opened 2020-04-25
Grey out charts after today's date


redstreet:
It would be less distracting to cut off the time axis at today's date. Currently, if "2020" is selected, for instance, the graph extends the line all the way to the end of 2020 in the Balance Sheet->Net Worth graph:

![image](https://user-images.githubusercontent.com/691755/80271954-42055800-867a-11ea-85ce-78c4e239402d.png)

yagebu:
If think it's more consistent if the time filter sets the start and end date irrespective of the current date. If you want to cut off today, just set the time filter to `2020 - today`

redstreet:
I agree that it's more consistent in its current form, but I wonder if it can be made more useful, intuitive, and user friendly.

I'm thinking it's rare that anyone wants to see to the chart go beyond today in most use cases. The filter would still allow for that when explicitly set; but would cut off to today where it makes sense. What's intuitive?

Easier cases:
2020: 2020-01-01 - today
2018 - 2020: 2018-01-01 - today
2018 - 2020-08-31: 2018-01-01 - 2020-08-31

Trickier cases. These are uncommon, so they matter less IMO:
2019 - 2021: 2019-01-01 - 2021-12-31
2021 - 2024: 2021-01-01 - 2024-12-31

Anyway, that's my two cents. 'today' helps, thanks for the pointer. Please feel free to close this if you think it doesn't provide much value.

redstreet:
BTW, `today` doesn't seem to work in 1.14. Is it newer? I don't see it in the help either.

yagebu:
> I agree that it's more consistent in its current form, but I wonder if it can be made more useful, intuitive, and user friendly.

I personally prefer the current behaviour and find it more intuitive that any special casing based on the current date. Of course the chart points beyond the current date don't provide much value, but I find the chart more useful due to their more symmetric structure. In your `2018-2020`, I'd want all the three years to be of the same width...

> BTW, today doesn't seem to work in 1.14. Is it newer? I don't see it in the help either.

Sorry, it's `day`, so `year - day` for the current year for example. It's all documented in the help pages (filters).

jpluscplusm:
Sorry to add-to/re-open a closed discussion, but I had a very small insight into this issue, and a compromise suggestion.

[...]

PR #1282, opened 2021-05-23
Implement #1094: Desaturate future dates in charts

jeffwheeler:
This change implements #1094, graying out dates after today. There are some subtleties in getting the charts to not have discontinuities, which is why `isUpToToday` is not the strict inverse of `isFuture`.

Screen Shot 2021-05-22 at 5 47 43 PM

Please let me know any changes you'd like to see. I imagine you might prefer slightly different names.


yagebu (changes requested):
Thank you for the PR :)

I'd rather not have duplicate code for all these `<path>`s. Two alternatives that could work:

- The path could be kept as a single element and a `<filter>` with an appropriate value for `filterUnits` could be used (see https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/filterUnits)
- Maybe a rectangle overlaying the chart could also work. I'm not sure exactly the same look can be achieved, but that's not necessary IMHO. Maybe white/grey with a low opacity? Or grey and some playing around with `mix-blend-mode` (https://developer.mozilla.org/en-US/docs/Web/CSS/mix-blend-mode)?

Just having an overlaid rectangle would also make this easier to apply to the bar charts.

jeffwheeler:
@yagebu Appreciate the feedback. I've reimplemented the change using a `<filter />` element per your description, setting `filterUnits` as appropriate.

Here are screenshots. It's trivial to change the exact styling to however you like (perhaps in a followup commit you can tune it to your tastes). In these screenshots, I manually set `today` to an old date because the sample data doesn't extend past today's real date.

Screen Shot 2021-06-05 at 4 17 22 PM
Screen Shot 2021-06-05 at 4 17 42 PM

You're correct that this avoids duplicated code, but I think there are two issues with this approach:

1. It makes the implementation for bar charts more difficult. In bar charts, it is difficult to translate today's date into an x position.
2. In my opinion, styling is more awkward at the boundaries. For example, when `today` is also marked with a `circle`, we apply the filter to only a fraction of the line.

Screen Shot 2021-06-05 at 4 23 53 PM

I've left the two commits unsquashed, so you can play with both options.

yagebu:
I don't have much experience with SVG filters and in retrospect `filterUnits` seems rather unrelated to the issue at hand, so that probably wasn't a very helpful suggestion - sorry about that. Great job on getting the `<filter>` to work :)

I've played around with this filter, used a `<feColorMatrix type="saturate" ...` and tried to shorten the code a bit. Setting the x attribute on the filter actually seems sufficient so now the filter is really quite concise. I've pushed my changes to your branch.

> It makes the implementation for bar charts more difficult. In bar charts, it is difficult to translate today's date into an x position.

We can still use a CSS filter for the bar charts, so I don't think that's much of a problem.

> In my opinion, styling is more awkward at the boundaries. For example, when today is also marked with a circle, we apply the filter to only a fraction of the line.

I guess shifting the filter by 1 or two pixels should fix that.

frontend/src/charts/LineChart.svelte
```diff
@@ -108,9 +107,12 @@
       transform={`translate(0,${innerHeight})`}
     />
     <g class="y axis" use:axis={yAxis} />
-    <g filter={hasFutureData ? "url(#futureColorFilter)" : ""}>
+    <g>
```
jeffwheeler, line 110: Since you've moved the `filter` property into the children elements, you can remove this `<g />` entirely. I added it in the previous commit but it's not needed anymore.

jeffwheeler:
I tried it out, looks good! Thanks for the additions.

frontend/src/charts/LineChart.svelte
yagebu: Thanks :)

yagebu:
> I tried it out, looks good!

Cool, then I'll merge this.

> I had no idea you could push to my branch...

There's a checkbox to allow/disallow this when you create the PR (which is enabled by default nowadays). First time I've used it :D

yagebu (approved):

redstreet:
Nicely done 🙂. Thank you, everyone! 🙏

PR #1285, opened 2021-05-23
Add non-periodic budgets

aronsky:
Current budgets in fava are great, but have their limits. Specifically, it's not possible for the same category to have overlapping budgets either for different intervals. This is obviously an conscious choice, but this PR attempts to add this capability (to an extent), while maintaining backward compatibility and staying true to fava's budget implementation.

Case in point: I have 2 dogs, and I'm buying them food, which is a great candidate for a monthly budget of the `Expenses:Pets` account. I also have a vet subscription for them, which I renew annually, and would like to budget as well. Currently, fava leaves me with 3 options to deal with this:
1. Create separate sub-accounts - `Expenses:Pets:Food` and `Expenses:Pets:Vet`. Workable in this case, but I can see this getting out of hand:
   ```
   2021-01-01 custom "budget" "Expenses:Pets:Food" "monthly" 100 USD
   2021-06-01 custom "budget" "Expenses:Pets:Vet "yearly" 800 USD
   ```
2. Create an annual budget that's the sum of the expected vet charge + 12 times the monthly expected expense on food. This will work fine for annual views, but will be inaccurately represented in monthly views:
   ```
   2021-01-01 custom "budget" "Expenses:Pets "yearly" 12*100 + 800 USD
   ```
3. Create an extra budget entry for the month on which the vet subscription is renewed. With the current implementation, the user has to calculate the sum of the vet subscription and the food expense for that entry, and add another entry for the next month to reset the budget for food expenses only. Too complicated:
   ```
   2021-01-01 custom "budget" "Expenses:Pets" "monthly" 100 USD
   2021-06-01 custom "budget" "Expenses:Pets" "monthly" 100 + 800 USD
   2021-07-01 custom "budget" "Expenses:Pets" "monthly" 100 USD
   ```

My solution is to introduce  non-periodic budgets. Unlike the regular budgets in fava, these non-periodic budgets are only applied to the period from their start date to the date of the next interval. Furthermore, they are applied on top of the existing budgets (both periodic  and non-periodic), so one-time expenses can be budgeted for without fidgeting with the rest of the budget for the same account:
   ```
   2021-01-01 custom "budget" "Expenses:Pets" "monthly" 100 USD
   2021-06-01 custom "budget" "Expenses:Pets "month" 800 USD
   ```
This will budget 100 USD monthly, and in June the budget will be upped by 800 USD for a single month.

For now, I chose to use `day`, `week`, etc. (in place of `daily`, `weekly`, etc.) to differentiate non-periodic budgets from their periodic counterparts. I'm open to suggestions regarding different naming, or use of an extra argument to the budget entry.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 /.coverage
 /.eggs
 /.tox
+/.vscode
 /.idea/
 /build/
 /pip-wheel-metadata/
@@ -13,6 +14,7 @@
 /src/fava.egg-info
 /htmlcov
 /venv
+/.venv
 
 node_modules
 *.pyc
--- a/src/fava/core/budgets.py
+++ b/src/fava/core/budgets.py
@@ -2,6 +2,7 @@
 import datetime
 from collections import Counter
 from collections import defaultdict
+from itertools import chain
 from typing import Dict
 from typing import List
 from typing import NamedTuple
@@ -13,6 +14,7 @@ from beancount.core.number import Decimal
 from fava.core.module_base import FavaModule
 from fava.helpers import BeancountError
 from fava.util.date import days_in_daterange
+from fava.util.date import get_next_interval
 from fava.util.date import Interval
 from fava.util.date import number_of_days_in_period
 
@@ -25,6 +27,7 @@ class Budget(NamedTuple):
     period: Interval
     number: Decimal
     currency: str
+    periodic: bool = True
 
 
 BudgetDict = Dict[str, List[Budget]]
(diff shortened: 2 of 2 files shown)


yagebu:
Hey, thank you for opening a PR! Sorry for taking ages to take a look at this.

This is not something I'll really want to merge, mainly considering how complicated this makes the budget matching logic. Also conceptually, I find the syntax somewhat confusing - it looks similar to setting a new budget but its actually a whole separate kind of budget that is added to my existing budget which continues.  What would happen if I set a `month` budget multiple times in a month? I'd expect them to accumulate, but I guess that is not what happens.

Of your suggested workarounds, 1. seems to be the cleanest to me - and if you want to change the budget for just one month, I find 3. to be nicer and more explicit than the new syntax. But you prefer the suggested syntax and that's fine - I don't even think much is gained from adding this to Fava itself, I believe that this could be more easily implemented as a plugin, which turns your syntax into 3. :)

aronsky:
Hi, thanks for the response!

I understand the issue with the syntax being confusing. It's partially due to my attempt to keep everything backwards-compatible. It's probably possible to make a more clear syntax that allows for both periodic and one-time budgets, but making it a breaking change for everyone using budgets would be... undesirable, to put it mildly :) Maybe another approach could be to introduce a differently named setting (something like `budget2`, but better sounding), and develop a new syntax for it, that'll allow for both types of budgets without making it confusing. That would also allow to resolve the second issue you raise (overlapping budgets) properly, because the syntax could be improved to deal with that. If that's something that you'd consider, I'll be happy to get started on it.

The idea of a plugin is a good one, too. I'll probably go that route in the meanwhile.

yagebu:
Yes, I think another name for it would probably be good, and I guess this "budget2" wouldn't even need to have an interval specified (and could simple apply on the given day).

In any case, I think it would be better off in a plugin instead (which would be welcome at https://github.com/beancount/fava-plugins)

Issue #716, opened 2018-03-28
Stacked charts


yagebu:
As suggested by @vlaci in #258:

> I think a stacked bar chart would make sense e.g. for the Monthly Changes graphs. That would make much easier to see the dynamics how accounts' contributions changed from month to month to the expenses

Just like in the treemaps and sunburst charts, amounts of the opposite sign than expected from the account (e.g. a negative value for some expenses subaccount) are problematic and will just get ignored. Alternatively they could be displayed under the x-axis I guess.

yagebu:
Some ideas from #1251:


> Studying how expenses change over time (trying to pinpoint where a rise in expenses came from currently requires flipping back and forth between the top-level expenses chart and charts for expense sub-accounts)

> Visualizing market value of investment portfolios relative to cost basis over time

> [This is another cool example](https://imgur.com/a/0eDD9KE) of a stacked chart that I've been wanting to replicate in Fava.

aclindsa:
@yagebu (bringing discussion here from https://github.com/beancount/fava/issues/1251):

> I guess this could be implemented quite similarly to the existing charts, trying to have the new data structures that are generated by Fava align with the existing ones (in some sense the data required is a combination of the existings hierarchical account charts and the time series charts).

I haven't dug into the code to be certain, but I wonder if the necessary data structure already exists. I noticed that if I add `&conversion=units` on some reports (the Balance Sheet, for example), the line chart appears to simultaneously show the number of units I have of different commodities, varying over time. I think that is the same "shape" of data you would need for a stacked bar chart, right?

> If possible, I think it would also be good if these new charts are not entirely new charts (in terms of the code generating the charts) but we add this as an additional feature to the existing code.

By this, do you mean that you believe the existing bar chart code could and should be modified such that all bar charts become 'stacked' in a way? (And that the bar charts which exist today would then become special cases which happen to only have one data series)

aclindsa:
@yagebu I've come up with some other design questions while thinking about this, and would appreciate your opinion:

When visualizing an account hierarchy with a stacked chart, it is easy for there to be too many accounts to display them all in a meaningful way. I see two potential approaches to address this: 1) Combing accounts with values less than a certain threshold into 'other' and displaying it as the combined total of the smallest accounts, and 2) Rolling amounts belonging to sub-accounts beyond a certain depth into their parent accounts. Perhaps a combination of the two could also be used.

[...]

PR #1286, opened 2021-05-24
WIP: Stacked charts

jeffwheeler:
This pull request has a draft implementation of stacked charts (#716).

Screen Shot 2021-05-28 at 5 54 23 PM
Screen Shot 2021-05-28 at 5 54 32 PM

I know of the following issues:
- ~~Budgets are broken.~~
- Multiple operating currencies is broken. (I believe this was already broken, though? For example, I'd expect [this page](https://fava.pythonanywhere.com/huge-example-file/account/Income:US:Federal:PreTax401k/?conversion=units&interval=quarter) ("Changes" chart) in the example file to show `IRAUSD` changes, but the chart legend shows `USD` and no data.)
- The `descend` logic may be suboptimal. Basically, when `descend=True` is set in the `interval_totals` function call, it calculates inventories for all accounts one level below the listed accounts. But I bet this could be a lot smarter.
- ~~I would like a line chart on top of the bar showing the `total_inventory` value, which is the net of all values listed.~~
- Code needs significant cleanup.

I would appreciate any feedback.


aronsky:
This is awesome! I am currently working on adding features to budgets, but this was on my TODO list - happy to see you already got it working, and judging by the screenshots - it looks great!

jeffwheeler:
I've updated this pull request with graphical and code improvements. The known issues in the original post remain mostly accurate.

ksesong:
I feel the "net/sum" marker unnecessarily complex, a straight line would fit better and be less obtrusive.

A quick mockup:
image

jeffwheeler:
@ksesong Sure. I have updated accordingly. I've made the bar thicker than in your mockup so that it's easier to hover on top of, but I've parameterized it in `netValueMarkerHeight` so you can play with the value. :)

ksesong:
I would keep it the same thickness as the axis thickness, or the line chart thickness. Making it too thick would confuse it with an account space. You could add some invisible padding around it to make it easier to hover/select.

aclindsa:
> I would keep it the same thickness as the axis thickness, or the line chart thickness. Making it too thick would confuse it with an account space. You could add some invisible padding around it to make it easier to hover/select.

How thin can the account portions of the stacked bars get, and how different will their colors be from this marker? One benefit of the previous implementation was that it made the net/sum marker more visually distinct from the account data.

ksesong:
Looking at the tree map charts, the label is always in dark grey, with a colored background, so I figured that using the same color scheme should work out fine. I am not certain of the relevance of the marker in the non-"net profit" charts, but these are only my observations, @yagebu should probably have a look.

jeffwheeler:
I quite like the current version, actually.

1. The problem with the circles for the "net value" was that it didn't work as well when the lines were dense, e.g. over many years or smaller intervals. The underlying data became invisible.
2. I believe the current version is still distinguishable from a category for two reasons: 1) it is the only gray value, and 2) it is partially transparent, which allows users to see the underlying data. When users hover, though, I fade away the underlying data and make the net value more visible:

Screen Shot 2021-05-29 at 1 54 34 PM

aclindsa:
@jeffwheeler Thanks for taking the time to put this together! I am happy to see this being worked on. I had some time to try this out and have a few comments from my first impressions (and I'm sorry for how much there is - it just shows how much I want and care about this feature):

- I find the 'Net' bar to be more useful in some contexts than others. For example, I've found it *always* useful and desirable in 'Net Profit' charts, but that it can obscure smaller some smaller accounts when viewing 'Expenses' charts. I wonder if it might make sense to only enable it in certain contexts?
- What determines which order the bars are stacked in? Mine were neither sorted alphabetically or by largest/smallest value (trying out https://github.com/jeffwheeler/fava/commit/d3723c719f872b329d22f9c1f1042164e309808e). I wonder if it would make sense to sort account values so that the largest cumulative values are closest to the x-axis.
- Having the bars inside the chart be 'clickable' and take you to that account's view is a great idea! This feature might be more 'discoverable' to folks if it was done in a way so that the browser would present a 'link' cursor instead of the standard one.
- It might also be useful if we found a way for clicking on an account in a stacked bar chart to take you to a view which showed basically the same chart, but broken down by that account's sub-accounts instead! For example, if I clicked an 'Expenses:Taxes' bar in the top-level stacked bar chart for Expenses, it would be really neat to be taken to a chart showing the same expenses chart, but this time limited to only 'Expenses:Taxes' and broken down by its sub-accounts. That would make it easier to dig into the account hierarchy in a more fine-grained way since the current implementation (probably rightfully so) only displays stuff at the granularity of up to second-level accounts.
- Would it make sense to use these stacked charts for the 'Net Worth' chart, too? (I think my answer is "yes!", but I'm curious if others feel differently)
- It would make it easier to 'find' a bar in the stacked bar chart if hovering over that bar's name or color in the legend visually highlighted the bars (the same way hovering over the bars does now).
- Overall this looks awesome - I am very excited for this feature!

jeffwheeler:
@aclindsa Thanks for all the feedback! Many of your comments are things I had thought about myself, I very much agree with your sentiments.
 
> * I find the 'Net' bar to be more useful in some contexts than others. For example, I've found it _always_ useful and desirable in 'Net Profit' charts, but that it can obscure smaller some smaller accounts when viewing 'Expenses' charts. I wonder if it might make sense to only enable it in certain contexts?

This is something I've been debating too. There are two parts I haven't figured out:

1. Currently, hovering on the "net" bar is the only way to see the total value. So, if I were to remove that, I would need to find a different way to show that. One idea is to include it on _all_ tooltips?
2. For the most part, "Expenses" or "Income" charts are all a single sign. Except... they're not always. For example, when I get a refund, I treat that as a a negative change to the Expense account. For this reason, the "net" value actually makes sense to show the net expenses of a given month. Not sure the best way to handle this.

> * What determines which order the bars are stacked in? Mine were neither sorted alphabetically or by largest/smallest value (trying out [jeffwheeler@d3723c7](https://github.com/jeffwheeler/fava/commit/d3723c719f872b329d22f9c1f1042164e309808e)). I wonder if it would make sense to sort account values so that the largest cumulative values are closest to the x-axis.

I believe it should be alphabetical, with the first values closest to zero, and the last values furthest from zero. This can be adjusted by calling `.order(stackOrderAscending)` or similar right before (or after) `.offset(stackOffsetDiverging)` (options [documented here](https://github.com/d3/d3-shape)). But I don't see a good way to make it ordered by largest on the inside. This seems not so easy. 😟 

> * Having the bars inside the chart be 'clickable' and take you to that account's view is a great idea! This feature might be more 'discoverable' to folks if it was done in a way so that the browser would present a 'link' cursor instead of the standard one.

Great idea, I've just added this.

> * It might also be useful if we found a way for clicking on an account in a stacked bar chart to take you to a view which showed basically the same chart, but broken down by that account's sub-accounts instead! For example, if I clicked an 'Expenses:Taxes' bar in the top-level stacked bar chart for Expenses, it would be really neat to be taken to a chart showing the same expenses chart, but this time limited to only 'Expenses:Taxes' and broken down by its sub-accounts. That would make it easier to dig into the account hierarchy in a more fine-grained way since the current implementation (probably rightfully so) only displays stuff at the granularity of up to second-level accounts.

I agree, I'd like this too. I think this would be best to implement in a second change which allows directly linking to pages with specific charts selected. Right now, I link to `account/Income/?time=2021` or similar, but if it was possible to link to `account/Income/?time=2021&activeChart=changes` or similar, we could implement this idea. Right now, you have to select the "Changes" chart after clicking an account.

> * Would it make sense to use these stacked charts for the 'Net Worth' chart, too? (I think my answer is "yes!", but I'm curious if others feel differently)

Yes!

> * It would make it easier to 'find' a bar in the stacked bar chart if hovering over that bar's name or color in the legend visually highlighted the bars (the same way hovering over the bars does now).

Yes! I'd like this too. Perhaps in a second commit? It can be extended to line charts too.

> * Overall this looks awesome - I am very excited for this feature!

Thanks :)

jeffwheeler:
Screen Shot 2021-06-10 at 8 46 26 PM

I now support budgets. I think the way I've graphically displayed them is pleasing, but as always, visuals are subjective.

jeffwheeler:
Rebased

jeffwheeler:
Fixed tests.

Apologies for not fixing those before, I didn't realize they were legitimate failures. I've had a lot of trouble running tests locally.

yagebu:
Overall, this looks great, thank you very much :) I don't know right now when I'll find the time to review this thoroughly, but it's high up on my Fava todo list.

> I've had a lot of trouble running tests locally.

What kind of trouble? Anything that could be helped by better documentation?

xinluh:
Took this out for a spin - this is so good! A few bugs / comments:
- The "net" bar currently have the wrong sign when `invert-income-liabilities-equity` setting is on. It should also be reversed.
image
- The height of the net bar might be better to scaled with its width instead of being fixed. At its current height, it works okay with fewer bars on the graph, but with longer time period and anything at monthly frequency or less, it starts to look intrusive:
image
- Is there any way to make the adjacent colors more contrasting, i.e. use ordinal color map instead of a continuous one? It can be really hard to tell different sections of the bar apart; here is an example where all colors of the reddish-purple shades are basically indistinguishable to my eyes : 
image
- Also echo @aclindsa's comment that the net bar is less helpful and blocks other useful information in the Expenses/Income settings where all category are normally expected to be of the same sign. There are definitely edge cases with "wrong" sign postings in those accounts (like refunds), but there are already have plenty precedences of graphs not working well in these edge cases (like in sunburst plot). Potentially showing the negative part of each category as separate below the axis block might be more visually helpful than the net bar here.

jeffwheeler:
> > I've had a lot of trouble running tests locally.
> 
> What kind of trouble? Anything that could be helped by better documentation?

I ran into an issue [similar to this|https://stackoverflow.com/questions/58403923/failed-to-load-plugin-import-declared-in-eslint-config-standard-canno]. The second comment resolved it. I don't understand how all this Javascript stuff works at all, haha.

jeffwheeler:
> Took this out for a spin - this is so good! A few bugs / comments:
> 
> * The "net" bar currently have the wrong sign when `invert-income-liabilities-equity` setting is on. It should also be reversed.

Thanks for flagging. I've fixed this. Actually, the net value bar was correctly signed, but the other values were wrong.

> * The height of the net bar might be better to scaled with its width instead of being fixed. At its current height, it works okay with fewer bars on the graph, but with longer time period and anything at monthly frequency or less, it starts to look intrusive:

Good idea. Done. It's now the width*0.2, bounded by `minNetValueMarkerHeight` (`2`) and `maxNetValueMarkerHeight` (`6`).

> * Is there any way to make the adjacent colors more contrasting? It can be really hard to tell different sections of the bar apart; here is an example where all colors of the reddish-purple shades are basically indistinguishable to my eyes :
> * Also echo @aclindsa's comment that the net bar is less helpful and blocks other useful information in the Expenses/Income settings where all category are normally expected to be of the same sign. There are definitely edge cases with "wrong" sign postings in those accounts (like refunds), but there are already have plenty precedences of graphs not working well in these edge cases (like in sunburst plot). Potentially showing the negative part of each category as separate below the axis block might be more visually helpful than the net bar here.

I'm less sure how to approach these two.

yagebu (changes requested):
Hey, I took a first look at this PR :)

I really like it overall. I personally think that the net sum marker should just be a thin line and that the total should be part of the tooltip, since I find the total to be quite important and I don't think one should have to hover some small bar to see it. Highlighting the hovered account is nice but I found that the hover effect caused quite bit of flickering (and with more accounts the fading had the appearance of switching account colors) - maybe the hovered account could be highlighted without fading all other bars?

As you've stated in your initial comment, the python side looks quite suboptimal and the frontend parts could deserve some cleaning up, so I haven't left any code comments yet. Could you try to get this PR into a more reviewable shape? I've recently refactored the charts code to be more modular, so your changes to `index.ts` will have to be to `bar.ts` now - maybe that helps with the rebase.

> Multiple operating currencies is broken. (I believe this was already broken, though? For example, I'd expect this page ("Changes" chart) in the example file to show IRAUSD changes, but the chart legend shows USD and no data.)

The bar charts only show values in the operating currencies, the operating currency on the link page is USD while the account only has IRAUSD - so nothing is shown as expected. Could you try to get multiple currencies to work? (probably the easiest way to test this is to select a conversion currency, which - for the charts at least - will get added to the list of operating currencies)

varac:
@jeffwheeler What's the state, I'd love to see this merged soon :bar_chart:

jeffwheeler:
Hi @varac, thanks for the feedback.

To be honest, I kinda lost interest after this PR churned a lot. I haven't been working on Fava anymore. But anyone should feel free to continue working on this and bring it to the finish.

Jeff

yagebu:
Superseded by #1408

PR #1296, opened 2021-06-16
Dockerfile: do not pass environment variables to command-line

SakuragawaAsaba:
Since `BEANCOUNT_FILE` (`BEANCOUNT_FILES` as well) and `FAVA_`-prefixed
environment variables will be used, passing them in command-line only creates
duplicates.


yagebu (approved):
Thanks :)

PR #1304, opened 2021-07-24
add tree view

xiaoniu-578fa6bff964d005:
Similar to https://github.com/beancount/fava/pull/1137.
All functions are implemented in frontend.

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -37,6 +37,7 @@
         "d3-selection": "^3.0.0",
         "d3-shape": "^3.0.0",
         "d3-time-format": "^4.0.0",
+        "path-browserify": "^1.0.1",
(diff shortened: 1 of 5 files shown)


yagebu (changes requested):
Thanks for the PR!

This doesn't pass the CI as you have barely added any types - could you try to fix that? The code that creates the folder tree should be tested. I'm not really keen to add the `path-browserify` dependency, I think for our case a splitting by regex might suffice - if not, you could turn the paths into lists of its elements and pass that to the frontend. And I think that the creation of the folder tree could be achieved by using the `stratify` function that already exists, that would shorten the new code a bit.

yagebu:
Closing this stale PR, happy to reopen if this is continued :)

PR #1307, opened 2021-08-07
add filter for posting amount

xinluh:
Add ability to filter by posting amount. This is helpful for locating a specific transaction or locating transaction of certain amount (i.e. all transactions last month > $100)


yagebu (changes requested):
Hey, thanks for this PR!

Since we already support filters on posting attributes (with `any()` and `all()`), this amount filter should work similarly. Then we also wouldn't have to special-case the "amount" key but could rather check for all attributes if they want to do a numerical comparison.

xinluh:
@yagebu perhaps there is a different perspective here on usability and feature discoverability too. I have spent a lot of time trying to figure out how to filter by amount - I figured it out once but promptly forget it and have to spend same amount of time next time because it somehow didn't wire in my brain (something like `any(amount:"1 USD")` - that doesn't work btw). It seems unnatural and verbose to type out so much to search for something so simple.

Another potential disconnect is how often amount search is useful - often used feature probably deserved better product shortcut (i.e. straightforward `amount:"1 USD"`) , less often used feature maybe can be delegated to the more verbose version. I find amount search very useful - oftentimes by amount is the easiest way to search for a particular transaction, especially when reconciling matching transactions like reimbursement.

And lastly, addressing the concern of "special-case". Are you concerned about:
- special casing in code,  i.e. code complexity? This doesn't seem like a big issue to me, since there needs to be additional logic to handle numerical comparison anyways, and special casing for just amount seems okay since amount is the only common numerical field beside meta
- special casing in user experience? this goes back the point of usability. I _think_ it is natural to have an `amount:...` filter as a nice complement to the `payee:` or `narration:` search. It seems self explanatory and unlikely to cause false positive matches. I would agree that It is slightly weird that the amount search works on posting instead of entries, but IMO there is no confusion because entries always have amount = 0.

yagebu:
Beancount transactions don't have "amounts" and even postings don't have "amounts" - postings have units and cost potentially.  So "amount" as the shortcut to search for posting units is quite inconsistent IMHO. I agree that searching for amounts (but just for exact matches) deserves a shortcut, but this isn't it. Why not add this amount search to the search of payee and narration (and comment for notes) that we do when a search string is given without any attribute?

In addition, the full syntax consistent with Fava's filter syntax should also be implemented: so that e.g. `any(cost: "20 USD")` works.

nzbart:
As someone who would use this feature, I personally think it's important some kind of range syntax is allowed for - such as `>20 USD`. It'd also be great to have an imperfect feature in the product rather than a perfect one that doesn't exist, although I know that can create a maintenance headache in the future. Even only being able to search for exact amounts would be better than the current situation.

yagebu:
Closing this stale PR

PR #1316, opened 2021-08-29
Ellipsize long narration entries in the journal

psimonyi:
This patch makes all entries in the journal table consistently one line high, and if the payee + narration (+ tags + links) is too long to fit on one line, it is cut off with an ellipsis.  The full narration can be displayed by clicking to show postings/metadata or by clicking the date to show the entry source.

### DIFF
--- a/frontend/css/journal-table.css
+++ b/frontend/css/journal-table.css
@@ -159,18 +159,23 @@
 }
 
 .journal .description {
-  display: flex;
   flex: 1;
-  align-items: center;
   padding-left: 8px;
+  overflow: hidden;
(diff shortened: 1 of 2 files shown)


yagebu (changes requested):
Thanks for the PR, I like this change! :)

This branch seems to be quite outdated, which probably causes the CI failures - could you rebase it against the latest commit on the main branch?

psimonyi:
That makes sense.  I've been carrying these patches for a while.

yagebu (approved):
Thanks again and congratulations on your first contribution to Fava :)

PR #1317, opened 2021-08-29
Make sidebar query links fill in the full query

psimonyi:
It's handier to fill in the full query instead of `run <queryname>` so the user can see what the query was and edit it.

### DIFF
--- a/src/fava/templates/_aside.html
+++ b/src/fava/templates/_aside.html
@@ -20,7 +20,7 @@
     {% if id == 'query' and user_queries %}
     <ul class="submenu">
       {% for query in user_queries %}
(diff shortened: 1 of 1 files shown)


yagebu (changes requested):
Thanks! I don't really use this feature but I agree that this is probably more useful :)

Same as the other PR, could you rebase this?

psimonyi:
Rebase done here too.
I use this most often as a cheap sort of parameterized query.  So for example I have a query that pulls last month's transactions with some conditions but sometimes I want the month before last or the current month, if I don't do my finances on time.

yagebu (approved):
Thanks again :)

> I use this most often as a cheap sort of parameterized query. So for example I have a query that pulls last month's transactions with some conditions but sometimes I want the month before last or the current month, if I don't do my finances on time.

I think when I added this, I also wanted to use this to highlight the existence of the `run..` syntax. But like this it should be much more useful (and in Fava, one doesn't really need to know about `run..`)

PR #1322, opened 2021-09-01
Sunburst colours

psimonyi:
This PR is a bunch of related changes to the sunburst chart.  Some of the changes could probably be separated out if you wanted.  But since they came up together and the solution to most of these issues is to build a new tree from which to draw the chart, I implemented them all together.

- Colour conflicts: sometimes adjacent wedges in the chart had the same colour, which is visually confusing, especially without any borders.  Solution: add white borders and ensure adjacent wedges always have different colours.
- Meaning of colours: Assigning consistent colours to different accounts is an interesting idea, but the colour is used in so few places that I didn't realize that was happening until I read the code.  This experiments with using colour to emphasize the hierarchy by assigning each first-level wedge a colour family, and each of its children gets a shade from that family.  That highlights the hierarchy for up to 3 levels of account nesting.
    - More levels could be added later, for example by using different brightness levels for the third-level wedges, though that would require giving up the perceptual uniformity we get from the Lch colour space.
    - This scheme does mean there is a level (currently the 3rd) at which adjacent wedges are assigned the same colour, which I was trying to avoid, but now the same-coloured accounts are related to each other and so the separation with a border should be enough to avoid confusion. 
- Tiny wedges: Small slivers mess up the colour conflict detection because you can have a wedge less than 1px wide between two wedges of the same colour.  Also, tiny wedges are just messy when they're really too small to see.  Solution: replace all wedges that are only a couple of pixels wide, grouping them into a special "(other)" wedge.
- Negative balances: Accounts can have negative balances but the sunburst chart cannot represent negative quantities.  Previously situations like this could be just buggy (e.g. an account with a negative balance but subaccounts with positive balances would result in extra wedges floating at the second level).  Solution: don't represent them in the chart, and additionally do not represent siblings of negative accounts either (except at the first level) because the proportions will be wrong.
- Colour choice: This is pretty subjective, but I've chosen lighter, less-saturated colours for the wedges because mostly they're fairly large areas of the screen.

Also,
-  My account hierarchy is fairly shallow, so if you have a deeply-nested account hierarchy I'd like to know how this feels!
- I'm not sure of the best way to get the account sign from the original data to `sunburstTree`.


psimonyi:
Finally got eslint running locally, so I can look at some of the lint failures.

yagebu:
Hey, thanks for this PR :) I was initially waiting for this to get into a state that passes CI and then it dropped off my radar. Overall this is something I'd be happy to add - I'm not a big user of the sunburst chart myself but these look like well-founded improvements. Are you still keen to continue working on this? Did you run into some roadblocks getting CI to pass?

psimonyi:
Ah, thanks for the reminder.  Yeah, I remember getting stuck trying to figure out what the linters were trying to get me to change and then other stuff came up and I forgot about it.  I'll try to get back into it this coming week.

redstreet:
Looks like an interesting PR, @psimonyi. https://github.com/redstreet/fava_investor uses the sunburst for its allocation, and I personally use deeply nested sunbursts. Would you still be interested in getting this PR in?

yagebu:
This has gotten pretty stale, so I'll close this PR.

Happy to have this reopened if you ever feel like it :)

PR #1323, opened 2021-09-02
Add before_insert_entry hook

pandu-rao:
This PR provides a hook to process an entry before it is inserted. Extension use cases:
  1. validate/format entry prior to insertion
  2. switch file which stores the entry
  3. etc.

### DIFF
--- a/src/fava/core/extensions.py
+++ b/src/fava/core/extensions.py
@@ -93,6 +93,10 @@ class ExtensionModule(FavaModule):
         for ext in self.exts_for_hook("after_entry_modified"):
             ext.after_entry_modified(entry, new_lines)
 
+    def before_insert_entry(self, entry: Directive) -> None:
+        for ext in self.exts_for_hook("before_insert_entry"):
+            ext.before_insert_entry(entry)
+
     def after_insert_entry(self, entry: Directive) -> None:
         for ext in self.exts_for_hook("after_insert_entry"):
             ext.after_insert_entry(entry)
--- a/src/fava/core/file.py
+++ b/src/fava/core/file.py
@@ -168,6 +168,7 @@ class FileModule(FavaModule):
                 insert_options = fava_options["insert-entry"]
                 currency_column = fava_options["currency-column"]
                 indent = fava_options["indent"]
+                self.ledger.extensions.before_insert_entry(entry)
                 fava_options["insert-entry"] = insert_entry(
                     entry,
                     self.ledger.beancount_file_path,
--- a/src/fava/ext/__init__.py
+++ b/src/fava/ext/__init__.py
@@ -51,6 +51,9 @@ class FavaExtensionBase:
             self.config = None
         self.name = self.__class__.__qualname__
 
+    def before_insert_entry(self, entry: Directive) -> None:
+        """Called before an `entry` is inserted."""
+
     def after_entry_modified(self, entry: Directive, new_lines: str) -> None:
         """Called after an `entry` has been modified."""
 
--- a/tests/test_extensions.py
+++ b/tests/test_extensions.py
@@ -12,3 +12,14 @@ def test_report_page_globals():
     assert result == [("PortfolioList", "Portfolio List")]
 
     extension_report_ledger.extensions.after_write_source("test", "test")
+
+
+def test_before_insert_entry():
+    """Extensions provide a before_insert_entry hook method."""
+    hook_ledger = FavaLedger(
+        data_file("extension-report-example.beancount")
+    )
+
+    assert bool(getattr(hook_ledger.extensions, "before_insert_entry"))
+
+    hook_ledger.extensions.before_insert_entry("entry")


yagebu:
Something went wrong with the merge/rebase here... Also this doesn't seem to do anything yet, are you planning to continue this PR?

pandu-rao:
I've shown the changes below. 

I don't know why the rebase/merge is failing. I can create another PR if that helps.

```
diff --git a/src/fava/core/extensions.py b/src/fava/core/extensions.py
index ecafdaa7..dbe6d916 100644
--- a/src/fava/core/extensions.py
+++ b/src/fava/core/extensions.py
@@ -93,6 +93,10 @@ class ExtensionModule(FavaModule):
         for ext in self.exts_for_hook("after_entry_modified"):
             ext.after_entry_modified(entry, new_lines)

+    def before_insert_entry(self, entry: Directive) -> None:
+        for ext in self.exts_for_hook("before_insert_entry"):
+            ext.before_insert_entry(entry)
+
     def after_insert_entry(self, entry: Directive) -> None:
         for ext in self.exts_for_hook("after_insert_entry"):
             ext.after_insert_entry(entry)

diff --git a/src/fava/core/file.py b/src/fava/core/file.py
index 2226096a..fc3c85eb 100644
--- a/src/fava/core/file.py
+++ b/src/fava/core/file.py
@@ -168,6 +168,7 @@ class FileModule(FavaModule):
                 insert_options = fava_options["insert-entry"]
                 currency_column = fava_options["currency-column"]
                 indent = fava_options["indent"]
+                self.ledger.extensions.before_insert_entry(entry)
                 fava_options["insert-entry"] = insert_entry(
                     entry,
                     self.ledger.beancount_file_path,

diff --git a/src/fava/ext/__init__.py b/src/fava/ext/__init__.py
index 861b7a24..7334e3b6 100644
--- a/src/fava/ext/__init__.py
+++ b/src/fava/ext/__init__.py
@@ -51,6 +51,9 @@ class FavaExtensionBase:
             self.config = None
         self.name = self.__class__.__qualname__

+    def before_insert_entry(self, entry: Directive) -> None:
+        """Called before an `entry` is inserted."""
+
     def after_entry_modified(self, entry: Directive, new_lines: str) -> None:
         """Called after an `entry` has been modified."""
```

yagebu:
> I don't know why the rebase/merge is failing.

Maybe try an interactive rebase against the main branch and remove all commits but yours.

And in any case, I don't see how that hook does anything yet (except being called, but then why have it in addition to the `after_insert_entry` hook). A test would also be good too.

pandu-rao:
> Maybe try an interactive rebase against the main branch and remove all commits but yours.
I'll try that.

> And in any case, I don't see how that hook does anything yet (except being called, but then why have it in addition to the `after_insert_entry` hook). A test would also be good too.

I created the `before_insert_entry` hook because I'm using includes to include/organize beancount files. Without this hook, when I use the fava editor widget, the new entry is not created in the file that is open in the editor. Instead, the new entry is always created in the main beancount file.

This hook enables me to write an extension which selects the active file to insert the new entry.

If this makes sense I can look into writing a test case as well.

pandu-rao:
Also, I don't see a test case for the `after_insert_entry` hook. So I don't have a reference test case for `before_insert_entry`.

Would something like this suffice?

```python
def test_before_insert_entry():
    """Extensions provide a before_insert_entry hook method."""
    hook_ledger = FavaLedger(
        data_file("extension-report-example.beancount")
    )

    assert bool(getattr(hook_ledger.extensions, "before_insert_entry"))

    hook_ledger.extensions.before_insert_entry("entry")
```

yagebu:
> Without this hook, when I use the fava editor widget, the new entry is not created in the file that is open in the editor. Instead, the new entry is always created in the main beancount file.

With this hook, how do you determine the currently open file? And how do you get the entry to be inserted into that file? The way that this hook is currently designed (not returning any value), it shouldn't mutate the entry.

pandu-rao:
> With this hook, how do you determine the currently open file? And how do you get the entry to be inserted into that file? The way that this hook is currently designed (not returning any value), it shouldn't mutate the entry.

I have created an extension for that (shown below). Let me know if there is a better way to solve this problem. 

```python
"""Activate-File hook for Fava.

When beancount files are organized using includes, Fava's web-based beancount
file editor creates new entries only in the main file (even if the currently
loaded file is an included file). The correct behavior is to create entries in
the active file.

This extension uses the before_insert_entry hook to activate the file currently
loaded in the editor. The after_insert_entry hook is used to reset the activation.

"""

from urllib.parse import urlparse, parse_qs

from flask import request
from fava.ext import FavaExtensionBase


class ActivateFile(FavaExtensionBase):
    CACHED_FILE_PATH = ''

    def before_insert_entry(self, entry):
        referer = 'referer'.title()
        url = request.headers[referer]
        url = urlparse(url)

        file_paths = parse_qs(url.query).get('file_path')
        if not file_paths:
            return

        beancount_file_path = file_paths[0]
        ActivateFile.CACHED_FILE_PATH = self.ledger.beancount_file_path
        self.ledger.beancount_file_path = beancount_file_path

        return

    def after_insert_entry(self, entry):
        self.ledger.beancount_file_path = ActivateFile.CACHED_FILE_PATH
        return

```

yagebu:
Ah, that's a creative hack :D I'm not sure whether I want to add this hook for mainly this use case though (since your extension really depends on Fava's internal behavior).

So far I've always considered the transaction adding dialog and the editor as two completely separate things, so the current behavior is the intended correct behavior and the `insert-entry` options are the main way to make your transactions end up in different positions and/or files. Are the `insert-entry` options not sufficient to get the transactions into the intended files in your case? Also, if you're already in the editor, isn't it quicker to add a transaction directly there? - what's the main advantage of the dialog for you in that case?

pandu-rao:
> Ah, that's a creative hack :D ...

A hack, no doubt. I'm glad you think it is creative! :-)

You're right about it relying on Fava's internal behavior. Something I'm not too
thrilled about.

> the insert-entry options are the main way to make your transactions
end up in different positions and/or files.

I looked into using insert-entry options before developing this
extension. However, I ran into a problem where it could create most entries
correctly by using the regex functionality. But it fell short when creating
opening-balance entries.

Now thanks to this discussion, I decided to add a suffix to the opening balance
entry. e.g "Equity:Opening-Balances:Chase".

With the above convention, I can create a directive like this (in the
chase.beancount file, say) ...:

```
1970-01-01 custom "fava-option" "insert-entry" ".*:Chase"
```

... and everything seems to work. I'll be testing this further.

I like your proposed solution for the following reasons:

  1. Does not need any new code (especially something that relies on an internal
     mechanism).
  2. The new convention, -an account suffix for opening balances- nicely
     separates amounts that all add up to "Equity:Opening-Balances".

Thanks for your help.

yagebu:
Nice that we found an alternative (and less hackish) solution for you :)

I'll close this PR since we have no use case for this hook anymore - if someone needs it in the future, we can still add this hook :)

Issue #1293, opened 2021-06-08
Directory structure for uploading document


Fjanks:
Hey! I really like the feature to upload files with drag and drop! However, I would like to have the files stored in a different way, because right now, you need to rename directories when renaming accounts and you need to consider the order of postings in a transaction, which I prefer not to be relevant. 
Is it possible to change the method how the files are stored? If not, do you think this could be implemented, e.g. with extensions, so that everybody can use his/her favourite method? Or maybe adding an option to control whether fava creates a directory structure for the documents or just throws all documents in one directory?

yagebu:
What kind of directory structure do you want to store your files in? How do create Beancount document entries for those files?

The current mechanism is intended to work together with Beancount's automatic creation of document entries and hence makes sure that they show up in Fava as documents after being moved. I'm not sure whether I want to add an option that breaks that. So maybe allowing this to be changed by an extension could be preferable (and more flexible).

Fjanks:
I would like to have a directory structure that does not depend on the account structure, for example all files in the same directory. The reason I tried to accomplish this with a new fava-option is just because it seemed to be a simple way and I don't know yet how to do it with an extension. If it breaks Beancount's automatic creation of document entries, I understand that you don't like to add it. 

Is it possible to do this with an extension? If Fava relies on Beancount for finding the documents, does that mean I would need a Beancount plugin to help Beancount finding the documents and in addition a Fava extension that changes where Fava stores uploaded documents?

yagebu:
> Is it possible to do this with an extension? If Fava relies on Beancount for finding the documents, does that mean I would need a Beancount plugin to help Beancount finding the documents and in addition a Fava extension that changes where Fava stores uploaded documents?

It is not yet possible to do this with an extension, however an extension hook to allow for an extension to determine the document location could be added to Fava. And yes, you would also need a Beancount plugin to find those documents and generate Document entries for them.

PR #1326, opened 2021-09-06
Documents directory structure

Fjanks:
As mentioned in #1293, it would be nice to have a choice where the uploaded documents are saved. I implemented it by adding the new fava-option "documents-directory-structure". There are four methods so far (by-account, by-account-root, by-fileprefix, none), where "by-account" is the method that was used before and which remains the default method.

### DIFF
--- a/src/fava/core/documents.py
+++ b/src/fava/core/documents.py
@@ -1,6 +1,7 @@
 """Document path related helpers."""
 import os
 from os import path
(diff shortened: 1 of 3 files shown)


yagebu:
Closing this PR. As written in the linked issue, I think this customization should be enabled with an extension hook and not an option.

PR #1352, opened 2021-12-18
Update development.rst: Add extra commands

mhansen:
These were some commands that I had to google how to install.

### DIFF
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -14,8 +14,11 @@ the tests, you will need `tox` and to run the linters you will need
     git clone https://github.com/beancount/fava.git
     cd fava
     # using a virtual environment is optional, but recommended
+    pip install --user virtualenv
     virtualenv -p python3 venv
     . venv/bin/activate
+    pip install tox
+    pip install pre-commit
     make
     pre-commit install  # add a pre-commit hook to run linters
     pip install --editable .


yagebu:
It already says that these tools are required above the code block and I'd rather keep it that way, since for most users it is probably preferable to install those tools with their platform package manager instead of pip.

mhansen:
I wasn’t sure which package manager they’d be (had to google it, at first I
thought they’d be npm, and some of the names like “tox” were hard to google
with irrelevant results coming up).

 I thought if we’re using virtual envs and these are python modules we
would prefer to use the venv over the OS package manager, in case other
programs need other versions? And the OS package managers are often out of
date? I’m not very familiar with norms on python projects though.

At any rate I’d like to make it a bit clearer how to install these deps.

yagebu:
Thanks for the feedback. I'll try to expand the paragraph before the code block a bit to explain how these tools can be installed and link to their homepages so that it's clear which tools are meant. tox and pre-commit are both rather compatible since they're both just script runners, so installing from package managers is fine (and should ensure that you don't have an outdated version in your venv a year from now).

mhansen:
Thanks, sounds good! By the way, once I got these tools set up, it was a
nice development experience with quick reload to see the changes, cheers.

mhansen:
Thanks! Looks good

PR #1355, opened 2022-01-10
allow negative account values in treemap/hierarchy

ahaldane:
Hi, 

This PR handles negative accounts in the treemap visualization. 

Let me know if you think it's useful, and I can do any extra tidying or add some unit tests, if you can help point me to what is needed. Much appreciated.

## Description of problem this solves: 

The treemap visualization for expense and income (and treemaps in general) can't handle negative values, which can appear in those accounts. That is, it renders n overlapping mess of boxes.

Example of problem:

 * I have accounts `Expenses:Taxes:TY2020` and `Expenses:Taxes:TY2021`
 * I want to visualize a beancount file that only covers transactions in 2021. 
 * In 2021 I payed $1000 of TY2021 taxes in paystubs (`balance Expenses:Taxes:TY2021 1000 USD`, and I got a $200 refund for 2020 taxes (`balance Expenses:Taxes:TY2020 -200 USD`). 
 * This means that the parent node `Expenses:Taxes` has a net balance of `800 USD`. 

So, when d3 draws the treemap it allocates an area of 800 units to cover `Expenses:Taxes`, but within that area it tries to put an area 1000 for `Expenses:Taxes:TY2021`, and it does not show `Expenses:Taxes:TY2020` at all. What ends up happening is the area of size 1000 overflows past the parent box and covers/hides the neighbor areas in the treemap, making the treemap ugly/illegible.

## My proposed fix

Show negative areas in hash coloring. See image where the "Temple" account is negative:

demo_treemap

  * Pro: The treemap has no jumpled overlaps or hidden accounts which made it difficult to read.
  * Con: Negative accounts now add to the total area, which could be confusing. I.e, the percentages in the tooltips no longer add to 1.

I expect that opposite-sign subaccounts for income/expenses are pretty rare/small, so i thought the pro outweighs the con.

## Implementation

This is the first time I've really used typescript and reactjs, and I only vaguely know d3, so suggestions appreciated. I couldn't figure out how to annotate types in the `reduce` calls to satisfy the lint check.

I added and extra "cumulbalance" field to the `AccountHierarchyDatum` object. That seems ugly, but I couldn't figure out a better way to communicate both the "absolute area (value)" and the "net balance" of each account from the hierarchy code to the treemap code.  That is, previously the code just set the node.value field to equal the total account value, but in this PR I needed node.value to equal the absolute-value-sum of all descendant accounts (to use as he box area in treemap), but I still needed to communicate the account net balance to display in the tooltips.

## Alternatives (edit)

A simpler alternative to this PR is to only include positive subaccounts. That's a one-line change, see my alternate branch https://github.com/ahaldane/fava/tree/treemap_positive_only last commit.

### DIFF
--- a/frontend/src/charts/Treemap.svelte
+++ b/frontend/src/charts/Treemap.svelte
@@ -31,12 +31,20 @@
     return $treemapScale(node.parent.data.account);
   }
 
+  function mask(d: AccountHierarchyNode) {
+    const node = d.data.dummy && d.parent ? d.parent : d;
+    if (node.data.cumulbalance < 0) {
+        return "url(#treemapHatch-mask)";
+    }
+    return null;
+  }
+
   function tooltipText(d: AccountHierarchyNode) {
-    const val = d.value ?? 0;
-    const rootValue = root.value || 1;
+    const val = d.data.cumulbalance ?? 0;
+    const refValue = root.data.cumulbalance || 1;
 
     return `${$ctx.currency(val)} ${currency} (${formatPercentage(
-      val / rootValue
+      val / refValue
     )})<em>${d.data.account}</em>`;
   }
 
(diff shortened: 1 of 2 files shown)


yagebu:
Thank you for this PR, in particular for the detailed and well-written description and reasoning for this change!

I also like the other solution better though, so I'll merge your alternative PR instead :)

PR #1356, opened 2022-01-10
only include positive accounts in treemap

ahaldane:
This is an alternative to #1355. Sorry for double post noise.

Again, the problem it solves is that treemaps can't handle negative sub-balances which  sometimes occur. Unlike the other PR, this one simply removes any negative accounts from the treemap.

After playing with both solutions I actually like this one better, so please close the other PR if you agree.

This solution is both simpler to implement, and also makes the treemap plot less confusing.


yagebu:
> After playing with both solutions I actually like this one better, so please close the other PR if you agree.

I think I also like this one better

yagebu:
Thank you for both PRs :)

PR #1361, opened 2022-01-13
use more modern type annotations

yagebu:

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -7,7 +7,9 @@ repos:
       rev: v2.6.0
       hooks:
           - id: reorder-python-imports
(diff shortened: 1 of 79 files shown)


lutorm:
I'm trying to install fava for the frist time and, after pip3-installing it, I get "SyntaxError: future feature annotations is not defined" when I try to run it. This is with python 3.6.8 on a CentOS 8 machine.

yagebu:
Fava has dropped support for Python 3.6 (which has reached its EOL) with version v1.21 (see https://github.com/beancount/fava/blob/67852e3f504758807ce504aa895b132d70172bfb/CHANGES). You'll need a more recent version of Python to run Fava.

PR #1362, opened 2022-01-15
Correct HTML in journal table

mechanarchy:
Fix regression since a06a60d3c2b33e1007ac5ab22497844bba7f5b71, which broke the formatting for metadata key `document`. The generated HTML will have two closing `</a>` tags; this PR fixes that.


yagebu:
Thanks :)

PR #1366, opened 2022-01-16
Add support for the query 'clear' command

### LINKED ISSUE #1365: How to clear fava query history
Is it possible to clear the query history in fava web gui? I tried to type `clear` and submit but nothing happens. I also tried to clear site usage, but it keeps the same. I also read *Fava's Features* in http://localhost:5000/beancount/help/features, but didn't notice is it possible.

image

mechanarchy:
Entering `clear` in the Query page will now clear all query history. This fixes #1365 

I have had trouble with the pre-commit hook on my machine as I am currently stuck with old versions of node/npm, so eslint is not fully operable.

### DIFF
--- a/frontend/src/query/Query.svelte
+++ b/frontend/src/query/Query.svelte
@@ -55,7 +55,8 @@
     const query = query_string;
     if (!query) {
       return;
-    } else if (query.trim().toUpperCase() === "CLEAR") {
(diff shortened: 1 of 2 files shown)


yagebu (approved):
Thanks, this LGTM :)

PR #1368, opened 2022-01-16
Fix exact matches not showing in the autocomplete list

### LINKED ISSUE #1359: 'Go to account' box in the sidebar does not work for exact matches
Typing in the exact name of an account and pressing <kbd>Return</kbd>/<kbd>Enter</kbd> does not allow you to view the account. For example, on the [Fava demo instance](https://fava.pythonanywhere.com/example-beancount-file/income_statement/), type `Expenses:Food:Restaurant` into the 'Go to account' box and press the enter key. This was annoying for me when copy/pasting account names from my beancount files in a text editor, to view them in Fava. 

I believe this is because the `AccountSelector` component only navigates based on the selected item in the dropdown list, not the content of the textbox, and the dropdown code specifically returns an empty list when there is an exact match:
https://github.com/beancount/fava/blob/16fcbe7d40ad578eed494b1eb4ef902f22ea248a/frontend/src/AutocompleteInput.svelte#L47-L48

This patch fixes the behaviour for exact matches; I have also removed the `<span>` CSS height rule as it was clipping the text when showing an exact match.
```patch
diff --git a/frontend/src/AutocompleteInput.svelte b/frontend/src/AutocompleteInput.svelte
index 4ed7148f..8dd75cf2 100644
--- a/frontend/src/AutocompleteInput.svelte
+++ b/frontend/src/AutocompleteInput.svelte
@@ -44,8 +44,7 @@
         suggestion,
         innerHTML: fuzzywrap(val, suggestion),
       }));
-    filteredSuggestions =
-      filtered.length === 1 && filtered[0].suggestion === val ? [] : filtered;
+    filteredSuggestions = filtered;
     index = Math.min(index, filteredSuggestions.length - 1);
     if (selectFirst && index < 0) {
       index = 0;
@@ -178,7 +177,6 @@
   }
 
   li :global(span) {
-    height: 1.2em;
     padding: 0 0.05em;
     margin: 0 -0.05em;
     background-color: var(--autocomplete-match);

```

mechanarchy:
This fixes #1359

### DIFF
--- a/frontend/src/AutocompleteInput.svelte
+++ b/frontend/src/AutocompleteInput.svelte
@@ -44,8 +44,7 @@
         suggestion,
         innerHTML: fuzzywrap(val, suggestion),
       }));
-    filteredSuggestions =
-      filtered.length === 1 && filtered[0].suggestion === val ? [] : filtered;
+    filteredSuggestions = filtered;
     index = Math.min(index, filteredSuggestions.length - 1);
     if (selectFirst && index < 0) {
       index = 0;
@@ -178,7 +177,6 @@
   }
 
   li :global(span) {
-    height: 1.2em;
     padding: 0 0.05em;
     margin: 0 -0.05em;
     background-color: var(--autocomplete-match);


yagebu:
Hey, thanks for the PR :) I want to solve this a bit differently: I think having a different logic (automatic selection of the first account in the dropdown) was a mistake and it should work like other dropdowns in Fava (and most other places) is better. See #1371, which now just passes through an "Enter" if no account is selected. This also allows one to pick accounts that are not explicitly declared.

Issue #1338, opened 2021-10-14
Sort order in Journal view is not in URL and therefore not preserved through browser navigation


nzbart:
When in the journal view, it's possible to sort by columns such as "Naration/Payee". However, when changing the view - such as by clicking on a specific Payee to "drill in" - this sort order is lost when coming back to the previous page using the browser back button or by clicking the `x` on the filter edit field.

I think that this would not be a problem if the sort order was stored in the URL as with the filter criteria.

yagebu:
A PR for this would be welcome :)

nzbart:
@yagebu I'll see if I can find time to contribute. I'm no Python expert, and it's a big job to get up to speed with an unfamiliar project's toolset and way of doing things (coding standards, tests, etc.) - even if the change itself isn't huge.

PR #1375, opened 2022-01-24
fava_options: reverse entries

alycda:
I prefer that entries not be reversed, so I added an option to toggle this without the need to resort the table after page is loaded.

usage:
```
2022-01-24 custom "fava-option" "reverse_entries" "false"
```


yagebu:
Hey, thanks for the PR :)

This is not something that I want to add an option for. Storing the sort order in the URL as suggested by #1338 instead or putting it in localStorage would be a better solution in my opinion.

Issue #1021, opened 2019-11-26
Transactions with forecasting feature not shown on fava


mckelvin:
The forecasting feature is implemented via the [`beancount.plugins.forecast`](https://bitbucket.org/blais/beancount/src/default/beancount/plugins/forecast.py) plugin. When the plugin is enabled, transactions like

```
  2014-03-08 # "Electricity bill [MONTHLY]"
    Expenses:Electricity 			50.10 USD
    Assets:Checking			       -50.10 USD
```

are not shown in fava however it works correctly with bean-web.

yagebu:
What do you mean by "not shown" - in the Journal or are they also not reflected in the balances? By default, the journal only diplays cleared and pending transactions, have you tried showing all transactions by toggling the `x`?

mckelvin:
> have you tried showing all transactions by toggling the x?

It works now after clicking the `x` in the `Journal` page.

```
option "operating_currency" "USD"

plugin "beancount.plugins.forecast"

2001-01-01 open Expenses:Electricity USD
2001-01-01 open Assets:Checking USD

2014-03-08 # "Electricity bill [MONTHLY]"
  Expenses:Electricity 			50.10 USD
  Assets:Checking			       -50.10 USD

```

Thanks for the help!

PR #1376, opened 2022-01-25
Forecast Plugin

### LINKED ISSUE #458: Add support for scheduled transactions
Now that we're going into the direction of actually changing the Beancount source file (with the form-based entry feature), I'd like to suggest adding support for scheduled transactions.

This is re: a discussion on the [Beancount mailing list](https://groups.google.com/forum/#!msg/beancount/pcZI0xEnps8/ykloTZQ2FAAJ) on the topic of GnuCash -> Beancount conversion. See also https://github.com/henriquebastos/gnucash-to-beancount/issues/1

Apparently, [GnuCash](https://www.gnucash.org/docs/v2.6/C/gnucash-guide/txns-sxn1.html) supports scheduled transactions by showing a dialog when the schedule occurs, with all the data pre-filled, so it is a one-click operation for the user to create the instance of the scheduled transaction.

This would save a lot of work with transactions such as monthly utilities, rent, salary, etc.

I suggest we:

- [ ] Figure out a way to record the necessary data for the schedule
- [ ] Enhance the current form-based transaction creation, so we can pre-fill it with data, which the user can confirm with one click (after optionally amending it)
- [ ] Add support to show a notification when a schedule occurs (on and after the scheduled date, until an instance has been recorded)
- [ ] Coordinate with the @henriquebastos (and authors of similar tools) so we use the same syntax. This will make GnuCash scheduled transactions imported using these tools work automatically in Fava.

My initial thought is to use the first (or perhaps last?) occurrence of the scheduled transaction as the template for the new one. We could use a link (or tag) to link all occurrences together. To record the data about the schedule (e.g. start-date, end-date, repeat-every-x), I'm leaning toward using a `custom` with the necessary data as meta-data. We would link that to the occurrences using the same link (or tag)

alycda:
I have "finished" the forecasting plugin based on previous code I found.

usage:
```
plugin "plugins.forecast" "{'years':5}"

2022-01-05 # "Groceries"
  frequency: "weekly"
  interval: 2
  Liabilities:Credit
  Expenses:Household:Groceries                          100 USD
```

resolves #458 , #1021  

-----

I do a lot of manual entries in the editor. My current workflow to "clear" one of these transactions is to simply 

1. copy the entire entry
2. change the date to the next occurrence 
3. paste the copied entry, changing the flag (and removing the metadata).  
 
I know this could easily be handled by a simple button, but I wanted to share since this plugin has been exactly what I've wanted from all previous budgeting apps I've ever used.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -58,6 +58,7 @@ install_requires =
     markdown2>=2.3.0,<3
     ply
     simplejson>=3.16.0,<4
+    DateTime>=4.3
 
 [options.entry_points]
 console_scripts =
--- a/src/fava/plugins/forecast.py
+++ b/src/fava/plugins/forecast.py
@@ -1,0 +1,130 @@
+"""Adding a forecasting feature to Beancount via a plugin
+
+see https://github.com/beancount/beancount/blob/ab3fdc613fd408e5f6d8039b2fe7eb37c0b31a5e/experiments/plugins/forecast.py
+also https://github.com/beancount/beancount/blob/d841487ccdda04c159de86b1186e7c2ea997a3e2/beancount/parser/lexer.l#L127-L129
+
+This entry filter plugin uses existing syntax to define and automatically
+insert future transactions based on a convention.
+
+A User can create a transaction like this:
+
+    plugin "fava.plugins.forecast" "{'years':5}"
+
+    2022-01-24 open Assets:Checking USD
+
+    2021-02-01 # "Rent"
+        frequency: "monthly" ; required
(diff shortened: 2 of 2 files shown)


yagebu:
I don't want to ship a lot of plugins with Fava (only the ones that tie in with some of its core functionality), so I'll close this PR. With some tests, this would be welcome over at https://github.com/beancount/fava-plugins

fdw:
Hi @alycda, I'd be interested in something like this. Do you plan to add it to fava-plugins?

Issue #1377, opened 2022-01-29
Don't suggest closed accounts when adding a transaction


fdw:
When I add a transaction, I'm still suggested similarly-named accounts that have been closed for quite a while. As this results in an invalid transaction anyway, maybe it would be nicer to not suggest these accounts at all.

Fava version: `1.20.2.dev0+gb49d0540.d20211019`

yagebu:
This has come up before in #492. Since the frontend part is more advanced nowadays, this might be more feasible to implement nowadays. A PR would be welcome :) Note that we can't just filter out closed accounts but we rather need to filter out accounts that were closed before the date of entry that is currently being added.

fdw:
Ah, sorry, I had searched but didn't find that issue :(

Do you have a pointer where/how this might be implemented?

yagebu:
No problem :)

We basically need a mapping of account names to close dates in the frontend. From the Python side we could supply this in `fava.core.attributes` via `_layout.html` to the frontend, which would be loaded in `frontend/src/stores/index.ts`. This mapping could then be used in the `Transaction.svelte` component to filter the account suggestions.

PR #1389, opened 2022-02-12
Don't suggest closed accounts

fdw:
With this change, closed accounts will not be suggested anymore for transactions.

I hope this is okay, but please tell me if I've made a mistake 🙂
I'm also wondering if it makes sense to have `accounts` and `closing_dates` separately, or if `accounts` should maybe be a dictionary with account info as its value. This way, only one variable would be necessary, and it can easily be extended in the future.

Also, a huge thank you to @yagebu for your help!

Issue: #1377

### DIFF
--- a/frontend/src/entry-forms/Transaction.svelte
+++ b/frontend/src/entry-forms/Transaction.svelte
@@ -4,14 +4,14 @@
   import { emptyPosting, Transaction } from "../entries";
   import type { Posting } from "../entries";
   import { _ } from "../i18n";
-  import { account_details, accounts, payees } from "../stores";
+  import { payees } from "../stores";
 
   import AddMetadataButton from "./AddMetadataButton.svelte";
   import EntryMetadata from "./EntryMetadata.svelte";
   import PostingSvelte from "./Posting.svelte";
 
   export let entry: Transaction;
-  let suggestions: string[];
+  let suggestions: string[] | undefined;
 
   function removePosting(posting: Posting) {
     entry.postings = entry.postings.filter((p) => p !== posting);
@@ -21,18 +21,12 @@
     entry.postings = entry.postings.concat(emptyPosting());
   }
 
-  $: suggestions = $accounts.filter(
-    (value) => $account_details[value].close_date >= new Date(entry.date)
-  );
-
   $: payee = entry.payee;
   $: if (payee) {
-    suggestions = [];
+    suggestions = undefined;
     if ($payees.includes(payee)) {
       get("payee_accounts", { payee }).then((s) => {
-        suggestions = s.filter(
-          (value) => $account_details[value].close_date >= new Date(entry.date)
-        );
+        suggestions = s;
       });
     }
   }
@@ -98,6 +92,7 @@
       bind:posting
       {index}
       {suggestions}
+      date={entry.date}
       add={addPosting}
       move={movePosting}
       remove={() => removePosting(posting)}
--- a/frontend/src/entry-forms/Transaction.svelte
+++ b/frontend/src/entry-forms/Transaction.svelte
@@ -4,14 +4,14 @@
   import { emptyPosting, Transaction } from "../entries";
   import type { Posting } from "../entries";
   import { _ } from "../i18n";
-  import { payees } from "../stores";
+  import { account_details, accounts, payees } from "../stores";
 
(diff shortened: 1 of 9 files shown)


src/fava/core/attributes.py
```diff
@@ -71,6 +73,11 @@ def load_file(self) -> None:
         currency_ranker = ExponentialDecayRanker()
         payee_ranker = ExponentialDecayRanker()
 
+        self.closing_dates = {
+            account: self.ledger.accounts[account].close_date
```
yagebu, line 77: sorry, I had forgotten about FavaLedger.accounts - since we have it already, let's just pass that to the frontend (maybe as account_details) and avoid the work here

frontend/src/entry-forms/Transaction.svelte
```diff
@@ -21,12 +21,18 @@
     entry.postings = entry.postings.concat(emptyPosting());
   }
 
+  $: suggestions = Object.keys($closing_dates).filter(
```
yagebu, line 24: we should filter the (ranked) array of accounts here

yagebu (changes requested):
Thanks for the PR :)

I had forgotten about FavaLedger.accounts - I think we can just pass that to the frontend. I think we want the accounts as both a list/array (which is ranked) and as a dict/object (which contains metadata, close date, etc.)

frontend/src/entry-forms/Transaction.svelte
fdw: You mean using `$accounts` instead of `$closing_dates`? I tried that first, but had problems getting it to work.
However, if we now want to use `$account_details`, should I still start with `$accounts`?

src/fava/core/attributes.py
fdw: Alright. Then I think I need a pointer how to implement correct serialization for `AccountData` - the naive way didn't work 😉

frontend/src/entry-forms/Transaction.svelte
yagebu: Yes to preserve the ranked order of `$accounts`, we should still start with `$accounts` and filter that to all non-closed accounts (with the help of `$account_details`)

fdw:
Alright, I've added the changes. Thanks again :)

One more idea/question: Do you keep the opening dates for accounts somewhere? If they were part of `AccountDetails`, we could also ignore accounts that aren't open yet. (But maybe in a separate PR 😉)

src/fava/core/attributes.py
```diff
@@ -71,6 +74,11 @@ def load_file(self) -> None:
         currency_ranker = ExponentialDecayRanker()
         payee_ranker = ExponentialDecayRanker()
 
+        self.account_details = {
```
yagebu, line 77: We don't need this copy here, we can just pass the ledger.accounts dictionary to the frontend.

frontend/src/entry-forms/Transaction.svelte
```diff
@@ -21,12 +21,18 @@
...
   }
 
+  $: suggestions = $accounts.filter(
+    (value) => $account_details[value].close_date >= new Date(entry.date)
+  );
+
   $: payee = entry.payee;
   $: if (payee) {
-    suggestions = undefined;
+    suggestions = [];
```
yagebu, line 30: This is not ideal, since it completely erases the suggestion array during the fetch (or if the fetch fails)

frontend/src/entry-forms/Transaction.svelte
```diff
@@ -21,12 +21,18 @@
     entry.postings = entry.postings.concat(emptyPosting());
   }
 
+  $: suggestions = $accounts.filter(
+    (value) => $account_details[value].close_date >= new Date(entry.date)
```
yagebu, line 25: This does the string->date conversion for every account, which is unnecessary, doing so once before the loop would be more efficient.

yagebu (commented):
Some more comments, but I'll just address them during the merge (and move the filter logic to the account input component so that we can also use it for other entry types).

frontend/src/entry-forms/Transaction.svelte
yagebu: Also, `$account_details[value]` might be undefined.

yagebu (approved):

fdw:
Awesome, thank you very much! 🙂

PR #1408, opened 2022-04-08
Stacked barcharts(Take 2)

PhracturedBlue:
This is a new attempt to get stacked barchats merged into Fava.
It is heavily based upon the work by @jeffwheeler in #1286 

I have taken a slightly different approach in the following ways:
* ~Only the largest 'n' (n=10) accounts will be split.  All others will be dumped in an 'Other' bucket.~  I reverted this as highlighting seems to be sufficient, simplifies theinterface, and provides more information
* Multiple currencies are supported
* Both the original single bar as well as stacked bars are available and user selectable
* ~The legend only contains currencies, not sub-accounts~ There is no legend for stacked-bar graphs.  The currency legend makes no sense, and a per-account legend is too big to be useful
* Budgets are not supported for stacked bars.  If any of the stacked accounts have credits, then the sum of the stacked bars will not be the same as the sum of the values (because some bars will be negative cancelling out parts of the positive bars).  I  couldn't think of a good way to visualize budgets in this context

The following work is still to do:
- [x] Update tests to pass
- [x] Fix lint issues
- [x] Fix hover highlighting on stacked bar
- [x] Allow localizing 'Others' bucket
- [x] Allow configuring the bucket count
- [x] Support clicking an bar-section to jump to a specific sub-account
- [x] Use a custom color selection rather than lifting the sunbust code
- [x] Decide what to do with legend for multiple currencies

I would like feedback on if I'm taking the right approach. I don't often program in js/ts and have never worked with svelte before.

### DIFF
--- a/frontend/src/charts/BarChart.svelte
+++ b/frontend/src/charts/BarChart.svelte
@@ -204,11 +204,7 @@
                     highlighted = "";
                   }}
                   use:followingTooltip={() => tooltipText($ctx, bar.data, name)}
-                  on:click={() => {
-                    if (!name.startsWith(":")) {
-                      router.navigate(urlForAccount(name));
-                    }
-                  }}
+                  on:click={() => router.navigate(urlForAccount(name))}
                 />
               {/if}
             {/each}
--- a/frontend/src/charts/BarChart.svelte
+++ b/frontend/src/charts/BarChart.svelte
@@ -1,15 +1,25 @@
 <script lang="ts">
-  import { extent, merge } from "d3-array";
+  import { filter, max, merge, min, sum } from "d3-array";
   import { axisBottom, axisLeft } from "d3-axis";
-  import { scaleBand, scaleLinear } from "d3-scale";
+  import { scaleBand, scaleLinear, scaleOrdinal } from "d3-scale";
+  import { stack, stackOffsetDiverging } from "d3-shape";
+  import type { Series } from "d3-shape";
   import { getContext } from "svelte";
   import type { Writable } from "svelte/store";
 
   import { ctx } from "../format";
+  import { urlForAccount } from "../helpers";
+  import router from "../router";
+  import { barChartMode } from "../stores/chart";
 
   import Axis from "./Axis.svelte";
-  import type { BarChart, BarChartDatumValue } from "./bar";
-  import { currenciesScale, filterTicks, setTimeFilter } from "./helpers";
+  import type { BarChart, BarChartDatum, BarChartDatumValue } from "./bar";
+  import {
+    currenciesScale,
+    filterTicks,
+    hclColorRange,
+    setTimeFilter,
+  } from "./helpers";
   import { followingTooltip } from "./tooltip";
 
   export let data: BarChart["data"];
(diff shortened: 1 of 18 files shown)


PhracturedBlue:
Example with multiple currencies:
image

yagebu:
Thanks for picking this up!

Overall, this looks good :) Could you add some tests for this functionality (on both the Python and the JS side of things)?

One thing I'm not sure about yet is the way that accounts are grouped into the "other" bucket. I'd want to show as many accounts as possible and that would depend more on the relative account balance than just on the account being one of the 10 largest ones. For example, I could have an account, where one child account has 99% - then showing the next 9 accounts separately is pointless, whereas in another account I might have 20 equally sized child accounts that could all be shown. So I think it would be better to do this in the frontend and only group accounts into the "other" bucket that are too small to be visible.

PhracturedBlue:
Thanks for looking at it.
I'll look into adding some tests.  

I disagree that I want to see as many bars as possible.  With too many bars , the colorspace becomes too big, and I can't distinguish the colors.  this is the primary reason for limiting the bar-count.  If you set the bar-count larger:
```
2000-01-01 custom "fava-option" "num-stackedbar-accounts" "100"
```
The display works fine without any filtering, but I don't find the results pleasing.  Maybe a better colorspace selection could help with this, but I was unsuccessful in finding something better.

I don't think doing the pruning in the frontend provides any additional value, but if you prefer it, the function can be implemented there just as easily.
You need to either prune before you've scaled the data (which is really no different than doing it in the backend), or prune after you've done the stacking (in which case you can't group the data together, because there is no guarantee that small items are together...this is basically the same as setting 'num-stacked-accounts to a large number.  I guess another option would be to scale the values, do pruning, and then stack the scaled values.  That might be an interesting approach, though again, I don't think it will really address the problem I was trying to address differently than doing it at the beginning as I am now.

A more interesting idea to me was to group accounts into their parent group to make them big enough (i.e. If I had a bunch of Expenses:CreditCard:Dining, Expenses:CreditCard:Clothing, Expenses:CreditCard:HouseSupplies, that were below a threshold, to group these into Expenses:CreditCard).  The problem was coming up with a heuristic that actually worked well for this, since not all accounts are collapsable this way.  This could be paired with the ability to 'zoom-in' to the graph, to allow expanding categories, but I'm not sure what the interface to that would look like.  It is basically just assigning a filter, and rather than complicating the interface, maybe it is better to just use the existing filter interface for this.

PhracturedBlue:
After further consideration, I think we can just remove the filtering altogether.  When I initially implemented it, I didn't have the hover-highlighting working.  But with hover highlighting, I can easily see a given account across time (which was the whole point of the stacked-bars in the 1st place).  Here is a real-life example:
image

and with highlighting:
image

PhracturedBlue:
Is there anything additional I can do to get this pull request to move forward?

jeffwheeler:
This patch is great, I'd also like to have it merged

yagebu:
> Is there anything additional I can do to get this pull request to move forward?

No, this looks pretty good already and I just want/need to find the time to review it again and do some minor refactorings.

yagebu:
Merged - thanks again :)

jord1e:
I just want to thank you and @jeffwheeler , awesome feature!

(sorry for the notification)

PR #1409, opened 2022-04-10
Fix position of </p> in journal table

upsuper:
I believe the `</p>` is misplaced. The corresponding `<p>` seems to start way before the `{% if %}` block, so there is no reason the `</p>` should be inside this block.


yagebu:
Thanks :)

Issue #1069, opened 2020-03-07
Price column in Journal page does not follow precision of price in beancount file


harishcm:
It appears that the Price column in Fava's Journal page does not follow the precision of the price provided in the beancount file.

Below is a sample showing a beancount file transaction with a price attached to the posting to 4 decimal places.

![image](https://user-images.githubusercontent.com/159447/76147369-93f81b80-60d6-11ea-973c-80a7d15b443c.png)

The Price column in the Journal page however only shows the price to 2 decimal places.

![image](https://user-images.githubusercontent.com/159447/76147379-b25e1700-60d6-11ea-8afa-3850e8620ef8.png)

Is the above display behaviour the expected one? It differs from bean-web, shown below

![image](https://user-images.githubusercontent.com/159447/76147723-32d24700-60da-11ea-8736-62e875687efc.png)

The main issue is that the display of the Journal page in Fava makes it appear as though the price doesn't tally with the other legs of the transaction, i.e 

37.00 x 1.55 = **57.35** which does not round off to **57.23**.

I also attach beancount's dcontext option reported by Fava below. The above behaviour is the same in both Firefox and Chromium. Fava version is 1.14. Beancount version is 2.2.3.

![image](https://user-images.githubusercontent.com/159447/76147533-4c728f00-60d8-11ea-9387-30a109679b7f.png)
 
Many thanks.

yagebu:
> Is the above display behaviour the expected one? 

It is the currently expected behaviour. My reason for using the display context here was that I mainly use the `@` syntax with a missing price - the price is then inferred, which usually leads to numbers with lots of decimal places.

We could show the full price as a tooltip on hover maybe?

harishcm:
I see the problem of lots of decimal places in bean-web when using the `@` syntax with a missing price.

Could I suggest using the `fractional_max` part of the display context for determining the decimal places for the price column? 

I notice that `fractional_max` adjusts to the decimal places of the price input into the beancount file, i.e.

`37.00 EUR @ 1.5467 SGD` results in `fractional_max=4` for SGD 
`37.00 EUR @` results in `fractional_max=2` for SGD

Alternatively, using the exact price written in the beancount file may be an option, but this could deviate from the display context.

PR #1416, opened 2022-04-18
Adds fava-option to configure number of digits after comma in Fava.

clsacramento:
Adds an option to set the precision display in Fava, keeping the default behavior of using Precision.MOST_COMMON from Beancount.
Related: https://github.com/beancount/fava/issues/1069


clsacramento:
@yagebu you seen to be the one that contributed mostly to fava. I was hoping if you could suggest someone to review this PR. Thanks in advance.

yagebu:
Hey, thanks for the PR. Allowing the user to override the display precision is a feature that Fava should have, however I think it would be best to implement this per-currency, allowing one to override the precision via metadata on the commodity - which I've implemented in #1435

PR #1426, opened 2022-06-08
Add option: precision-overwrite

WayneY:
Beancount outputs numbers with two modes of precisions and fava takes one of them to render the nubmers on webpages. But neither beancount nor fava give an option to set the user prefered precision. Since it is a display issue which has nothing to do with the calculation, I think it is better to add such an option for uses to make their own choices.

The option behaves like this:

Input as a syntax of   `Currency1:Precision1, Currency2:Precision2 ...`  (Yes obviously the currency symbol should not contain : or , )   to set the prefered precision currency by currency.  If one currency has been set, it will be rendered with the setted precision; if one currency has not been set, it will go as before.


IguanaBen:
Just to add to why this is useful, if you have a ledger file of
```
option "operating_currency" "USD"
option "inferred_tolerance_default" "USD:0.01"  ; This option makes no difference

2022-01-01 open Assets:Bank
2022-01-01 open Equity:Shareholder

2022-01-01 * "Opening"
  Assets:Bank	                            1000.00 USD
  Equity:Shareholder

;This line adds changes USD precision from two decimal places to four in the display of numbers
2022-01-02 price CAD 0.7755 USD;
```
Then all fava queries / reports, etc, return the four digits of precision on the USD because of that price entry. Below is an example using a query:

image

That same query in bean-query has 2 digits of precision:

```
$ bean-query example.beancount
Input file: "Beancount"
Ready with 4 directives (2 postings in 1 transactions).
beancount> select date,account,position,cost(position)
   date         account          position    cost_position
---------- ------------------ -------------- --------------
2022-01-01 Assets:Bank         1000.00   USD  1000.00   USD
2022-01-01 Equity:Shareholder -1000.00   USD -1000.00   USD

beancount>
```

Using this patch with the addition of the line 

```
2022-01-01 custom "fava-option" "precision_overwrite" "USD:2"
```

gets the queries in fava (and the reports, etc) to have two decimal places and match outputs of bean-query.

yagebu:
Thanks for the PR :) This is somewhat of a duplicate of #1416 (just a per-currency instead of a global precision override). This is something that Fava should have, however instead of adding a new fava-option, using commodity metadata would be preferable IMHO - see #1435

Issue #1229, opened 2021-02-04
Improve AddEntry modal on mobile


taimrli:
First of all thanks for this great frontend!

Something that bothers me a bit about using it is that the AddEntry form is really cumbersome to use on mobile.

Would there be a way to be able to have it also non-modal... like a 'normal' sub route? Perhaps an additional option?

Unfortunately, I have too little expertise in Svelte, let alone Python, to make myself useful in this...

yagebu:
On mobile, the modal should already fill most of the screen. How would making it its own page make it more easily usable?

taimrli:
The problem is, the modal fills _more_ than the screen... which leads to scrolling back an forth:
![Screenshot_20210206-110426_Brave](https://user-images.githubusercontent.com/18010667/107137997-d55e0000-6911-11eb-8cb5-cfc4e9b10e73.jpg)

yagebu:
I think this should be fixed by improving the CSS for the modal overlay on mobile, not by creating a separate page. Mobile browsers have some odd behaviours and mobile screen sizes are also not very consistent so that's not a trivial task. A PR improving the CSS for the modal would be welcome :)

yagebu:
Should be fixed by #1431 :)

PR #1431, opened 2022-06-21
Fix modal size on mobile

lukasdietrich:
Hey!

With this pull request I propose a fix for #1229.

First of all, I do not understand _why_ the browser behaves as it does.
The modal looks good on desktop and keeps looking good when resizing the browser window to mobile dimensions.
When using the developer tools to test mobile devices, however, you can activate touchscreen emulation, which breaks modals on desktop too.

Touchscreen Emulation

Using viewport units instead of `left: 0;` and `right: 0;` etc. to stretch the `.overlay` to the screen size, seems to be working fine, even on mobile.

I tested the patch on desktop Firefox as well as mobile Chrome and Firefox.

### DIFF
--- a/frontend/src/modals/ModalBase.svelte
+++ b/frontend/src/modals/ModalBase.svelte
@@ -55,6 +55,8 @@
 <style>
   .background {
     position: fixed;
+    top: 0;
+    left: 0;
     width: 100%;
     height: 100%;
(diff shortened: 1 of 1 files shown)


yagebu:
Thanks for the PR and putting in the work to get this fixed - this looks great :)

Issue #1436, opened 2022-07-02
Currency conversion select


ibigbug:
The current logic check if the val is an ISO currency: https://github.com/beancount/fava/blob/6ec42909454a882c70fec8097e72ff7071edd10c/frontend/src/stores/chart.ts#L37

This becomes interesting when a stock symbol is also a currency symbol:

https://github.com/beancount/fava/blob/6ec42909454a882c70fec8097e72ff7071edd10c/frontend/src/lib/iso4217.ts#L112
https://finance.yahoo.com/quote/NIO/

So `NIO` will show up in the conversion select:

<img width="205" alt="image" src="https://user-images.githubusercontent.com/543405/177014488-de5c5f69-97d6-44ff-bfe0-396dafdc08d5.png">

But it's not a "currency" in my current bookkeeping, but a stock that I hold.

Any ideas to hide the "stock" in the "currency" list?

yagebu:
This is a duplicate of #1226

Issue #1226, opened 2021-01-26
The "Converted to" List filter cannot distinguish currency and commodity


anels:
![image](https://user-images.githubusercontent.com/1177049/105902411-77592080-5fd3-11eb-9124-132a4c3c64ad.png)
When a symbol of a commodity (ticker) is the same as a currency symbol (e.g. 'AMD') it will be displayed on the "Converted to" list which is annoying.

In [the discussion on this issue](https://github.com/beancount/fava/issues/1041#issuecomment-581032924), it seems the filter is based on https://www.currency-iso.org/en/home/tables/table-a1.html. I think maybe fava should add an option to allow users to specify the *Currencies* to make it easier to manage.

yagebu:
I'm not too keen to add an option for this.

jeffwheeler:
I want to add that I was also very confused by this. I hold a variety of equities that happen to share names with currencies. I didn't realize that's why they were listed until I saw this bug report, thanks!

<img width="185" alt="Screen Shot 2021-05-11 at 8 58 16 PM" src="https://user-images.githubusercontent.com/14618/117916547-38a65800-b29c-11eb-89e5-26c824f9190f.png">

yagebu:
Well, I guess due to US ticker symbols commonly clashing with ISO currency names, that we should add an option to specify the list of conversion currencies. A PR would be welcome.

yagebu:
This option has been implemented in #1439

PR #1439, opened 2022-07-04
Added fava currency_conversion opt

ibigbug:
Related to #1436 #1226

Add a new fava opt: `currency_conversion` - feel free to suggest a better name please

If provided, the chart show currency conversion select will use the user provided, otherwise fallback to the current logic.

no validation on the option input, which means user can convert into commodities that are "non-currency".


ibigbug:
sounds good :)

could you point me where is the options help page doc located ?

yagebu:
> could you point me where is the options help page doc located ?

src/fava/help/options.md

yagebu (approved):
Thanks again! I've made some small changes to the help text

PR #1440, opened 2022-07-04
Bugfix: Mark extension generated content as safe

Anks:
I run Fava with a custom extension for my own reports. After the recent commit ca9e3882, it looks like the HTML generated by the extension's template is being escaped. 

This commit adds wraps the content with `Markup()` , similar to how [Help HTML](https://github.com/beancount/fava/blob/ca9e3882c7b5fbf5273ba52340b9fea6a99f3711/src/fava/application.py#L388) was explicitly marked safe.

### DIFF
--- a/src/fava/application.py
+++ b/src/fava/application.py
@@ -345,7 +345,9 @@ def extension_report(report_name: str) -> str:
         )
         content = render_template_string(template, extension=extension)
         return render_template(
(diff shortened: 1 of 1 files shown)


yagebu:
Thanks for the fix, merged in 6956476f0662664d88dd25c9d04ed7650fe90404 (I've only moved the Markup call up one line - the return value of render_template_string should be safe and I think it's easier to see this way).

PR #1441, opened 2022-07-04
Update setup.ts

newboydj169:
to override the TAB key behavior.


yagebu:
Thank you for opening this PR - could you explain the intention of this change?

newboydj169:
Previous version(at lease from v1.17) support using Tab inside the codemirror editor to indent.
But now you press Tab, the cursor will jump to broswer titlbe.

yagebu:
See https://codemirror.net/examples/tab/ for why this is the intentional default behaviour of CodeMirror. I'll add a note to the help pages and the enable the indent with tab behaviour.

PR #1468, opened 2022-08-07
Allow extensions flexibility in chart placement

redstreet:
Currently, the chart is a part of `_layout.html`. This means extensions like https://github.com/redstreet/fava_investor/ cannot place a menu bar above the chart. With this patch, extensions can place the chart as they see fit in their own templates.

### DIFF
--- a/src/fava/templates/_layout.html
+++ b/src/fava/templates/_layout.html
@@ -40,11 +40,11 @@
     <aside>{% include "_aside.html" %}</aside>
     <article>
       {%- endif %}
-      <svelte-component type="charts"></svelte-component>
       {% block content %}
       {% if content %}
       {{ content }}
       {% else %}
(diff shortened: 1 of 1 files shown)


redstreet:
@yagebu this should be a simply change that only affects extensions in a minor way. WDYT?

redstreet:
@yagebu gentle ping.

yagebu:
Since this would break all current extensions that use charts, this is not a change that I want to make lightly.

I think if we make this breakage, this should include some more changes. Currently, there can only be one chart box per page, since we create a global `chart_data` list that all the chart data gets pushed into (and then this implicitly fills the chart). I think a more explicit approach of having each report adding a `<svelte-component type=charts>` element with the desired data would be better.

yagebu:
See #1491

redstreet:
Excellent, thanks for the pointer!

PR #1470, opened 2022-08-21
account tree table: fix end date in links

andreasgerstmayr:
Beancount and Fava interpret the end date as exclusive internally. The Fava time filter interprets the end date as inclusive.

This commit updates the links in the account tree table (account -> changes) to generate a correct time filter.


yagebu:
Hey, thanks for the fix. However, I think it's even better if we just use the corresponding interval filter formats, see #1475

andreasgerstmayr:
Thanks! I've created a small follow-up PR in #1476 to also update the links in the table cells.

PR #1476, opened 2022-09-03
account tree table: use time filter intervals for table cells

andreasgerstmayr:
Use new time filter also for table cells (follow-up fix for #1475)


yagebu (approved):
Thanks!

PR #1486, opened 2022-10-08
Sort amount/inventory/position in query table as number, not string

xinluh:
The query table currently sorts amount, inventory and position types as strings (for example, see https://fava.pythonanywhere.com/example-beancount-file/holdings/ when sorting by the 5th column)
image

This PR makes columns of these types sort numerically instead. Obv. with these quantities that comes with a unit, it isn't  always directly comparable numerically without some conversion, but I think the numerical sort is at least more useable over string sort, especially when the column is of the same currency (which is not uncommon).


yagebu:
Thanks! Yes, this is probably a behaviour that is more useful. At some point, I'll move the query rendering to the frontend, then this could be improved to behave better when multiple commodities are involved.

PR #1491, opened 2022-10-13
charts: use svelte-component to explicitly add charts

yagebu:
Previously, there could only be one ChartSwitcher component on the page
since the charts were pushed to a global list.

This now adds an explicit component to render the charts.

### DIFF
--- a/frontend/src/charts/ChartSwitcher.svelte
+++ b/frontend/src/charts/ChartSwitcher.svelte
@@ -2,38 +2,64 @@
   import { onMount } from "svelte";
 
   import { bindKey } from "../keyboard-shortcuts";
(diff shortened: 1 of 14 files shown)


yagebu:
Extension authors that use charts: See https://github.com/beancount/fava/pull/1491/files#diff-27dba36bfd508e3fe9e3270378f71c0d5296144d632aacaaf1c4fd6ef870dd23 for an example of how to use the new charts.

redstreet:
Great, this is very helpful! fava_investor has been updated, and finally, the chart placement is not awkward any more in it :).  [Demo](https://favainvestor.pythonanywhere.com/example-beancount-file/extension/Investor/?module=aa_class).

PR #1503, opened 2022-11-05
Income statement graphs: re-add name of interval to graph labels

MJochim:
Currently, the labels below the income statement graphs are pretty confusing, because there are two graphs labeled "Income" and two labeled "Expenses". This was not always so: The current screenshot at [1] shows the graphs usefully labeled as "Income" vs. "Monthly Income" (ditto for Expenses). However, this seems to have been lost in commit a80ede86.

This time, I chose "Income (Monthly)" instead of "Monthly Income" because it makes translations easier. In German, for example, "Monatlich Einkommen" would be ungrammatical.

[1] https://beancount.github.io/fava/index.html


yagebu:
Thanks! :)

PR #1524, opened 2022-12-29
extensions: allow extensions to include script tags

derpferd:
Currently extensions that include javascript in script tags in the html have an issue where the scripts are only run some of the time (script doesn't run on a partial load but does run on a full refresh). This fixes makes the scripts run on both partial and full loads.

### DIFF
--- a/frontend/src/helpers.ts
+++ b/frontend/src/helpers.ts
@@ -44,3 +44,27 @@ export function urlForSource(file_path: string, line: string): string {
 export function urlForAccount(account: string): string {
   return urlFor(`account/${account}/`);
 }
+
+/*
+ * Set the inner HTML of an element in a way that will execute the contents of any script tags.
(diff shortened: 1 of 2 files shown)


yagebu:
Thanks for the PR, but this is not something I want to add - and with #1600 we have a PR also providing a solution for this :)

derpferd:
Checked out solution in #1600. This is a good solution for single javascript files however if you require third party javascript like I do, I would need to pre-download and compile my javascript and the third party into a single javascript file.

Example:

```
<script 'https://code.jquery.com/jquery-3.7.0.min.js'/>
<script 'Extensions/XYZ/main.js'/>
```

yagebu:
Bundling the code is one solution but you should also be able to use imports in your Javascript file to import any other Javascript (modules).

PR #1527, opened 2022-12-31
treeview

xiaoniu-578fa6bff964d005:
Similar to https://github.com/beancount/fava/pull/1304 and https://github.com/beancount/fava/pull/1137.

### DIFF
--- a/frontend/src/reports/editor/treeview.ts
+++ b/frontend/src/reports/editor/treeview.ts
@@ -1,0 +1,90 @@
+export interface FileNode {
+  name: string;
+  path: string;
+  action: () => void;
+}
+export interface FolderNode {
+  name: string;
+  subfolders: FolderNode[];
+  subfiles: FileNode[];
+}
+
+const zip: <A, B>(a: A[], b: B[]) => [A, B][] = <A, B>(a: A[], b: B[]) =>
+  a.reduce((l: [A, B][], k: A, i) => {
+    const n = b[i];
+    if (n !== undefined) {
+      l.push([k, n]);
+    }
+    return l;
+  }, []);
+
+function shorten_folder(folder: FolderNode): FolderNode {
+  if (folder.subfiles.length === 0) {
+    const subfolder = folder.subfolders[0];
+    if (subfolder !== undefined) {
+      const new_name = `${folder.name}/${subfolder.name}`;
+      return shorten_folder({
+        name: new_name,
+        subfolders: subfolder.subfolders,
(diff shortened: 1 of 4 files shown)


frontend/src/reports/editor/treeview.ts
```diff
@@ -0,0 +1,90 @@
...
+  path: string;
+  action: () => void;
+}
+export interface FolderNode {
+  name: string;
+  subfolders: FolderNode[];
+  subfiles: FileNode[];
+}
+
+const zip: <A, B>(a: A[], b: B[]) => [A, B][] = <A, B>(a: A[], b: B[]) =>
```
yagebu, line 12: zip from d3-array could probably be used here instead

frontend/src/reports/editor/treeview.ts
```diff
@@ -0,0 +1,90 @@
...
+    name: folder.name,
+    subfolders: folder.subfolders.map(shorten_folder),
+    subfiles: folder.subfiles,
+  };
+}
+function _source_tree(
+  paths: string[][],
+  filenodes: FileNode[]
+): [FileNode[], FolderNode[]] {
+  const groupBy = <T>(arr: T[], key: (i: T) => string) =>
```
yagebu, line 43: group from d3-array could probably be used here instead

frontend/src/reports/editor/treeview.ts
```diff
@@ -0,0 +1,90 @@
...
+    return { name: dir, subfolders, subfiles };
+  });
+  return [r1, r2];
+}
+export function source_tree(
+  files: string[],
+  goToFileAndLine: (filename: string, line?: number) => void
+): [FileNode[], FolderNode[]] {
+  files.sort();
+  const paths = files.map((file) => file.split("/"));
```
yagebu, line 80: This won't work for paths on Windows...

yagebu (changes requested):
Hi, thanks for having another go at this.

I've left some small comments but haven't had an in-depth look at the code yet. Could you add documentation and tests for the main logic in treeview.ts ?

xiaoniu-578fa6bff964d005:
Thanks for reviewing it and the comment. I hope I can take a look later this month.

frontend/src/reports/editor/treeview.ts
yagebu: I guess just splitting by forward and backward slashes might be good enough for our use case here

PR #1528, opened 2023-01-02
fix safari not updating values on autocomplete on enter

### LINKED ISSUE #809: Keyboard entry filtering doesn't work on safari.
Creating issue based on @victoraraujo01 's comment in https://github.com/beancount/fava/issues/768#issuecomment-403682834

When entering in a filter in safari (e.g. year), the entry doesn't seem to work when entered via keyboard. 
beancount_fava_safari_filter_bug

vanillacode314:
fixes #809

### DIFF
--- a/frontend/src/AutocompleteInput.svelte
+++ b/frontend/src/AutocompleteInput.svelte
@@ -71,6 +71,7 @@
       } else {
         dispatch("enter", input);
       }
+      dispatch("select");
     } else if (event.key === " " && event.ctrlKey) {
       hidden = false;
     } else if (event.key === "Escape") {


yagebu:
Hi thanks for the PR - can you give some reasoning on how you arrived at this change? It just seems to dispatch two events for one keypress - either select-select or enter-select. How did that fix the updating of the values in Safari?

It would seem surprising that having the select event twice would fix anything so my assumption would be that Safari does not properly handle the case where the component fires the enter event. In that case, the submit on the form should be called - maybe it isn't? I've tried to add another workaround (which just adds additionally treats the enter event on the input like the submit) - #1616. Does that fix it as well?

yagebu:
@vanillacode314: I've merged #1616 and will close this PR. It would still be great if you could check whether that fixes it (and why).

seltzered:
Took me a while to setup, but did a brief test of [6293d09](https://github.com/beancount/fava/commit/6293d09c216740a90f7d24d18ad4c6764c5812b8) on Safari 16.5 and #809 looks resolved

PR #1538, opened 2023-01-21
Budget calculations: option for "whole interval" under/over budget comparison

carderne:
## Background
The budget view in `/account/Expenses/changes/` shows green/red undershoot/overshoot indicators, that are calculated on a rolling basis based on how far you are through the interval (week/month/quarter/year).

This can be useful for things like food, which are evenly spread through the month, so you want to know how you're doing based on how far you are, but many things are bursty, and so you really just want an indication of how much of the budget you've used up.

## Example scenario
You set a `Month` budget for `Food` to `100 USD`.
On 15 January, you've spent `70 USD` in that category.

With the current behaviour, your budget for 1-15 January is calculated as `50 USD` andyou are `-20 USD` in the red.

With the "whole interval" behaviour, you are at `+30 USD` in the green.

## Changes
This PR adds an option called `budget-whole-interval` (need a better name), which, if set to `true` in the beancount file will change the behaviour as described above.

Linted and all tests pass with `SNAPSHOT_UPDATE=1 make test`.

## Alternative implementation
I haven't added any tests, or experimented thoroughly enough to know if this modification doesn't break any other uses. (Nor do I have a good enough mental model of what this could break.)

A more limited approach might be a change something like the below in [charts.py](https://github.com/beancount/fava/blob/74108b45bd2e5edbf20e876dc5486b6493d125e2/src/fava/core/charts.py#L191-L195):
```diff
diff --git a/src/fava/core/charts.py b/src/fava/core/charts.py
index 66e4866e..782af7e9 100644
--- a/src/fava/core/charts.py
+++ b/src/fava/core/charts.py
@@ -190,6 +190,7 @@ class ChartModule(FavaModule):
                 )
             budgets = {}
             if isinstance(accounts, str):
+               budget_end = get_next_interval(end, interval) if fava_options.budget_whole_interval else end
                 budgets = self.ledger.budgets.calculate_children(
-                    accounts, begin, end
+                    accounts, begin, budget_end
                 )
```

## Docs etc
Haven't added anything yet, if you're keen to include this (or something like it) then I will of course work on those.


carderne:
Have been using this for two weeks now with no issues. Only thing I can imagine is if you have future-dated stuff that you don't want showing up.

yagebu (changes requested):
I do not want to add an option for this, but I think the "whole interval" version could just become the new default, it seems to be the more consistent behaviour (since we do not really show to the user that the last interval might be "shorter").

I believe this implementation is not correct yet, since it only completes the last interval and not the first one, which might also be incomplete.

Also, there should be tests :)

carderne:
Have made it default to this behaviour, now it's a one-liner :).

Not sure I understand the first interval use case? I would need to either re-write this function https://github.com/beancount/fava/blob/9bc3a09fd915ed1b09bc4c378c4ef88285d60886/src/fava/util/date.py#L105-L107 but going backwards, or maybe just subtract an interval and then use that function.

Either way it's a bit of extra complexity, so wanted to check that you're keen for that.

carderne:
Ok have done as suggested, added a function `get_prev_interval` and used that to get the _real_ whole interval. It's possibly not the most elegant, but I rather kept it similar to `get_next_interval`.

Added some tests and updated the existing ones. Don't really understand the snapshots stuff, but have updated them and seems to make sense.

yagebu (approved):
Thanks, merged in 0452ad21

PR #1541, opened 2023-01-28
use only currencies in data for bar chart

### LINKED ISSUE #1534: "Changes" chart for accounts doesn't show anything if the account doesn't use any operating currency
In the account page, the "Changes" chart should probably not use operating currencies, but the currencies that the account holds.

upsuper:
This PR tries to fix #1534 by
* gathering the currencies used in the data, and
* using the most mentioned ones.

Operating currencies are still preferred when they are used in the data. We pick at most `max(5, len(operating currencies))` currencies in case the account has too many currencies used.

### DIFF
--- a/frontend/src/charts/bar.ts
+++ b/frontend/src/charts/bar.ts
@@ -1,3 +1,4 @@
+import { rollup } from "d3-array";
 import type { Series } from "d3-shape";
 import { stack, stackOffsetDiverging } from "d3-shape";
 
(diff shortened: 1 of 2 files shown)


frontend/src/charts/bar.ts
```diff
@@ -56,26 +56,66 @@ const bar_validator = array(
...
+    return map;
+  }, new Map<string, number>());
+
+  const toShow = [];
+  // Always take operating currencies if they are used in the data.
+  for (const currency of operatingCurrencies) {
+    if (inData.has(currency)) {
+      toShow.push(currency);
+      inData.delete(currency);
+    }
```
upsuper, line 84: I'm not sure whether the order of currencies is meaningful. If not, we can simply set infinity to all used operating currencies.

upsuper:
@yagebu mind taking a look at this PR?

frontend/src/charts/bar.ts
```diff
@@ -56,26 +56,66 @@ const bar_validator = array(
...
+/** Calculate the currencies to use for the chart. */
+function calculateCurrenciesToShow(
+  data: {
+    budgets: Record<string, number>;
+    balance: Record<string, number>;
+  }[],
+  operatingCurrencies: string[]
+): string[] {
+  // Count the usage of each currency in the data.
+  const inData = data.reduce((map, interval) => {
```
yagebu, line 68: This could be handled with a one-liner with the `rollup` function from `d3-array` (https://devdocs.io/d3~7/d3-array#rollup)

frontend/src/charts/bar.ts
yagebu: It's not very important but we should try to have the order be mostly consistent. Sticking to the order of the operating currencies makes sense IMHO

yagebu (changes requested):
Thanks for the PR, I like this change :)

Could you add some tests that cover this new functionality?

upsuper:
@yagebu I've updated the PR based on your comments. Mind taking another look?

upsuper:
@yagebu could you take another look?

yagebu (approved):
LGTM :)

Issue #1424, opened 2022-05-18
Export Fava as a static webpage


rougeth:
Hi all, I'm using Beancount to do accounting for a community around open source software. I have to deal with conferences, meetups, sponsorships, etc. We are going to publish all the .beancount files publicly on Github and I'd love to have an export of fava for read-only use (idea is a static webpage hosted in Github pages). Do you know a way of doing that? Maybe a plugin?

Cheers.

rougeth:
[Frozen-Flask](https://pythonhosted.org/Frozen-Flask/) might be an option, I'll check if that can be done as an extension.

yagebu:
Since Fava has many interactive parts that require data that needs to be calculated on-demand, exporting it as a static webpage is not really feasible

rougeth:
What about a read-only mode, an option for disabling import, editor, options and help?

daneoshiga:
One option would be craft file permissions so the user running fava has only read on .beancount files.. 

something like .beancount files with 644 permissions and fava running with other user besides the .beancount files owner

yagebu:
> What about a read-only mode, an option for disabling import, editor, options and help?

Even without making any changes to the Beancount file, Fava is not "static": filters and queries for instance

rougeth:
> One option would be craft file permissions so the user running fava has only read on .beancount files..
> 
> something like .beancount files with 644 permissions and fava running with other user besides the .beancount files owner

Good idea, I'm gonna try it later. Thanks Shiga :)

> Even without making any changes to the Beancount file, Fava is not "static": filters and queries for instance

Read-only, even if not static, would be enough for my use case. With a static option, it'd be just cheaper to host :) (but yeah, losing filters, queries, etc)

garyp:
I wonder if this could be implemented by compiling fava to WASM and running it from the browser? See for example https://simonwillison.net/2022/May/4/datasette-lite/. If something like Datasette can be made to work with WASM with only minor modifications then perhaps fava could as well.

Rudd-O:
Add a link in the .beancount files to your organization's Web page where users can browse a live Fava which has a checkout of your repo that's read only from the perspective of the UNIX user running Fava.

yagebu:
I'll close this issue as I do not see a viable way to enable this without tons of complexity. If someone else wants to explore the viability, feel free to.

PR #1546, opened 2023-02-17
Create a read-only mode

rougeth:
Hi! I'm creating a read-only mode to Fava that prevents any non GET http request to be processed. This mode is activated by using `--read-only` flag in Fava CLI.

My use case for this PR is that my org already has a well defined process for editing the .beancount files using only the CLI, and we want to make the data available to a broad audience through Fava's UI.

I considered making the files read-only on the OS level (as suggested in the #1424), but having the `--read-only` flag and returning HTTP 401 seemed more appropriate. If you're ok with the change, I can work on the docs.

- Tests passing
- Lint passing
- Discussion: https://github.com/beancount/fava/issues/1424

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,7 +22,7 @@ disable = [
 
 [tool.pylint.design]
 # max args for a function / method
-max-args = 8
+max-args = 9
 
 [tool.pylint.similarities]
 ignore-imports = true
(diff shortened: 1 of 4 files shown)


muriloviana (approved):

yagebu (approved):
Thanks, nice minimalistic solution to implement this :)

yagebu:
Please do fix the one mypy error

rougeth:
To accommodate the new argument on `main()`, I had to increase the `max-args` to 9.

Issue #904, opened 2019-04-10
Upload statements to import directory for ingest


mondjef:
Apologies if this exist already but I have not found how to do it if does.  I do not always have physical access to my box that runs the instance of fava for example when I'm at work.  However I would like to download my bank statements and then import them in fava when I have spare time but I can't as I can't get these statements uploaded so that fava can then import.

Can this feature be added or does it already exist?  My thinking is it should allow one to drag a file over the import menu item and the file will be uploaded to the configured import directory.

yagebu:
This could be added, a PR would be welcome :)

As a workaround you could add a dummy account and upload your documents to the account folder for now and then add this folder as one of the import directories.

mondjef:
I would really love to do a PR but my experience in programming/hacking are limited.  I have looked at the mechanics of the drag and drop functionality of uploading documents on accounts as I think it should work sort of similar to that but I cannot quite follow everything on the code side...too many moving parts.  If anyone can point me a bit better maybe I can piece something worthy together but I think I am out of my league.

mondjef:
yes the account upload feature can serve as a work around but it is kind of hacky and it would be better to have a cleaner more intuitive out of the box method I think long term for users.  In the meantime I have a fix though, thanks

yagebu:
This was implemented in #1551

PR #1551, opened 2023-02-22
Add support for uploading files for importing directly from the import screen

cheif:
This PR add support for uploading documents directly into the import folder from the import screen, I've used the standard `<input type="file">` element, so this should be usable on mobile/desktop, and supports multiple files.

Fixes: #904

image

### DIFF
--- a/frontend/src/reports/import/Import.svelte
+++ b/frontend/src/reports/import/Import.svelte
@@ -45,7 +45,14 @@
       : null;
 
   onMount(() => router.addInteruptHandler(preventNavigation));
-  $: files = data;
+  $: {
+    // Since the user can change e.g. account/name for imported files, we only add any new files here.
+    // This avoids overwriting the users changes
+    const existingFileNames = files.map((file) => file.name);
+    files = files.concat(data.filter(
+      (updated) => !existingFileNames.includes(updated.name)
+    ));
+  };
 
   /**
    * Move the given file to the new file name (and remove from the list).
--- a/frontend/src/reports/import/Import.svelte
+++ b/frontend/src/reports/import/Import.svelte
@@ -1,12 +1,18 @@
 <script lang="ts">
   import { onMount } from "svelte";
 
-  import { deleteDocument, get, moveDocument, saveEntries } from "../../api";
+  import {
+    deleteDocument,
+    get,
+    moveDocument,
+    put,
+    saveEntries,
+  } from "../../api";
   import type { Entry } from "../../entries";
   import { isDuplicate } from "../../entries";
   import { urlFor } from "../../helpers";
   import { _ } from "../../i18n";
-  import { notify } from "../../notifications";
+  import { notify, notify_err } from "../../notifications";
   import router from "../../router";
   import { fava_options } from "../../stores";
   import DocumentPreview from "../documents/DocumentPreview.svelte";
@@ -39,9 +45,7 @@
       : null;
 
   onMount(() => router.addInteruptHandler(preventNavigation));
-  onMount(() => {
-    files = data;
-  });
+  $: files = data;
 
   /**
    * Move the given file to the new file name (and remove from the list).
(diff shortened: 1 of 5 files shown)


yagebu (commented):
Thanks for the PR, this looks great :)

However, this svelte component is not yet really reload-safe. The `$: files = date` line will overwrite any changes the user made to the file names or accounts. Could you try to see if that can be avoided?

This PR would also get bonus points for allowing drag-and-drop ;)

cheif:
> Thanks for the PR, this looks great :)
> 
> However, this svelte component is not yet really reload-safe. The `$: files = date` line will overwrite any changes the user made to the file names or accounts. Could you try to see if that can be avoided?

Great feedback, I tried to make sure the we don't overwrite `files`, rather just append new entries to it. I'm not sure if it's a reasonable approach though, this is my first time dealing with Svelte. 😅 
> 
> This PR would also get bonus points for allowing drag-and-drop ;)

AFAIK the `<input type="file">` component already supports drag and drop (although I guess that would depend on the browser as well), so I think that's covered already 😄

frontend/src/reports/import/Import.svelte
```diff
@@ -39,9 +45,14 @@
...
 
   onMount(() => router.addInteruptHandler(preventNavigation));
-  onMount(() => {
-    files = data;
-  });
+  $: {
+    // Since the user can change e.g. account/name for imported files, we only add any new files here.
+    // This avoids overwriting the users changes
+    const existingFileNames = files.map((file) => file.name);
+    files = files.concat(
```
yagebu, line 52: I think we want this to be `files = data.map(...);` to ensure that we have a consistent order of files and also correctly handle removal of files. For each file in `data`, we should check whether it already exists in `files` and if so, take that object instead. [Or, to also correctly handle changing importers, not take the whole existing object from `files` but go over all importers and check whether for that given file.name,importer_name combination we already have an `account` and/or `newName` and copy those over]

yagebu (commented):
> Great feedback, I tried to make sure the we don't overwrite files, rather just append new entries to it. I'm not sure if it's a reasonable approach though, this is my first time dealing with Svelte.

Yes, definitely the right approach, but see my comment for a minor adjustment that I think would make this more robust. Great first-time Svelte work!

> AFAIK the <input type="file"> component already supports drag and drop (although I guess that would depend on the browser as well), so I think that's covered already

That's true - I was thinking of a more elaborate drag-and-drop where the file would be uploaded directly (and the whole filelist `<div>` could be a drop target). However that might not really be worth the effort (in particular since that means adding a couple of different event handlers for all the drag* events)

frontend/src/reports/import/Import.svelte
cheif: That's a great suggestion!

I looked at this and took some sort of middle-ground where I just used the `importers` array from the existing entry (if there is one). I think this should cover most cases, I guess in theory the list of importers might have changed, but I guess that won't happen i practice? I think it made the code a bit easier to follow at least.

To get this working I also had to switch to use `router.reload()` when moving and deleting files, I've verified that this works now, and I guess the only reason it wasn't used before was due to this reload-logic not being in place?

cheif:
@yagebu Any more comments on this, or do you think we can get it merged? 😀

src/fava/core/ingest.py
```diff
@@ -184,3 +188,28 @@ def extract(self, filename: str, importer_name: str) -> Entries:
             )
 
         return new_entries_list[0][1]
+
+
+def filepath_in_primary_imports_folder(
```
yagebu, line 193: This function needs to replace any separators in the given filname like `filepath_in_document_folder` does. Will fix that in the merge.

yagebu (approved):

sumanchapai:
Why was this PR closed? I have been looking for a feature like this in beancount where I can upload invoices but not sure if this feature is available. I would also be open to contributing if it doesn't exist yet.

PR #1552, opened 2023-02-22
Make sure that the `beancount-files` dropdown is on top of other elements

cheif:
On mobile this is currently obstructed by the time/account/filter fields, this just adds a z-index so that the dropdown is always on top.

NB: After this PR the dropdown is still obstructed by the non-expanded menu, but you can at least see/interact with the elements.

## Current UI
image

## With this fix
image

### DIFF
--- a/frontend/src/sidebar/Header.svelte
+++ b/frontend/src/sidebar/Header.svelte
@@ -81,6 +81,7 @@
     border-bottom-right-radius: 3px;
     border-bottom-left-radius: 3px;
     box-shadow: 0 3px 6px var(--transparent-black);
+    z-index: 1;
   }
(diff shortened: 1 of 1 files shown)


frontend/src/sidebar/Header.svelte
```diff
@@ -73,6 +73,7 @@
     position: absolute;
     top: var(--header-height);
     left: 19px;
+    z-index: 1;
```
yagebu, line 76: Please use one of the CSS variables for this z-index (they're all defined in style.css)

yagebu (changes requested):
Thanks for the PR :)

PR #1561, opened 2023-03-03
Show `errors` in `Editor`

cheif:
This PR add support for showing errors inline in the editor, it uses the [builtin support](https://codemirror.net/examples/lint/) in CodeMirror together with the errors we get from `beancount`.

## UI
image


yagebu:
Nice, thank you! (I'll probably refactor this to always use a global `errors` store instead of just having the `errorCount` as we do now).

PR #1569, opened 2023-03-24
DOC: development.rst: update pip install url from bitbucket to github

westurner:


yagebu:
Thanks for pointing this out - however your fix is also not correct, since it would install fava instead of Beancount ;) I've updated the document in the commit that closed this PR :)

westurner:
Ah, good cal.
Thanks

PR #1571, opened 2023-03-24
Add Entry form now presents last-saved date as default for new entries

jtpereyda:
When adding a sequence of transactions, it is likely enough that some subsequences will have the same date. Therefore, the last-entered date is a reasonable default when adding multiple transaction.

With this change, when one adds a second transaction, the default date will be that of the previously entered transaction. If it's the first transaction, the default date will be today.

Functional-tested in my browser. Not aware of appropriate automated testing for this UI tweak.

### DIFF
--- a/frontend/src/entry-forms/Transaction.svelte
+++ b/frontend/src/entry-forms/Transaction.svelte
@@ -11,12 +11,14 @@
   import { _ } from "../i18n";
   import { notify_err } from "../notifications";
   import { payees } from "../stores";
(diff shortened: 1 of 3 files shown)


yagebu (approved):
I like it, thank you for this change :)

 On merge, I will refactor this a bit - the way that you implemented it, storing the value to local storage is not really needed since we only use the value just after submitting an entry.

jtpereyda:
Thanks @yagebu ! I think it also wouldn't have worked with other entry types prior to your refactors. Cheers!

PR #1600, opened 2023-04-30
Add ability for extensions to have a javascript module

### LINKED ISSUE #1175: Fava extensions with <script> tags
I wanted to write a fava extension with some dynamic logic which I put inside a `<script>` tag in my template `.html` file.

It works well if I load the the extension directly using the url. However, it I click on the extension link in the sidebar (while somewhere else in Fava), it seems to ignores the `<script>` tag completely.

After some digging around, I think it's because of the way Fava intercepts all clicks on links and will in most cases asynchronously load the content of the page and replace the `<article>` contents with them.

More specifically, I think, it is because of [this code in `loadURL`](
https://github.com/beancount/fava/blob/e57fc202ddbc4d1ccc4b90b1990b78fab985b4a9/frontend/src/router.ts#L137):
```js
      const article = document.querySelector("article");
      if (article) {
        article.innerHTML = content;
      }
```

Googling around a bit, it turns out that script elements inserted using innerHTML do not execute when they are inserted. (see https://www.w3.org/TR/2008/WD-html5-20080610/dom.html#innerhtml0).

Although I'm not very familiar with HTML/JS, but I think one solution to this is to use `document.write()` instead of setting `innerHTML`. (See https://developer.mozilla.org/en-US/docs/Web/API/Document/write)

Currently, I'm using a workaround (that I'm not a super fan of):
```
<img src="1" style="display:none" onerror='eval(document.getElementById("myscript").innerHTML);'>
```
which works, by forcing a an `eval`, which may be even worse than using innerHTML.

I hope this is something that can be fixed fairly easily, because it'll really increase the functionality of Fava's extension framework.

tschicke:
This is my interpretation of what was laid out in the comments of #1175 to add the ability for extensions to provide js code to run. It currently supports hooks for initializing the extension and for page loads. On extension report pages, the current js module is also exposed as a property on the window, allowing e.g. buttons like
`<button onclick="extension.customClickHandler()">Click me</button>`
Let me know if there is anything that needs to be changed or moved (potentially the extension code in main.ts).

Fixes #1175


src/fava/application.py
```diff
@@ -375,6 +375,18 @@ def extension_report(report_name: str) -> str:
         return abort(404)
 
 
+@app.route("/<bfile>/extension_module/<extension_name>/")
```
yagebu, line 378:
```suggestion
@app.route("/<bfile>/extension_js_module/<extension_name>.js")
```

frontend/src/main.ts
```diff
@@ -124,7 +225,7 @@ function init(): void {
     errors.set(val.errors);
   });
 
-  router.trigger("page-loaded");
+  initExtensions();
```
yagebu, line 228: Loading and running the extension JS code shouldn't block the `page-loaded` event IMHO but should just run asynchronously.

src/fava/ext/portfolio_list/PortfolioList.js
```diff
@@ -0,0 +1,7 @@
+export default {
+  onReloadButtonClick() {
+    location.reload();
+  },
+  onExtensionInit() {},
```
yagebu, line 5: This extension should showcase some simple usecases of all the functions that we might call from the frontend - otherwise it is not really possible to test this functionality

yagebu (commented):
Thanks for this PR :) This is the general direction I've envisioned for this.

I've left a couple of comments in-place and here are some general remarks from a first quick glance over the code:

- Maybe naming the endpoints and such "extension_js_module(s)" is a bit more explicit than "extension_module(s)".
- No polluting of the global namespace please

.pre-commit-config.yaml
```diff
@@ -36,4 +36,4 @@ repos:
             language: node
             entry: ./frontend/node_modules/eslint/bin/eslint.js --max-warnings 0
             require_serial: true
-            files: \.(js|ts|svelte)$
+            files: frontend/.*\.(js|ts|svelte)$
```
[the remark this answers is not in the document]
tschicke: I was having trouble getting the linter to run properly on the extension code. Initially I was seeing:
`ESLint looked for configuration files in fava/src/fava/ext/portfolio_list and its ancestors. If it found none, it then looked in your home directory.`
I tried moving the .eslintrc.js file to the root of the project from frontend, but then I got:
`ESLint couldn't find the config "airbnb-base" to extend from. Please check that the name of the config is correct.
The config "airbnb-base" was referenced from the config file in "/home/tschicke/dev/projects/fava/.eslintrc.js".`
Moving node_modules as well doesn't fix the issue, and causes a few other issues related to tsconfig.json, and I didn't think making another small node project with another node_modules to handle the extension code was worth it for some example code.
I agree it would be nice to lint the extension code, so if you have a simpler way to get that working I can try to do that.

src/fava/core/extensions.py
```diff
@@ -59,6 +59,27 @@ def exts_for_hook(self, hook: str) -> list[FavaExtensionBase]:
...
 
+    def extension_modules(self) -> list[str]:
+        """Find all extensions that have a javascript module that should be loaded."""
+        return [
+            ext.name
+            for ext_class, ext in self._instances.items()
+            if (
+                Path(getfile(ext_class)).parent
+                / f"{ext_class.__qualname__}.js"
+            ).exists()
```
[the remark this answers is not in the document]
tschicke: Since there's no display name like `report_title`, should it just be a boolean like `has_js_module`?

frontend/src/main.ts
tschicke: I originally did this because I'm using the `page-loaded` event to trigger the `onExtensionPageLoad` extension hook, so if `page-loaded` is triggered before the extensions are loaded, something else would need to call that hook, only once the modules are loaded and if you happen to be navigating directly to the extension page. If that seems ok I can change that around.

src/fava/ext/portfolio_list/PortfolioList.js
tschicke: Yeah sorry, I have an extension that is using this functionality locally, but I couldn't come up with something general that would be useful as a demo. I'll try to add some more concrete usecases, but ideas for demo functionality would be useful.

frontend/src/main.ts
```diff
@@ -60,10 +61,74 @@ function defineCustomElements() {
   customElements.define("svelte-component", SvelteCustomElement);
 }
 
+interface ExtensionModule {
+  onExtensionInit?: (this: ExtensionModule) => void;
+  onExtensionPageLoad?: (this: ExtensionModule) => void;
+}
+
+declare global {
+  interface Window {
```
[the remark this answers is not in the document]
tschicke: I did this so that js modules could be used to provide only callbacks for the HTML, without needing to use any of the "lifetime" hooks just to make those callbacks accessible to the HTML. So if you want to put a button in the HTML that calls some function defined in the module, without setting something in the global namespace, you'd need to define the `onExtensionPageLoad` hook, assign your custom function to some name in the global namespace, and use that name in the HTML, which seemed like boilerplate that it would be nice to avoid if possible.

frontend/src/main.ts
```diff
@@ -60,10 +61,74 @@ function defineCustomElements() {
...
+
+  initExtension(): void {
+    this.tryCall("onExtensionInit", []);
+  }
+
+  onPageLoad(): void {
+    this.tryCall("onExtensionPageLoad", []);
+  }
+
+  private tryCall<
```
tschicke, line 96: I'm also planning to get rid of this tryCall abstraction. The TypeScript trickery is more complicated than it's worth I think, and I don't think that it actually works properly if the functions have any arguments.

src/fava/json_api.py
```diff
@@ -179,6 +179,12 @@ def get_changed() -> bool:
 api_endpoint(get_ledger_data)
 
 
+@api_endpoint
+def get_extension_modules() -> list[str]:
```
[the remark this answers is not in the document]
tschicke: Thanks, I didn't realize how LedgerData worked. Done

src/fava/core/extensions.py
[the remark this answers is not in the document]
tschicke: I changed this to a `has_js_module` bool

src/fava/application.py
tschicke: Done

src/fava/ext/portfolio_list/PortfolioList.js
tschicke: I added a textbox and some buttons that allow you to specify an account name regex to override the config for PortfolioList and generate an account listing for a string entered in the frontend.

.pre-commit-config.yaml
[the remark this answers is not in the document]
yagebu: Thanks for the attempt, this doesn't seem like it's worth the effort.

frontend/src/main.ts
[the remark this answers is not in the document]
yagebu: I've changed the onExtensionPageLoad example to assign the event handler directly. Since I want to avoid adding stuff to the global namespace and HTML inline event handlers

frontend/src/main.ts
yagebu: This doesn't quite work out as it is currently implemented - changes to the ledgerData trigger a re-run of initExtensions which then spuriously triggers "page-loaded". I'll change the logic in my commit

yagebu:
Thanks again for the work on this :) I've squashed your commits and made some adjustments in a second commit.

PR #1603, opened 2023-05-02
remove ctrl-c as shortcut of show charts

upsuper:
Ctrl+C is the shortcut to copy content on Windows and Linux, but on Fava it also toggles the charts. I have to say this is very annoying. I believe it is pretty common that someone wants to copy some data from the page, while it is rare that someone wants to hide the chart. I'm not even sure it's even useful to allow hiding it on UI at all.

So I don't think we need a shortcut for this functionality. It might be fine to pick another shortcut, but I'm not sure it's even worth it.

### DIFF
--- a/frontend/src/charts/Chart.svelte
+++ b/frontend/src/charts/Chart.svelte
@@ -4,7 +4,6 @@
   import type { Writable } from "svelte/store";
 
   import { _ } from "../i18n";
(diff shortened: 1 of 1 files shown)


upsuper:
The lint-python test seems to be broken.

yagebu:
Thanks!

PR #1604, opened 2023-05-02
lint: fix lint-python failure on charts.py

upsuper:
This fixes the failure on lint-python test.


yagebu:
Thanks, this seems to be due to the types of Flask changing somewhat between versions and the pinning of versions didn't work to well, #1605 should fix this.

Issue #1530, opened 2023-01-05
How to do rest api in extension?


tarioch:
Is it possible to add custom rest api endpoints? I would like to create an extension which would need some dynamic parts.

yagebu:
See #1613

yagebu:
Implemented in #1613

PR #1613, opened 2023-05-23
Add custom endpoints to extensions

tschicke:
Thanks for merging #1600. This is a small WIP addon to that, that adds the ability for extensions to register their own endpoints that can be called at '/beancount/extension/<ext_name>/<endpoint>', so that the template html or the new js module can communicate with the extension backend, which I think opens up the possibility for much more advanced extensions. This is a pretty bare-bones version of the change that doesn't handle arguments particularly well, and might have issues with locking (I'm not sure how locking around requests works), but it has been working well for me in a few personal extensions that I've written in the past few weeks.

I'm not sure if this type of thing is something you'd want to support, so I wanted to see if this was interesting at all before spending too much time cleaning it up.

### DIFF
--- a/frontend/src/extensions.ts
+++ b/frontend/src/extensions.ts
@@ -6,26 +6,134 @@
 import { get as store_get } from "svelte/store";
 
 import { getUrlPath, urlFor } from "./helpers";
+import { fetch } from "./lib/fetch";
 import { log_error } from "./log";
 import { extensions } from "./stores";
+
+class ExtensionApi {
+  extension_name: string;
+
+  constructor(extension_name: string) {
+    this.extension_name = extension_name;
+  }
+
+  request(
+    endpoint: string,
+    method: string,
+    params: Record<string, string | number> | undefined,
+    body: object | undefined,
+    output: "json" | "string" | "raw" = "json",
+  ) {
+    const url = urlFor(
+      `extension/${this.extension_name}/${endpoint}`,
+      params,
+      false,
+    );
+    let opts = {};
+    if (body) {
+      opts =
+        body instanceof FormData
+          ? { body }
+          : {
+              headers: { "Content-Type": "application/json" },
+              body: JSON.stringify(body),
+            };
+    }
+    const request = fetch(url, { method, ...opts });
+    if (output === "json") {
+      return request.then((res) => res.json());
+    }
+    if (output === "string") {
+      return request.then((res) => res.text());
+    }
+    return request;
+  }
+
+  get(
+    endpoint: string,
+    params: Record<string, string | number>,
+    output: "json" | "string" | "raw" = "json",
+  ) {
+    return this.request(endpoint, "GET", params, undefined, output);
+  }
+
+  put(
+    endpoint: string,
+    body: object | undefined,
+    output: "json" | "string" | "raw" = "json",
+  ) {
+    return this.request(endpoint, "PUT", undefined, body, output);
+  }
+
+  post(
+    endpoint: string,
+    body: object | undefined,
+    output: "json" | "string" | "raw" = "json",
+  ) {
+    return this.request(endpoint, "POST", undefined, body, output);
+  }
+
+  delete(
+    endpoint: string,
+    body: object | undefined,
+    output: "json" | "string" | "raw" = "json",
+  ) {
+    return this.request(endpoint, "DELETE", undefined, body, output);
+  }
+}
+
+export interface ExtensionContext {
+  api: ExtensionApi;
+  getExt: (name: string) => Promise<ExtensionData>;
+}
 
 /**
  * The Javascript code of a Fava extension should export an object of this type.
  */
 export interface ExtensionModule {
   /** Initialise this Javascript module / run some code on the initial load. */
-  init?: () => void;
+  init?: (c: ExtensionContext) => void;
   /** Run some code after any Fava page has loaded. */
-  onPageLoad?: () => void;
+  onPageLoad?: (c: ExtensionContext) => void;
   /** Run some code after the page for this extension has loaded. */
-  onExtensionPageLoad?: () => void;
+  onExtensionPageLoad?: (c: ExtensionContext) => void;
 }
 
-async function loadExtensionModule(name: string): Promise<ExtensionModule> {
+export class ExtensionData implements Required<ExtensionModule> {
+  context: ExtensionContext;
+
+  extension: ExtensionModule;
+
+  constructor(context: ExtensionContext, extension: ExtensionModule) {
+    this.context = context;
+    this.extension = extension;
+  }
+
+  init(): void {
+    this.extension.init?.(this.context);
+  }
+
+  onPageLoad(): void {
+    this.extension.onPageLoad?.(this.context);
+  }
+
+  onExtensionPageLoad(): void {
+    this.extension.onExtensionPageLoad?.(this.context);
+  }
+}
+
+async function loadExtensionModule(
+  name: string,
+  getExtension: (name: string) => Promise<ExtensionData>,
+): Promise<ExtensionData> {
   const url = urlFor(`extension_js_module/${name}.js`, undefined, false);
   const mod = await (import(url) as Promise<{ default?: ExtensionModule }>);
   if (typeof mod.default === "object") {
-    return mod.default;
+    const context: ExtensionContext = {
+      api: new ExtensionApi(name),
+      getExt: getExtension,
+    };
+    return new ExtensionData(context, mod.default);
   }
   throw new Error(
     `Error importing module for extension ${name}: module must export "default" object`,
@@ -33,17 +141,17 @@ async function loadExtensionModule(name: string): Promise<ExtensionModule> {
 }
 
 /** A map of all extensions modules that have been (requested to be) loaded already. */
-const loaded_extensions = new Map<string, Promise<ExtensionModule>>();
+const loaded_extensions = new Map<string, Promise<ExtensionData>>();
 
 /** Get the extensions module - if it has not been imported yet, initialise it. */
-async function getExt(name: string): Promise<ExtensionModule> {
+async function getExt(name: string): Promise<ExtensionData> {
   const loaded_ext = loaded_extensions.get(name);
   if (loaded_ext) {
     return loaded_ext;
   }
-  const ext = loadExtensionModule(name);
+  const ext = loadExtensionModule(name, getExt);
   loaded_extensions.set(name, ext);
-  (await ext).init?.();
+  (await ext).init();
   return ext;
 }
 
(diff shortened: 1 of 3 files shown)


yagebu:
> I'm not sure if this type of thing is something you'd want to support

Sure, let's add this if you have a use case for it.

yagebu:
This fixes #1530

src/fava/application.py
```diff
@@ -308,6 +309,18 @@ def report(report_name: str) -> str:
             return render_template("_layout.html", active_page=report_name)
         return abort(404)
 
+    @fava_app.route("/<bfile>/extension/<ext_name>/<endpoint>")
+    def extension_endpoint(ext_name: str, endpoint: str):
+        response = g.ledger.extensions.custom_endpoint(
+            ext_name, endpoint, request.args
```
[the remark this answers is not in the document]
tschicke: I just realized that this isn't actually the most recent version of this change, and the newer one has slightly different handling of args, but yeah, importing the request in the extension makes way more sense

src/fava/application.py
```diff
@@ -308,6 +309,18 @@ def report(report_name: str) -> str:
             return render_template("_layout.html", active_page=report_name)
         return abort(404)
 
+    @fava_app.route("/<bfile>/extension/<ext_name>/<endpoint>")
+    def extension_endpoint(ext_name: str, endpoint: str):
+        response = g.ledger.extensions.custom_endpoint(
```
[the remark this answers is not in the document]
tschicke: Done

src/fava/core/extensions.py
```diff
@@ -96,6 +96,13 @@ def template_and_extension(
 
         raise LookupError("Extension report not found.")
 
+    def custom_endpoint(self, ext_name: str, endpoint: str, args):
+        for cls in self._instances:
+            ext = self._instances[cls]
+            if cls.__name__ == ext_name and not ext.endpoints == None:
+                if endpoint in ext.endpoints:
+                    return ext.endpoints[endpoint](args)
```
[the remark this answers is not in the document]
tschicke: I'm still using a dictionary, but it is managed by FavaExtensionBase now, not the extension itself. Let me know if you want that changed

tschicke:
> This fixes #1530

Oh thanks, I didn't notice there was already an issue for this.

tschicke:
I made the `extension_endpoint` decorator the main way for extensions to register endpoints. This can be used either without arguments (endpoint name = function name, methods = ["GET"]):
`@extension_endpoint
    def update_status(self):...`

Or with arguments (endpoint name, methods):
`@extension_endpoint("update", methods=["POST"])
    def update_endpoint(self):...`

I also created an ExtensionApi class in the frontend. This gets set on each extension module, and had get, put, post, and delete functions to call endpoints within the extension.

The last thing might be overkill, but it looks through all elements with class name "extension-handler" and goes through each "data-handler-<eventName>" attribute, converting the value to a handler for that event that has access to the extension module. This normally wouldn't be necessary, and you could just set an id on the required elements and init the event handlers in onExtensionPageLoad, but this makes it easier when the html template creates dynamic elements, like in a loop, where the handler for each element needs the loop variable, e.g. from one of my extensions:
`{% for key, value in extension.status()|dictsort %}
    <tr>
      <td>{{ key }}</td>
      <td><pre>{{ value }}</pre></td>
      <td><button class="extension-handler" data-handler-click="this.api.post('disable', {name:'{{ key }}'})">Disable</button></td>
      <td><button class="extension-handler" data-handler-click="this.api.post('enable', {name:'{{ key }}'})">Enable</button></td>
    </tr>
{% endfor %}
`
Without something like the "data-handler" thing, it's more difficult to do something like `{name:'{{ key }}'}` and you need to do something like store the key somewhere and refer back to it later. I think it cleans some stuff up, but I understand if it's a little too hacky (like the need for `// eslint-disable-next-line @typescript-eslint/no-implied-eval`)

I still need up update the documentation and add an example of the functionality to an included extension, but I wanted to get your opinion on this direction before doing that.

frontend/src/extensions.ts
```diff
@@ -19,6 +67,8 @@ export interface ExtensionModule {
   onPageLoad?: () => void;
   /** Run some code after the page for this extension has loaded. */
   onExtensionPageLoad?: () => void;
+
+  api: ExtensionApi;
```
yagebu, line 71: This is in the wrong place IMHO - this interface is for what the extension can provide. Maybe we could pass a "context" object ({ api: ExtensionApi}) to all of the event handlers instead?

frontend/src/extensions.ts
```diff
@@ -9,6 +9,54 @@ import { getUrlPath, urlFor } from "./helpers";
...
+    if (body) {
+      opts =
+        body instanceof FormData
+          ? { body }
+          : {
+              headers: { "Content-Type": "application/json" },
+              body: JSON.stringify(body),
+            };
+    }
+    return fetch(url, { method, ...opts }).then((res) => res.json());
```
yagebu, line 40: This should import fetch from lib/fetch to ensure that cookies are passed for example.

yagebu (commented):
> The last thing might be overkill, but it looks through all elements with class name "extension-handler" and goes through each "data-handler-" attribute, converting the value to a handler for that event that has access to the extension module. 

I'm not a fan of that - I generally dislike inline event handlers. I also think the frontend code should be mostly agnostic of how the extension does frontend logic (unless it's reusing some existing Fava logic so I could imagine having some more direct access to the Svelte components or custom stuff like the router) - it could do so using a full-fledged frontend-rendering framework like React or Svelte or for ergonomics closer to HTML with inline handlers something like htmx or Stimulus.

frontend/src/extensions.ts
tschicke: Yeah I didn't notice the comment above the interface. I think passing a context works, and it's flexible to be expanded in the future

frontend/src/extensions.ts
tschicke: Done

tschicke:
> I'm not a fan of that - I generally dislike inline event handlers. I also think the frontend code should be mostly agnostic of how the extension does frontend logic (unless it's reusing some existing Fava logic so I could imagine having some more direct access to the Svelte components or custom stuff like the router) - it could do so using a full-fledged frontend-rendering framework like React or Svelte or for ergonomics closer to HTML with inline handlers something like htmx or Stimulus.

No problem, I figured it was probably too much of a hack. I can achieve most of that functionality in an extension that uses onPageLoad, since like I mentioned some sort of ability to specify things inline makes working with loops in extension html much easier.
I think making something like React or Svelte accessible to the extension frontend would be the ideal way to solve this, but Svelte would require shipping the Svelte compiler with fava, and invoking it at runtime, which seemed like too much, and adding a runtime framework like React or one of the others seemed confusing since the core frontend already uses Svelte.

andreasgerstmayr:
> I think making something like React or Svelte accessible to the extension frontend would be the ideal way to solve this, but Svelte would require shipping the Svelte compiler with fava, and invoking it at runtime, which seemed like too much, and adding a runtime framework like React or one of the others seemed confusing since the core frontend already uses Svelte.

Tangentially related: I once did a PoC of loading React in an extension: https://github.com/andreasgerstmayr/fava-dashboards/blob/react/frontend/src/app.tsx. Registering custom routes would be awesome, as currently it's writing JSON in a `<script>` tag and loading that to bootstrap the React app.

tschicke:
> I can achieve most of that functionality in an extension that uses onPageLoad
After playing around with this, the one extra thing I need in order to do this in an extension rather than fava is an ability to get other extensions from within an extension. I added the getExt function to the context so that any extension can call that to query for an instance of another extension if they need to.

frontend/src/extensions.ts
```diff
@@ -6,44 +6,152 @@
...
   /** Initialise this Javascript module / run some code on the initial load. */
-  init?: () => void;
+  init?: (c: ExtensionContext) => void;
   /** Run some code after any Fava page has loaded. */
-  onPageLoad?: () => void;
+  onPageLoad?: (c: ExtensionContext) => void;
   /** Run some code after the page for this extension has loaded. */
-  onExtensionPageLoad?: () => void;
+  onExtensionPageLoad?: (c: ExtensionContext) => void;
+}
```
tschicke, line 100: I also ended up adding this:
```
init(c) {
    this.api = c.api;
},
```
to a bunch of the extensions I've been testing/prototyping with, since the inline event handler extension that I moved that inline loading code to sets up event handlers for other extensions that need a context, but aren't being called from within an "extension entrypoint", so it doesn't have access to those contexts. So I'm just storing it manually on the extension module so I can reference it from anywhere. It was also a little easier than having to pass around a context within the extension module.
What if we used the ExtensionModule type to specify the interface that extension JS code should export, but specify that Fava's extension initialization code will "construct" the module into a subtype of ExtensionModule that includes things like a context variable (or just api variable directly). That subtype would roughly be like what ExtensionData is in this change, but if we went this route I think that could be made into an interface instead of a class.

yagebu:
> I think making something like React or Svelte accessible to the extension frontend would be the ideal way to solve this, but Svelte would require shipping the Svelte compiler with fava, and invoking it at runtime, which seemed like too much, and adding a runtime framework like React or one of the others seemed confusing since the core frontend already uses Svelte.

There's no need for invoking the Svelte compiler at runtime - if you want to build Svelte components, your extension should have a build process and if you want to use Fava's components, we can expose those to extensions. It's really up to the extension to pick a frontend framework and use it.

frontend/src/extensions.ts
yagebu: I've refactored the ExtensionData class to be an internal implementation detail and not be exposed to the extension. I do not think that our code should be doing any mutation on the ExtensionModule or constructing some subtype. If this continues to be a pain point, I'd suggest a different solution: allow the init extension handler to also register the onPageLoad and onExtensionPageLoad handlers. That way, you can construct your extension class/function/module/whatever in the init function and have the context easily available.

tschicke:
> > I think making something like React or Svelte accessible to the extension frontend would be the ideal way to solve this, but Svelte would require shipping the Svelte compiler with fava, and invoking it at runtime, which seemed like too much, and adding a runtime framework like React or one of the others seemed confusing since the core frontend already uses Svelte.
> 
> There's no need for invoking the Svelte compiler at runtime - if you want to build Svelte components, your extension should have a build process and if you want to use Fava's components, we can expose those to extensions. It's really up to the extension to pick a frontend framework and use it.

Thanks, yeah I realized that was the intended way to do that after seeing how https://github.com/redstreet/fava_investor works. I have 5+ local extensions that I'm using/testing with so adding svelte/react to a bunch of different extensions would add a bunch of boilerplate, so I was trying to come up with a way that fava could do some of that work for me. But I agree that the flexibility of letting extensions use whatever frontend they want is better.

frontend/src/extensions.ts
tschicke: Ok no problem, it's not much of a pain point anyways, and I think setting `this.api = c.api;` in init() is a good enough workaround/hack for me for when I don't feel like using the context properly.

I also noticed you removed the 'output' parameter from the get, post, etc. functions. Is that because you want to treat that as a lower level parameter so if you want non-json you should just use request() directly?

PR #1616, opened 2023-05-24
try workaround for Safari not properly handling Enter on the filter form

yagebu:
Ref #809 and #1528.

### DIFF
--- a/frontend/src/AutocompleteInput.svelte
+++ b/frontend/src/AutocompleteInput.svelte
@@ -4,7 +4,11 @@
   import { keyboardShortcut } from "./keyboard-shortcuts";
   import { fuzzyfilter, fuzzywrap } from "./lib/fuzzy";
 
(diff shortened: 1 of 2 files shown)


seltzered:
@yagebu  I don't actively use macOS anymore, but if I find a moment to test will reply to this.

yagebu:
I've merged this change now in any case, there's little risk that it breaks anything.

> I don't actively use macOS anymore, but if I find a moment to test will reply to this.

This would still be very welcome :)

seltzered:
Took me a while to setup, but did a brief test of [6293d09](https://github.com/beancount/fava/commit/6293d09c216740a90f7d24d18ad4c6764c5812b8) on Safari 16.5 and #809 looks resolved. 

(I don't have an understanding for why, perhaps @vanillacode314 can chime in.)

PR #1619, opened 2023-05-27
Add calculation and disply for average of incomes and expenses

vahidmah:
This feature is calculating an average for incomes and expenses. I originally developed it for my own use but then thought of sharing it. Please check this PR and let me know what you think about it. 

This is a screenshot using **long-example**

image


yagebu:
Hi, thank you for submitting the PR :) I think these averages are a useful bit of information but I find them to be a bit out of place. Since the account trees are for the whole time range and the charts show information per interval, maybe the average could be added to the chart instead (as part of the tooltip?)?

vahidmah:
Hi,
I wasn't quite certain about where to place this information. One possible location could be the tooltip of the chart. The challenge arises from the fact that the average is calculated for the entire duration (for the year 2014 in the given example), while hovering over a specific bar only displays data relevant to that particular interval (Jan 2014, below). 

image

So, the question is, does this average value truly belong in the tooltip?

Maybe, another option could be incorporating a small summary section alongside the chart, where essential statistics like the average are showcased explicitly.

PR #1630, opened 2023-07-01
Fix offset of AppMenuItem

tschicke:
Commit c9a19191 changed the vertical padding of ".dropdown > span" from 0.75rem to 0.7rem, and added a top margin of 0.75rem to the "ul" that is in ".dropdown > span". At least on Firefox on Linux this causes the dropdown to be 1 pixel too far down, causing a gap between the hit detection for the hover target and the actual dropdown, which causes the dropdown to disappear when trying to move the mouse over it.

### DIFF
--- a/frontend/src/reports/editor/AppMenuItem.svelte
+++ b/frontend/src/reports/editor/AppMenuItem.svelte
@@ -48,7 +48,7 @@
     display: none;
     width: 500px;
     max-height: 400px;
-    margin: 0.75rem 0 0 -0.5rem;
+    margin: 0.7rem 0 0 -0.5rem;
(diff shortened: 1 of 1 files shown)


yagebu:
Thanks for the PR. I'm also using Firefox on Linux but it worked for me so far :D (maybe due to some default scaling that I've turned on).

For some reason, even with your patch, there's still a gap on smaller screen widths (maybe due to the base font size changing?). I'll try to come up with a fix.

yagebu:
See #1633 - does that also fix it for you?

tschicke:
Thanks for looking at this, yeah #1633 also fixes it.

PR #1631, opened 2023-07-01
Delete entries

tschicke:
This adds a doDelete function to the TypeScript API for calling DELETE endpoints, and adds a DELETE handler for the entry_slice endpoint to delete a given source slice. It also adds a button next to the save button in the SliceEditor to delete the given slice.

This allows an entry to actually be deleted without going into the editor. Currently, I believe the best you can do is save an empty slice, but that leaves an extra newline in the source file.

### DIFF
--- a/frontend/src/editor/DeleteButton.svelte
+++ b/frontend/src/editor/DeleteButton.svelte
@@ -1,0 +1,17 @@
+<script lang="ts">
+  import { _ } from "../i18n";
+
+  export let deleting: boolean;
+  export let onDelete: () => void;
+
+  $: buttonContent = deleting ? _("Deleting...") : _("Delete");
+</script>
+
+<button
+  type="button"
+  class="muted"
+  on:click={onDelete}
+  title={`${_("Delete")}`}
+>
+  {buttonContent}
+</button>
--- a/src/fava/json_api.py
+++ b/src/fava/json_api.py
@@ -289,6 +289,13 @@ def put_source_slice(entry_hash: str, source: str, sha256sum: str) -> str:
 
 
 @api_endpoint
+def delete_source_slice(entry_hash: str, sha256sum: str) -> str:
+    """Delete an entry source slice."""
+    g.ledger.file.delete_entry_slice(entry_hash, sha256sum)
+    return f"Deleted entry {entry_hash}."
+
+
+@api_endpoint
 def put_format_source(source: str) -> str:
     """Format beancount file."""
     return align(source, g.ledger.fava_options.currency_column)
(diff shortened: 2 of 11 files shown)


frontend/src/editor/DeleteButton.svelte
```diff
@@ -0,0 +1,12 @@
+<script lang="ts">
+  import { _ } from "../i18n";
+  import { modKey } from "../keyboard-shortcuts";
+
+  export let deleting: boolean;
+
+  $: buttonContent = deleting ? _("Deleting...") : _("Delete");
+</script>
+
+<button type="submit" value="Delete" title={`${_("Delete")} (${modKey}+s)`}>
```
yagebu, line 10: I'd just have this as a `type="button"` and then have a separate event handler for this one.

src/fava/json_api.py
```diff
@@ -288,6 +288,13 @@ def put_source_slice(entry_hash: str, source: str, sha256sum: str) -> str:
     return g.ledger.file.save_entry_slice(entry_hash, source, sha256sum)
 
 
+@api_endpoint
+def delete_source_slice(entry_hash: str, sha256sum: str) -> str:
+    """Delete an entry source slice."""
+    g.ledger.file.delete_entry_slice(entry_hash, sha256sum)
+    return f"Delete entry {entry_hash}."
```
yagebu, line 295:
```suggestion
    return f"Deleted entry {entry_hash}."
```

src/fava/json_api.py
```diff
@@ -288,6 +288,13 @@ def put_source_slice(entry_hash: str, source: str, sha256sum: str) -> str:
     return g.ledger.file.save_entry_slice(entry_hash, source, sha256sum)
 
 
+@api_endpoint
```
yagebu, line 291: Could you also add a test for this API endpoint?

yagebu (changes requested):
Thanks - this looks good, just some minor comments :)

frontend/src/editor/DeleteButton.svelte
tschicke: Yeah I think that simplifies things quite a bit

src/fava/json_api.py
tschicke: I corrected that and another typo that I missed before

src/fava/json_api.py
tschicke: I added a test for the delete endpoint, but it just uses the internal file api to re-add the deleted entry

yagebu (approved):

PR #1632, opened 2023-07-01
Batch editor extension

tschicke:
This exposes the SliceEditor svelte component through svelte-custom-elements, which allows extensions to use the SliceEditor by using 
`<svelte-component type="slice-editor"><script type="application/json">{{ <json props>|tojson }}</script></svelte-component>`

This also adds another built in extension, batch_edit, which uses the slice editor to provide an editor page where you can input (part of) a BQL query to limit to a set of entries, and get a slice editor for up to 20 of those entries one after the other, to allow you to view different sets of transactions/entries simultaneously while editing.

### DIFF
--- a/frontend/src/svelte-custom-elements.ts
+++ b/frontend/src/svelte-custom-elements.ts
@@ -5,9 +5,14 @@
 import type { SvelteComponent } from "svelte";
 
 import ChartSwitcher from "./charts/ChartSwitcher.svelte";
+import SliceEditor from "./editor/SliceEditor.svelte";
 
-const components = new Map<string, typeof SvelteComponent<{ data?: unknown }>>([
+const components = new Map<
+  string,
+  typeof SvelteComponent<Record<string, unknown>>
+>([
   ["charts", ChartSwitcher],
+  ["slice-editor", SliceEditor],
 ]);
 
 /**
@@ -17,7 +22,7 @@ const components = new Map<string, typeof SvelteComponent<{ data?: unknown }>>([
  * of the valid values in the Map above.
  */
 export class SvelteCustomElement extends HTMLElement {
-  component?: SvelteComponent<{ data?: unknown }>;
+  component?: SvelteComponent<Record<string, unknown>>;
 
   connectedCallback(): void {
     if (this.component) {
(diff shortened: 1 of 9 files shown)


yagebu:
I've been wanting to add some sort of batch editing functionality for a while :)  I'd like for this to be a builtin feature. Do you need the slice editor for some other purpose?

tschicke:
Yeah I knew batch editing had been talked about before, and it wasn't too hard to put this together as an extension, so I figured I'd try it out since the built in feature doesn't exist yet.
No, I don't specifically need the slice editor for anything else. `svelte-component` seemed like a nice way to expose some higher level UI functionality that fava implements to extensions, since they can't directly use svelte, but it seemed pretty specific to ChartSwitcher since it requires the component to have only a `data` prop. So I figured I'd start with exposing the slice editor since I had the idea for the batch editor that could use it.
If you want to hold off on the batch editor extension for the built in feature (or I guess even if you want to merge it but replace it later with a built in feature), are there any details/discussion anywhere about what the builtin editor would look like and how it would work? If I get some time and have some idea of what it should look like, I may try to attempt that.

BenjaminDebeerst:
@paulsc any chance you'd be willing to publish the patch/branch with the "apply all" button from your demo video? You extension is nice, but a lot less practical than the version shown here, since one needs to change to another view to reapply a change to the rest of the list.

No worries if this is a branch off of an old version of fava, I'd be happy to port it forward to keep it with the few other personal edits on my fork.

Thanks in advance!

paulsc:
> @paulsc any chance you'd be willing to publish the patch/branch with the "apply all" button from your demo video? You extension is nice, but a lot less practical than the version shown here, since one needs to change to another view to reapply a change to the rest of the list.
> 
> No worries if this is a branch off of an old version of fava, I'd be happy to port it forward to keep it with the few other personal edits on my fork.
> 
> Thanks in advance!

Hi @BenjaminDebeerst, thanks for your interest! Unfortunately I don't seem to have that code anymore. It broke almost immediately due to changes upstream. I've tried the manual patch and fixing it when upstream changes come -strategy before, but it in the long run it was always too annoying and better to stick to the extension system with a fixed API.

BenjaminDebeerst:
Thanks for your response. Should someone be interested, I re-built the "apply to all" checkbox on top of your plugin:

* I gave the plugin a new POST endpoint that applies the last edit to the currently filtered journal, reusing the existing logic: https://github.com/BenjaminDebeerst/fava-edit-replay/commit/73ddb067e82c8068c9ff2061f390e9e0ebe17df8
* I added a small patch to Fava that adds the checkbox and calls that endpoint: https://github.com/BenjaminDebeerst/fava/commit/26cdfb2d1d227a3bc8a2a6cd761fc6fdb9a2387c

The patch on Fava is designed deliberately to have small surface area with core Fava: All the new functionality is in a separate file and the touch-points in the `SliceEditor` are only a single added import, the checkbox and the additional call within the save function.

Forward-porting this patch shouldn't be that big of a deal:  I'm doing that with a small set of personal patches already.

PR #1636, opened 2023-07-04
Make tables wrap better

### LINKED ISSUE #1575: Price and cost columns always partially obscured
No matter how wide the screen is. Tried with Chrome, Edge.

image

Fava has been like this for as long as I remember. Are others seeing this too or am I doing something wrong?

tschicke:
Fixes #1575

Commit ca6d1d928 made it so that currencies don't wrap at all, which means in many cases currencies or amounts get cut off, whenever one or more of the following happen:
 - the currency name is too long
 - the amount is large
 - there are a lot of decimal places in an amount

This removes the white-space rule that was added by that commit, which allows amounts and commodities to wrap if necessary, so that they can be read.
It also changes the overflow of .num from hidden to scroll. This means that if an individual number or currency is too long to fit, but won't wrap because it's one word, at least you can scroll the box to see the rest

Finally it adds a line break between the value and the diff in _tree_table.html and moves the diff after the value. This is a similar issue where the combination of the value and the diff is often long enough to overflow the cell that it's in, so the line break means everything fits in a single cell

### DIFF
--- a/frontend/css/journal-table.css
+++ b/frontend/css/journal-table.css
@@ -60,7 +60,7 @@
 }
 
 .journal .postings .num {
-  overflow: hidden;
+  overflow: scroll;
   line-height: 16px;
 }
 
@@ -126,7 +126,6 @@
 
 .journal p > .num {
   width: 9rem;
-  white-space: pre;
   border-left: 1px solid var(--table-border);
 }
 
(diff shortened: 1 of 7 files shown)


yagebu:
Thanks :)

redstreet:
Very cool, thank you!

PR #1639, opened 2023-07-13
Create github workflow for pushing to ghcr.io

AmrikSD:
This PR should cause any push to `main` to automatically cause the Dockerfile to be built, tagged and pushed to `ghcr.io/beancount/fava:latest` 

whenever there is a commit _tagged_ , the Dockerfile should again be built, tagged, but this time pushed to `ghcr.io/beancount/fava:<tag>`

The idea being that instead of relying on other images that are floating around like yegle or alexiri which get stale an "official" image can be used instead.

### DIFF
--- a/.github/workflows/docker-image.yml
+++ b/.github/workflows/docker-image.yml
@@ -1,0 +1,38 @@
+name: Docker Image
+
+on:
+  push:
+    branches: [ "main" ]
+    tags: ["*"]
+
+jobs:
+
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v3
+
+    - uses: docker/login-action@v1
+      with:
+        registry: ghcr.io
+        username: ${{github.actor}}
+        password: ${{secrets.GITHUB_TOKEN}}
+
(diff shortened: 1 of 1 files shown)


AmrikSD:
I see my PR is failing because I've unticked a box, I unticked it because github calimed by keeping it ticked others would potentially be able to use it to see secrets on my repo and get to other repos, apologies.

Anyway, I'm happy to re-raise a different PR with it ticked, or just let you cherry pick these changes but I'm wondering is this even a thing you'd be interested in merging in @yagebu?

yagebu:
Hi, thank you for submitting a PR.

This is something that I do not want to merge. I do not want to get into the business of distributing docker images and take on the responsibility of dealing with possible vulnerabilities of other Python packages shipped in the image and other system libraries in the underlying image. I've personally never used the Dockerfile and it has been not been consistently maintained by the community.

I also do not really what is to be gained by distributing the image, which is not being reproducibly built - the install of Fava should work well enough with a one-line call to pip or rather should really involve another tool like pip-tools, pipenv, pdm, poetry or the like to ensure a reproducible set of dependencies if one wants to use it deploy Fava.

PR #1642, opened 2023-07-23
remove deprecated charts

yagebu:
- clipboard: use standard web APIs to copy to clipboard
- lint: enable eslint-plugin-deprecation
- lint: fix deprecation
- charts: remove deprecated way to add charts
- _layout: minor cleanup


mechanarchy:
Hi @yagebu, I am trying to update some broken custom extensions and I see that they are currently receiving the deprecated chart warning in v1.25. It looks like this PR removes the deprecated method, can you please point me to some docs or give me a quick overview of the new/correct method for rendering custom charts?

Thanks in advance!

yagebu:
See https://github.com/beancount/fava/pull/1491

PR #1655, opened 2023-08-02
Add .js route to application.py

### LINKED ISSUE #1653: Charts and menus not showing up in fava 1.25
The browser (both Firefox 115.0.3 and Edge 115.0.1901.188) refuses to load `app.js` with the following error message showing up in the console:

```
Failed to load module script: Expected a JavaScript module script but the server responded with a MIME type of "text/plain". Strict MIME type checking is enforced for module scripts per HTML spec.
```

I tried reinstalling fava using `pip` but the problem persists. There is no reverse proxy in place -- the URL I used to access fava is `http://127.0.0.1:5000`. The operating system I use is Windows 10 22H2.

What should I do to further investigate?

dousha:
A really sketchy (but working) way to override JS MIME type. Closes #1653.

### DIFF
--- a/src/fava/application.py
+++ b/src/fava/application.py
@@ -37,6 +37,7 @@ from flask import render_template_string
 from flask import request
 from flask import send_file
 from flask import url_for as flask_url_for
+from flask import send_from_directory
 from flask_babel import Babel  # type: ignore[import]
 from flask_babel import get_translations
 from markupsafe import Markup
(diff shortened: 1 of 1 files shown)


yagebu (changes requested):
This is too sketchy for my tastes, could you try something less invasive like the following?: https://github.com/pallets/flask/issues/1045#issuecomment-417670216

PR #1656, opened 2023-08-04
Fix editor error line number handling

tschicke:
This fixes a small bug in the logic to clamp an error's line number in the editor. Currently the line number gets forced to state.doc.lines because the min and max calls are flipped, so all errors show up at the end of the file.


yagebu (approved):
Thanks :)

PR #1658, opened 2023-08-06
Use a separate Jinja env per extension

andreasgerstmayr:
Hi, I updated my two Fava extensions to load the CSS via
```
<style type="text/css">{% include "style.css" %}</style>
```
(possible since #1643)

However, I noticed that whenever I load extension A first, extension B gets the wrong stylesheet, and vice-versa. Turns out Jinja's Environment caches templates by name (https://github.com/pallets/jinja/blob/86f28a9df0a97a3d3bfa3785b082651e2a8e994d/src/jinja2/environment.py#L958) and therefore the `style.css` file of the extension which was loaded first gets in the cache.

I've updated the code to use a (overlay-)Environment per extension, which resolves this issue.

### DIFF
--- a/src/fava/application.py
+++ b/src/fava/application.py
@@ -24,7 +24,6 @@ from urllib.parse import urlencode
 from urllib.parse import urlparse
 from urllib.parse import urlunparse
 
-import jinja2
 import markdown2  # type: ignore[import]
 from beancount import __version__ as beancount_version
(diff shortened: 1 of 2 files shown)


yagebu (approved):
Thanks! I've refactored this a bit to use a cached_property on the extension.

andreasgerstmayr:
Thanks for the review & merging :)

PR #1663, opened 2023-08-23
Add option to skip duplicates when exploring entries on import extract

zenitraM:
Hi! I use the Importer UI with some API-driven importers, for which the downloaded files are really only config files yet they read the full history from the origin bank APIs (like https://tariochbctools.readthedocs.io/en/latest/importers.html#nordigen).

Those output a big chunk of past entries, sometimes from multiple banks and not necessarily in chronological order. And, most of the time, a big percentage of those entries will be duplicated - in some importers the de-duplication is done by the importer against the bank's entry ID so I'm almost completely sure they are not false positives.

So I'm only interested (most of the time) on checking the non-duplicate entries, however in some cases (like combining multiple banks into a same importer, or reconciling a long tail of history) they will be 4-5 new/non-duplicate entries out of a list of >400 entries, in the middle of the list.

Browsing them and finding them with the next/back buttons is a bit awkward, so to make this easier, I've added a checkbox that will only make the prev/next buttons in the Extract window move only across the non-duplicate entries, which is enabled by default - which I think may be a more desired standard behavior?

It looks like this on the UI:

image
The checkbox make the buttons go to the next and previous non-duplicate items - and if there's none left, they will just go to the first/last entry (to finalize the import, in the latter case).

I found this quite useful for my workflow but it may not fit everyone's usage or the project's alikes - happy to address any feedback or tweaks. :blush:

### DIFF
--- a/frontend/src/reports/import/Extract.svelte
+++ b/frontend/src/reports/import/Extract.svelte
@@ -20,18 +20,11 @@
     currentIndex = 0;
   }
 
-  function submitOrNext() {
-    if (currentIndex < entries.length - 1) {
-      nextEntry();
-    } else {
-      save();
-    }
-  }
-
   function nextEntry() {
     if(skipDuplicates) {
-      for (let index = currentIndex+1; index < entries.length; index++) {
-        if (!isDuplicate(entries[index]!)) {
+      for (let index = currentIndex+1; index < entries.length; index+=1) {
+        const newEntry = entries[index];
+        if (newEntry !== undefined && !isDuplicate(newEntry)) {
           currentIndex = index;
           return;
         }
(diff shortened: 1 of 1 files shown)


yagebu:
Hi, thanks for the PR!

It might be confusing to have both "ignore duplicate" and "skip duplicates", those sound quite similar to me. Would maybe having this behaviour triggered when Ctrl-Clicking on the button also be an option instead of having the checkbox?

zenitraM:
Agree, I found it a bit confusing too but couldn't think how to express it better.
What about something like this? With `>` used for the next entry, `>>` to jump to the next non-duplicate.

image

zenitraM:
I finally built some scripts to do imports automatically (now I need time to opensource them...), so I no longer depend on Fava to import - if someone is willing to take on this work, feel free to take the changes in my branch!

PR #1670, opened 2023-09-05
tree table: restore v1.25 layout

andreasgerstmayr:
After upgrading to Fava 1.26 the layout of the Income Sheet changed, and all three columns are now in the same row, making them small:

### Fava v1.26
fava1 26

### Fava v1.25
fava1 25

### With changes from this PR
Almost pixel-for-pixel identical to v1.25:
pr

The sign of the Net Profit chart changed behavior, in v1.25 it was positive if income is greater than expenses and the `invert-income-liabilities-equity` option is set, but in v1.26 the sign is negative. This PR changes that, but it currently only works when the language is set to English. Should I create a `$options.name_net_profit` variable with the localized net profit? Alternatively, it could be flipped in the backend (`Tree.net_profit()` or `get_income_statement()`).

### DIFF
--- a/frontend/src/stores/accounts.ts
+++ b/frontend/src/stores/accounts.ts
@@ -1,5 +1,7 @@
 import { derived } from "svelte/store";
 
+import { _ } from "../i18n";
+
 import { account_details, fava_options, options } from ".";
 
 /** Whether an account should be collapsed in the account trees. */
@@ -16,7 +18,8 @@ export const invert_account = derived(
       ? (name) =>
           name.startsWith($options.name_income) ||
           name.startsWith($options.name_liabilities) ||
-          name.startsWith($options.name_equity)
+          name.startsWith($options.name_equity) ||
+          name === _("Net Profit")
       : () => false,
 );
 
--- a/frontend/src/reports/tree_reports/IncomeStatement.svelte
+++ b/frontend/src/reports/tree_reports/IncomeStatement.svelte
@@ -17,9 +17,14 @@
 {/if}
 
 <div class="row">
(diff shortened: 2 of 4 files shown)


frontend/src/stores/accounts.ts
```diff
@@ -16,7 +16,8 @@ export const invert_account = derived(
       ? (name) =>
           name.startsWith($options.name_income) ||
           name.startsWith($options.name_liabilities) ||
-          name.startsWith($options.name_equity)
+          name.startsWith($options.name_equity) ||
+          name === "Net Profit"
```
yagebu, line 20: simply compare to the translated string here `_("Net Profit")`

yagebu (changes requested):
Thanks for the PR.

> all three columns are now in the same row

They were always on the same row, they just happened to wrap before on your screen size. I adjusted the width of these cells to waste less space and that's a change that I'd like to keep. If you want to change the layout, we should rearrange these rows/columns explicitly (like it is done on the balance sheet for example).

frontend/src/stores/accounts.ts
andreasgerstmayr: D'oh! I didn't realize that the translations are available in the frontend. I'll update the PR, thanks!

yagebu (approved):
Since the Net Profit which is basically just Income + Expenses is inverted, we're treating it like Income, so I think it's better to have it in that column as well. I've made that change.

andreasgerstmayr:
Thank you! And thank you so much for your efforts in maintaining Fava, it's awesome 😃

PR #1680, opened 2023-09-17
Add instruction to install Babel for development

ludizhan:
I was following the development guide and encounter the following error in the step running `make`. 

```
...
pybabel compile -d src/fava/translations
make: pybabel: No such file or directory
make: *** [src/fava/translations/bg/LC_MESSAGES/messages.mo] Error 1
```

Figured out I could avoid the error with Babel installed by `pip install Babel` first.

Thought it would make sense to add this extra information in the development guide, or please let me know if I should have this fixed elsewhere.


yagebu:
Thanks :)

PR #1683, opened 2023-09-21
for pages with exactly two graphs: fix graph cycling shortcuts

### LINKED ISSUE #1677: 1.26 regression: cycle shortcut on account page
I noticed that with version 1.26, the shortcut 'c' to cycle between graphs no longer works on the account page. For example:

`http://localhost:2020/<my-filename>/account/Aufwaende:30-Haushalt/`

I can use Shift+C to cycle backwards, and I can also use C on other pages to cycle forwards, but not on this page. Version 1.25 did not have this issue. The dev tools console in Firefox shows no message.

MJochim:
fixes #1677

When a page has exactly two graphs, the two shortcuts for cycling to the previous and next graph need to be bound to the same graph. This is one way to allow this.

However, only the last instance of use:keyboardShortcut will show up in the HTML attribute data-key. That’s why I have chosen this shortcutForward() to be the last of the two.

*There are several ways to do this, I have chosen one with minimal code changes. A larger approach might be to allow an array of KeySpecs in keyboardShortcut. I am not sure whether you use the HTML attribute `data-key` in any way. It is not incomplete with my approach.*

### DIFF
--- a/frontend/src/charts/ChartSwitcher.svelte
+++ b/frontend/src/charts/ChartSwitcher.svelte
@@ -15,20 +15,18 @@
     charts.find((c) => c.name === $lastActiveChartName) ?? charts?.[0];
 
   // Get the shortcut key for jumping to the previous chart.
-  $: shortcutBackward = (index: number): KeySpec | undefined => {
-    const current = charts.findIndex((e) => e === active_chart);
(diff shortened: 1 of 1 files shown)


yagebu (approved):
Thank you for this fix!

PR #1685, opened 2023-09-23
add hooks for extension development: after_load_file and before_request

MJochim:
I have added two hooks to the extension system that I need in my private plugin. I think they could be useful to others as well.


yagebu (approved):
Thanks for the PR, LGTM :)

Issue #1477, opened 2022-09-04
Idea: Automatically open additional posting lines if not transaction is not balanced


fdw:
Sometimes, a transaction has three or more postings. In the current dialog, two posting lines appear by default, and I have to add one manually when I need one.

While this works, the button to add a new one is on the very right and on small screen (phone, for example), I have to scroll. For some additional comfort, I'd love Fava to recognize that the transaction is not balanced yet and add another posting automatically.

Is this something you find valuable, or is it too complex?

yagebu:
This shouldn't be too hard to do (for the simple case of postings without cost and maybe all of the same currency). A PR would be welcome :)

fdw:
Do you have a hint for me how/where this could be done? I'm still *very* new to Vue.

fdw:
Thanks for the pointers, they're very helpful 🙂

But to make this work, I need to sum the `Posting`s using their currency. Is there already a method to do that?

yagebu:
No there isn't any such logic in the frontend yet (so far the amounts are only treated as strings). You'll have to split the amounts, convert to a number, group by currency and then sum them.

fdw:
That means I would have to re-implement [Beancount's balancing](https://beancount.github.io/docs/beancount_language_syntax.html#balancing-rule-the-weight-of-postings)... do you have a pointer where I can leave such a general function that extracts the amount out of a posting? Can I do it in the same `Posting.svelte`?

yagebu:
I think `entries.ts` would be the right place for it. Feel free to refactor Amount and Posting to classes if that helps implementing that functionality.

fdw:
Sorry, one more question: I've noticed that when I try to enter a total price (in contrast to a per-unit price) using `{{<blah>}}` double braces, it's transformed into `{ 0 # <blah>}`. It works fine, but I can't find either syntax in the Beancount documentation. Have I used it wrong the whole time?

yagebu:
I'd be fine if this logic in the frontend didn't support cost at all (and ignored anything between the {})

fdw:
So, I've finally opened a PR for the simple case. I hope this is okay 🙂

sohamshanbhag:
A simpler fix for the usability problem would be to add an empty line whenever all lines in the input have text and/or amounts(irrespective of if they balance). Then, you could change the default 2 line input to one line input.

fdw:
[...]

PR #1691, opened 2023-10-03
Add new postings automatically if needed

fdw:
When the current postings don't balance, add a new line automatically.

This currently only works for simple amounts without costs or prices.

Issue: #1477

### DIFF
--- a/frontend/src/entry-forms/Transaction.svelte
+++ b/frontend/src/entry-forms/Transaction.svelte
@@ -91,6 +91,17 @@
       entry.postings = entry.postings;
     }
   }
+
+  function addPostingIfUnbalanced() {
+    const sum = entry.postings.reduce((agg: number, cur: Posting) => +cur.amount + agg, 0);
+    if (entry.postings.every((it: Posting) => it.amount !== "") && sum !== 0) {
+      addPosting();
+    }
+  }
+
+  function init(node: HTMLElement) {
+    node.focus()
+  }
 </script>
 
 <div>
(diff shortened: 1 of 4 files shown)


fdw:
Hey, @yagebu, could I please get a review? 🙂

frontend/src/entry-forms/Transaction.svelte
```diff
@@ -91,6 +91,17 @@
       entry.postings = entry.postings;
     }
   }
+
+  function addPostingIfUnbalanced() {
+    const sum = entry.postings.reduce((agg: number, cur: Posting) => +cur.amount + agg, 0);
```
yagebu, line 96: This does not even work for simple amounts `+"20 EUR"` is `NaN. This needs to do some more work to actually get the "amount" from the posting in a way so that i can be summed.

Please also add tests for this logic.

frontend/src/entry-forms/Transaction.svelte
```diff
@@ -91,6 +91,17 @@
...
+    if (entry.postings.every((it: Posting) => it.amount !== "") && sum !== 0) {
+      addPosting();
+    }
+  }
+
+  function init(node: HTMLElement) {
+    node.focus()
```
yagebu, line 103: This auto-focus behaviour should be avoided IMHO

yagebu (changes requested):
Thanks for the PR. I do not think the current implementation works well, see my comments.

yagebu:
Fixed in #1762 :)

PR #1694, opened 2023-10-11
Fix the style issue of cost in journal table.

kba977:
Adjust the spacing between price and cost so that the date in the cost is not compressed to the point where it cannot be seen.

e.g., in the figure below
image

This is followed by the effect of the adjustment
image

### DIFF
--- a/frontend/css/journal-table.css
+++ b/frontend/css/journal-table.css
@@ -117,7 +117,7 @@
 }
 
 .journal p > .num {
-  width: 9rem;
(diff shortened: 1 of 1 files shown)


yagebu:
Hi, thanks for the PR.

This is an area that needs improving, but this new fixed does not work on screens that aren't huge. I'll create a PR setting this to a relative value (which will be larger then the current one on wide enough screens).

yagebu:
See #1800

Issue #1407, opened 2022-03-31
Support for multiple levels of currency conversions


leipert:
I am new to beancount! Thank you for this great visualization tool!

Given a simple beancount file with three commodities: EUR, USD and some imaginary stock called `HOOL` which is tracked in `USD`.

```
option "operating_currency" "EUR"
option "operating_currency" "USD"

2022-01-01 open Equity:Opening-Balances
2022-01-01 open Assets:US:HOOL
2022-01-01 open Assets:Cash

2022-01-30 * "Some stock" ""
  Equity:Opening-Balances -100 HOOL
  Assets:US:HOOL 100 HOOL

2022-01-30 * "Some USD" ""
  Equity:Opening-Balances -100 USD
  Assets:Cash 100 USD

2022-01-28 * "Some EUR" ""
  Equity:Opening-Balances -100 EUR
  Assets:Cash 100 EUR

2022-01-30 price HOOL 100 USD
2022-01-30 price EUR 1.25 USD
2022-02-28 price HOOL 200 USD
2022-02-28 price EUR 1.12 USD
```

When Converting to USD, both HOOL and EUR are converted, but when converting to EUR, only USD is converted, but not HOOL:

| USD | EUR |
| - | - |
| <img width="735" alt="USD" src="https://user-images.githubusercontent.com/2906107/161144763-7027fcc2-ea5d-4f08-ab86-3c6fa5d15fa1.png"> | <img width="731" alt="EUR" src="https://user-images.githubusercontent.com/2906107/161144801-62c67746-5660-4dab-a7af-51b95b69120a.png"> | 

I had a quick look at the source and seems like that `beancount.core.convert.convert_position` only resolves "one level deep", there is also a `beancount.core.convert.convert_amount` which supports multiple levels of conversion, but not for positions, but just amounts.

In my mind there are multiple solutions:

1. Workaround, I could also put HOOL prices in EUR into the DOC
2. Extend the function on fava side (e.g. by permutating the pricemap, or extending the existing function, e.g. like with `get_market_value`)

carderne:
It will work if you change it as below. i.e. specify the cost.
I can't remember the mechanics of why (presumably the conversion is based on cost and not price) but this is a more realistic entry to have in your journal anyway.

```diff
@@ -7,6 +7,6 @@
 
 2022-01-30 * "Some stock" ""
-  Equity:Opening-Balances -100 HOOL
-  Assets:US:HOOL 100 HOOL
+  Assets:Cash
[...]

PR #1697, opened 2023-10-14
[Feature]: Support nested price conversion

fengkx:
Support nested price conversion Related issue #1407 

Always convert to biggest value using Bellman–Ford algorithm.

For example, when using **convert to HKD**, find the shortest(Smallest Conversion Rate) from HKD to others currency. $K$
then the price from others to HKD is $\frac{1}{K}$ (largest)

- [x] Unit test, linter
- [x] Description, comment


yagebu:
Hi, thank you for the PR.

To allow for multiple currency conversions, I prefer a more explicit approach where one either explicitly adds additional prices via a plugin (as is already possible) or by explicitly specifying the currencies to convert via/to. With #1732 the latter will also be possible. Also, the implementation here does not seem very performant, on some reports the load times seem to have doubled. So I do not intend to merge your PR - does #1732 also work for your use case?

fengkx:
I use the master branch, But can't find the correct way to use #1732.
image

I have a long list of price directive created by bean-price daily. I want all Other currency be converted into CNY when I choose converted to CNY.

yagebu:
If you have conversion rates from your other currencies to, e.g., HKD (and then prices for HKD in CNY), you could choose HKD,CNY as conversions to get everything converted to CNY.

fengkx:
I am using the main branch, But I can't find the option in the UI.
image

What if I have many other currenecies (or stocks) which also want to convert to CNY?

yagebu:
> I am using the main branch, But I can't find the option in the UI.

You can select multiple currencies from that dropdown

fengkx:
OK, I understand it now. The order in which I select affects the conversion. My pr uses the Bellman–Ford algorithm to find the shortest path, which makes it slow, while the current solution requires the user to find the correct path. I think the UI does not inform the user that the selection order would affect the results.

yagebu:
> OK, I understand it now. The order in which I select affects the conversion. My pr uses the Bellman–Ford algorithm to find the shortest path, which makes it slow, while the current solution requires the user to find the correct path. 

Exactly - it also allows/requires the user to pick a specific path. Finding shouldn't normally be hard AFAIKT, since it will usually be just via your operating currencies.

>I think the UI does not inform the user that the selection order would affect the results.

That's why I added a help page to describe it - I don't think this can be easily shown in the UI

> I have a question while I am devloping locally. How to set up the translations... It seems to require a POEDITOR_TOKEN

No, that token is only necessary to update the translations from poeditor.com (where they are edited). If you want to edit the translations, you'll have to do so at poeditor.

PR #1722, opened 2023-12-15
Add support for posting metadata

cruizh:
I wanted to be able to use posting metadata when importing. I noticed `bean-extract` generated the poting metadata correctly, but Fava's importing feature discarded posting metadata and only kept entry metadata.

This PR adds an optional `meta` field to `Posting` so that this information is not lost.
I also modified the `Posting` entry form to include a button to add metadata to the posting and show the posting metadata. This allows the user to enter posting metadata on import and on new transactions.

It looks like this on the UI:
image


src/fava/beans/abc.py
```diff
@@ -112,7 +112,7 @@ def price(self) -> Amount | None:
 
     @property
     @abstractmethod
-    def meta(self) -> Meta | None:
+    def meta(self) -> Meta:
```
yagebu, line 115: this change should be reverted - we still need to support postings with a `None` meta.

frontend/src/entry-forms/AddPostingMetadataButton.svelte
```diff
@@ -0,0 +1,22 @@
+<script lang="ts">
```
yagebu, line 1: This seems to be nearly an identical copy of the corresponding component for entries - please reuse that one

frontend/src/entry-forms/PostingMetadata.svelte
```diff
@@ -0,0 +1,107 @@
+<script lang="ts">
```
yagebu, line 1: This seems to be nearly an identical copy of the corresponding component for entries - please reuse that one

frontend/src/entries.ts
```diff
@@ -15,11 +16,19 @@ import {
 export interface Posting {
   account: string;
   amount: string;
+  meta: EntryMetadata | null;
```
yagebu, line 19: I think dropping the `| null` here and ensuring that this is always set in the frontend code even if not set in the Beancount entries might simplify reuse of the existing components

yagebu (changes requested):
Thanks for this PR :)

This looks mostly good but the duplicated code should be reduced, I've left some inline comments.

src/fava/beans/abc.py
cruizh: This should be fixed in https://github.com/beancount/fava/pull/1722/commits/33d1c8b094cd5115ff90fa3a29dd71ad70f80faa :)

frontend/src/entries.ts
cruizh: This should be fixed in https://github.com/beancount/fava/pull/1722/commits/b851eb3ee76617e2d00b63980489f354356777a2 :)

cruizh:
Thanks a lot for the quick response and the constructive criticism! I have been using Fava for a long time and appreciate all the hard work you put into it.

I have fixed the points you commented. I hope this makes it an easier change to accept.

Please let me know if I should improve it further. Again, thanks a lot!

yagebu (approved):

yagebu:
Thanks again, nice work :)

Issue #1738, opened 2024-01-16
Invalid data returned in API request: Validating union failed


przemator:
When in fava 1.27 in the Import section, I hit the Extract button and I get the following error:

`Invalid data returned in API request: Validating union failed`

Fava debugging shows no errors in the log. The API extract requests is a HTTP 200.

mekanics:
after a real quick debugging session, I found out that the validation of the api response fails at the `meta` property of a posting, as it was `undefined`. 

My guess it's that the `entry_meta_validator` should allow undefined
https://github.com/beancount/fava/compare/v1.26.4..v1.27#diff-f4c0cbf95171873a2db676ba3e01ec28b7661f21d303351ec4a49e13689bfbb2R21-R24

 or the api endpoint (`/api/extract`) should return the `meta` property in a posting

yagebu:
Thanks @mekanics for debugging this :)

przemator:
@mekanics thank you for the fix @yagebu could you tell me how I can use this fix? I'm currently unable to import any transactions using fava 1.27.

mekanics:
> @mekanics thank you for the fix @yagebu could you tell me how I can use this fix? I'm currently unable to import any transactions using fava 1.27.

Update to 1.27.1 🙂

PR #1739, opened 2024-01-19
Fix the posting validation

mekanics:
This pull request addresses the validation of the `meta` property within the `Posting` (#1738).

The API does not include a meta property in the JSON if it is empty. Therefore, the validation of this property should permit `meta` as optional in the frontend.

### DIFF
--- a/frontend/src/entries.ts
+++ b/frontend/src/entries.ts
@@ -6,6 +6,7 @@ import {
   defaultValue,
   number,
   object,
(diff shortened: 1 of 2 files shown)


mekanics:
in retrospect I think it would have been easier to just always add the `meta` property in the backend... 🙃

mekanics:
Closing in favor of #1740

PR #1740, opened 2024-01-19
[Fix] Add nullable meta to posting json

mekanics:
This pull request updates the outputted JSON to include a nullable `meta` field in the posting.

This resolves an issue in which the `meta` was not defined during the import and resolved into a `Invalid data returned in API request: Validating union failed` error . #1738


yagebu:
The `null` would still fail the validation, see #1741 for a working fix.

mekanics:
🤦🏼‍♂️ yes, of course

thx @yagebu for the real fix

PR #1744, opened 2024-01-26
properly encode document filename in url

upsuper:
When filename contains certain special characters like `#`, the URL would not be getting the right file. For example, if the document file is `/home/upsuper/beancount/receipt #1.pdf`, the URL would be `/document/?filename=/home/upsuper/beancount/receipt #1.pdf` whose meaning is requesting `/document/?filename=/home/upsuper/beancount/receipt ` and use the fragment `1.pdf`, which is wrong.

This PR makes it encode the filename as a URL component so that it would work regardless of characters used in the document filename.


yagebu (approved):
Thanks for the fix :)

Issue #1747, opened 2024-02-01
Print view improvements


slifty:
Thank you for the awesome project!

I'm generating reports for my tax accountants and having an issue printing out my income statement and balance sheet from the fava interfaces.  I think this *used* to work, so it's possible a recent release changed the print stylesheets.

Here's what the printed page looks like (notice how the bars stop at September and the expenses table is cut off early)

![image](https://github.com/beancount/fava/assets/208884/18bc1d24-98b2-4e04-924a-690f2470d910)

slifty:
I should also mention that more generally there are a few things I'd love to see improved about the print view (e.g. removing the menu items from the bar chart and showing net profit after expenses).

yagebu:
I've played around with this a bit (in Firefox). It seems that this print preview happens to quickly or freezes the page in place before the chart can re-adjust for the printed page size - in general I also find it to be not very reproducible, the layout seems to be pretty random. It seems that not setting explicit sizes on the svgs but instead just setting a viewBox works better - with that, it always seemed to include the full chart without cutting it off - so that should be fixed by 76fcf595.

PRs with improvements to the print styles (like hiding unneeded things) are always welcome :)

slifty:
Awesome and thanks so much for looking into it!

I'll take a look at improving print view further and open a PR asap (hopefully this doesn't mean Soon™, but rather, soon!)

GuillaumeHM:
Hello,

I'm trying to do the same as you: printing reports for a tier. Which "printer" program do you use ? 

I get blank pages or it works but with the right graph part missing. I tested :
- weasyprint : blank page
- wkhtmltopdf : blank page
- firefox: right part missing
- epiphany/webkit : right part missing

Portrait printing seems worse (wider part missing) than landscape.

The best would be weasyprint to work fine. Can any one explain why it doesn't ? The part missing is about a CSS property to adjust, I guess.

Thank you

slifty:
@GuillaumeHM are you using the latest version of Fava?  I know that there was a patch released recently to fix the right part.

GuillaumeHM:
Thank you. I upgraded from 1.26.2 yo 1.27.2 and here is the new status:

- weasyprint : blank page
- wkhtmltopdf : blank page
- firefox: WORKING
- epiphany/webkit : right part missing on landscape (plus datas missing on portrait, not sure if that is a regression)

[...]

PR #1753, opened 2024-02-05
Remove chart controls in print view

slifty:
This PR removes UX related to changing chart settings when printing a page.

*Print view before the PR:*
image

*Print view after the PR:*
image

Related to #1747

### DIFF
--- a/frontend/src/charts/ChartSwitcher.svelte
+++ b/frontend/src/charts/ChartSwitcher.svelte
@@ -72,4 +72,19 @@
   button:hover {
     color: var(--text-color-lighter);
   }
+
+  @media print {
+    button {
+      display: none;
+      border-left: none;
+    }
+
+    button + button {
+      border-left: none;
+    }
+
+    button.selected {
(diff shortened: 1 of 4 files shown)


frontend/src/charts/ChartSwitcher.svelte
```diff
@@ -72,4 +72,10 @@
   button:hover {
     color: var(--text-color-lighter);
   }
+
+  @media print {
+    button {
```
yagebu, line 77: mhh, let's maybe still show the currently selected one so that the chart has a label?

slifty: Ah right!  I think the "line chart" doesn't provide much value, but the "Net Worth" absolutely would, what do you think of that approach, or would you prefer both?

slifty: Realizing that's what you were asking for, and have updated

yagebu (approved):

PR #1759, opened 2024-02-12
Remove price directive when deciding start date

### LINKED ISSUE #1754: Suggestion: Remove price directive when deciding start date of the graph
I first posted this 2 weeks back in #1119, but that issue is closed, and I'm not sure if github informs the owner when someone comments on a closed issue. This is a copy of the same.

---

From the git history, it seems that #1119 was the issue which prompted adding Price directives for determining time axis limits. This was done so that variation in price would be shown on the chart, even when no transactions have taken place. Moreover, in #1593, the code was changed from using beancount.core.getters.get_min_max_date to using the code of beancount.core.getters.get_min_max_date directly. In that case, can the code be changed such that price is only considered when selecting _date_last, and not _date_start? The present implementation in src/fava/core/\_\_init\_\_.py, i.e.

    for entry in self.entries:
        if isinstance(entry, (Transaction, Price)):
            self._date_first = entry.date
            break
    for entry in reversed(self.entries):
        if isinstance(entry, (Transaction, Price)):
            self._date_last = entry.date + timedelta(1)
            break

leads to a large part of the graph in the start containing no data if the price values for those are given (which happen in my file due to automatically fetching historical data). The plot with Price in filter:
2024-01-25_01-21-48_screenshot

vs the one with Price not in filter:
2024-01-25_01-22-24_screenshot

(There is a very small transaction at the start of the second plot hence the plot looks empty there)

The change for the same would be

    for entry in self.entries:
        if isinstance(entry, (Transaction,)):
            self._date_first = entry.date
            break
    for entry in reversed(self.entries):
        if isinstance(entry, (Transaction, Price)):
            self._date_last = entry.date + timedelta(1)
            break

Making this change would also not affect #1119 since there would be no assets whose values would be changing before the first transaction.

_Originally posted by @sohamshanbhag in https://github.com/beancount/fava/issues/1119#issuecomment-1908485981_

sohamshanbhag:
The present implementation shows a large empty plot before the first transaction. This PR corrects it.

Fixes #1754.

### DIFF
--- a/src/fava/core/__init__.py
+++ b/src/fava/core/__init__.py
@@ -109,7 +109,7 @@ class FilteredLedger:
         self._date_first = None
         self._date_last = None
         for entry in self.entries:
-            if isinstance(entry, (Transaction,)):
+            if isinstance(entry, Transaction):
                 self._date_first = entry.date
                 break
         for entry in reversed(self.entries):
--- a/src/fava/core/__init__.py
+++ b/src/fava/core/__init__.py
@@ -109,7 +109,7 @@ class FilteredLedger:
         self._date_first = None
         self._date_last = None
         for entry in self.entries:
-            if isinstance(entry, (Transaction, Price)):
+            if isinstance(entry, (Transaction,)):
                 self._date_first = entry.date
                 break
         for entry in reversed(self.entries):


src/fava/core/__init__.py
```diff
@@ -109,7 +109,7 @@ def __init__(
         self._date_first = None
         self._date_last = None
         for entry in self.entries:
-            if isinstance(entry, (Transaction, Price)):
+            if isinstance(entry, (Transaction,)):
```
yagebu, line 112:
```suggestion
            if isinstance(entry, Transaction):
```

yagebu (approved):
Thanks :)

PR #1760, opened 2024-02-13
support fiscal year that starts one year later

### LINKED ISSUE #1719: Fiscal Year cannot end in the next year of Calendar Year
In my country (Japan), FY2023 starts at 2023-04-01 and ends at 2024-03-31, but Fava option cannot describe this.
`custom "fava-option" "fiscal-year-end" "03-31"` defines FY2023 as `[2022-04-01, 2023-04-01)`, and this is not my intention.

lo48576:
In some countries (like Japan), FY2024 starts at 2024-mm-dd and ends at 2025-mm-dd. This patch adds support to this kind of fiscal year system by accepting values such as `15-31` as the fiscal year end.

Closes #1719.

* * *

I don't usually write Python so please feel free to rewrite if something looks unnatural.

I disabled the lint PLR0912 (too-many-branches) for `substitute()` in src/fava/util/date.py, because it has already nearly maximum branches, and the code became harder to read when I tried to avoid this lint by using complex inline if statements.


yagebu (approved):
Thank you for the PR! I've refactored the code a bit - which was easy to do since you added good test coverage :)

PR #1761, opened 2024-02-17
Automatically add posting field when last posting is non-empty

### LINKED ISSUE #1477: Idea: Automatically open additional posting lines if not transaction is not balanced

sohamshanbhag:
This commit checks if information has been added to the last posting
when entering a transaction and adds an empty posting in this case.

I've run make check and make lint, but I'm not a typescript/svelte developer so please check the changes.

Partially fixes #1477.

### DIFF
--- a/frontend/src/entry-forms/Posting.svelte
+++ b/frontend/src/entry-forms/Posting.svelte
@@ -14,7 +14,6 @@
   export let date: string | undefined;
   export let move: (arg: { from: number; to: number }) => void;
   export let remove: () => void;
-  export let add: () => void;
 
   $: amount_number = posting.amount.replace(/[^\-?0-9.]/g, "");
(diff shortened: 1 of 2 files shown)


sohamshanbhag:
I've divided the commits such that one adds the automatic posting part, and another removes the + and the p buttons since they seem to be useless now. Feel free to discard the second one if you want to keep the functionality. Also, it seems redundant to have both the p button and the + button.

PR #1762, opened 2024-02-17
 Automatically add posting field when last posting is non-empty

sohamshanbhag:
This commit checks if information has been added to the last posting
when entering a transaction and adds an empty posting in this case.

I've run make test and make lint, but I'm not a typescript/svelte
developer so please check the changes.

I've divided the commits such that one adds the automatic posting part,
another removes the + and the p buttons since they seem to be useless
now, and another removes the default value of two postings. Feel free to
discard any ones if you want to keep the functionality.

Note that adding the automatic posting part also removes the 
'Allow entry to be empty' commit to work.

Also, it seems redundant to have both the p button and the + button.

Partially fixes https://github.com/beancount/fava/issues/1477.

### DIFF
--- a/frontend/src/entries.ts
+++ b/frontend/src/entries.ts
@@ -204,7 +204,7 @@ export class Transaction extends EntryBase {
     this.narration = "";
     this.tags = [];
     this.links = [];
-    this.postings = [emptyPosting(), emptyPosting()];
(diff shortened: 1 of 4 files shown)


sohamshanbhag:
Thanks for the application. I have no experience with js/svelte, so feel free to modify it as you see fit.

PR #1768, opened 2024-03-01
Add tooltip for add journal entry button

Blquinn:
Make the "Beancount" header at the top left a navigable button (takes you to income_statement). I think most people expect that to take you "home". As part of that I refactored the Link behavior into an un-styled shared component so that there can be links that don't have the sidebar styling.

I also added a tooltip to the '+' icon in the sidebar.

By the way, there was something messed up with prettier where it was formatting a file in pre-commit and then failing the lint after on some css rule in SidebarLink.svelte:32.


Blquinn:
Ah ok. I can remove the clickable header code and leave just the tooltip, if you'd be interested in merging that.

Also with regard to the linting issue. What I'm saying is that prettier is reformatting that code into a format that the linter rejects. So their configurations are conflicting somehow.

yagebu:
> Ah ok. I can remove the clickable header code and leave just the tooltip, if you'd be interested in merging that.

Yep, I'll add that change, thank you :)

> What I'm saying is that prettier is reformatting that code into a format that the linter rejects. So their configurations are conflicting somehow.

That shouldn't really happen. The other linters should not have any formatting rules enabled. Do you have some concrete example of how to trigger this, then I could try to reproduce this?

Blquinn:
Sorry for leaving the PR open without finishing it up.. forgot about this one!

Also the linting thing isn't happening anymore -- I cloned it on a different computer and I don't have the issue here. Maybe it only happens on windows, or when the file has been touched? I'm not sure. Probably nothing to worry about unless someone else experiences it.

PR #1774, opened 2024-03-12
Sort accounts store.

### LINKED ISSUE #1773: Accounts autocomplete is not sorted
For example (this is from the output of bean-example, using Fava built at HEAD):

Screenshot from 2024-03-11 20-22-45

I think I have a fix for this. I'll attach a PR.

umangv:
With this change, the accounts autocomplete (both in the sidebar and entry form) shows a sorted list of accounts.

I don't have a ton of experience with TS/Svelte so any pointers would be appreciated. Specifically, I don't know if it makes a difference whether sorting accounts, which is already a derived store, should be done (a) in place, (b) on a copy, or (c) upstream of the accounts store (maybe in the ledgerData store?). If you have thoughts, let me know, I'm happy to dive more into the Svelte docs and push a different solution.

This should fix  #1773

### DIFF
--- a/frontend/src/stores/index.ts
+++ b/frontend/src/stores/index.ts
@@ -47,7 +47,7 @@ export const base_url = derived(ledgerData, (v) => v.base_url);
 export const extensions = derived(ledgerData, (v) => v.extensions);
 
 /** The ranked array of all accounts. */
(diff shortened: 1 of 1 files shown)


umangv:
See comment in #1773. This is working as intended.

Issue #1778, opened 2024-03-18
WatchfilesWatcher stops watching a file that is replaced


josephw:
The new `WatchfilesWatcher` implementation (fb59849ccd45) has the issue reported in https://github.com/samuelcolvin/watchfiles/issues/235 . If the file is recreated, as some editors will do on saving, it will no longer be watched.

On saving a Beancount file using `vi`, a single event of `<Change.deleted: 3>` shows up, and causes the reload indicator to show in the UI, but nothing will show after that.

The workaround suggested in [that report](https://github.com/samuelcolvin/watchfiles/issues/235#issuecomment-1633622058):
> I would watch the entire directory with `recursive=False`, then check which file changes in your own logic.

works fine. e.g.

```python

files = ['some-file', 'another-file']

watch_dirs = set([Path(p).parent for p in files])

for changes in watch(*watch_dirs, recursive=False):
  for c in changes:
    if c[0] in [Change.added, Change.modified]:
      for f in files:
        try:
          if Path(f).samefile(c[1]):
            print(f)
        except FileNotFoundError:
          pass
```

Adding this approach to the existing behaviour isn't a trivial change, due to  `WatchfilesWatcher` taking both `files: Iterable[Path]` and  `folders: Iterable[Path]`. The implementation would need to determine the effective set of directories to watch, then filter the events to check for an exact match for targets specified by `files`, but accept anything that was specified by `folder`, which may require more state in `_WatchfilesThread`.

yagebu:
Fixed in #1781, thanks @josephw

PR #1781, opened 2024-03-24
Continue to watch files that are replaced.

josephw:
Fix for #1778.

Continue to watch files are replaced, such as by saving in vi, by watching the parent directory instead. This moves more logic into the watcher, and is the fix suggested in https://github.com/samuelcolvin/watchfiles/issues/235#issuecomment-1633622058 . To preserve the existing behaviour for directory watching, the watch remains recursive.

### DIFF
--- a/src/fava/core/watcher.py
+++ b/src/fava/core/watcher.py
@@ -6,7 +6,6 @@ import abc
 import atexit
 import logging
 import threading
-from itertools import chain
 from os import walk
 from pathlib import Path
 from time import time_ns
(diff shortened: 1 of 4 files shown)


src/fava/core/watcher.py
```diff
@@ -18,28 +17,45 @@
...
         atexit.register(self.stop)
 
-        for changes in watch(*self.paths, stop_event=self._stop_event):
-            for _, path in changes:
+        watch_paths = {f.parent for f in self._file_set} | self._folder_set
+
+        for changes in watch(
+            *watch_paths,
+            stop_event=self._stop_event,
+            ignore_permission_denied=True,
```
josephw, line 55: This causes the "old_deps" builds to fail. It's here for https://github.com/samuelcolvin/watchfiles/pull/224 ; it's only necessary if you end up watching a directory that contains unreadable files.

yagebu: That's a legimitate failure - we either need to bump the required version of watchfiles or write this in a way such that it doesn't fail with older versions. I guess bumping the version would be fine and easier.

yagebu (changes requested):
Thanks for the PR, this looks quite good, we'll need to address the old_deps failure though, see my inline comment.

Also, could you add a test that shows that this works?

src/fava/core/watcher.py
josephw: It's probably not a dealbreaker: this is only necessary in the case that the directory with the Beancount file also contains files you cannot read; I hit this testing with `/tmp`, and added this in response.

However, this only takes the minimum version from [`0.19.0`](https://github.com/samuelcolvin/watchfiles/releases/tag/v0.19.0) to [`0.20.0`](https://github.com/samuelcolvin/watchfiles/releases/tag/v0.20.0), which seems reasonable so I've bumped that in the PR.

josephw:
I've added a test, which fails on the last line due to the expected timeout with the current version. There's a `time.sleep` in there which isn't ideal, but it seems like the simplest way to ensure the file's modification time is greater than the previous clock read.

yagebu (approved):
Thanks :)

> There's a time.sleep in there which isn't ideal, but it seems like the simplest way to ensure the file's modification time is greater than the previous clock read.

I've reduced the sleep time a bit but yeah, I also haven't yet thought of a great way to remove those sleeps.

josephw:
Thanks.

(If I _had_ to remove those sleeps, I'd look at mocking the time check, or setting the file's timestamp to the future. Both add a degree of complexity that I don't think is worth it.)

PR #1797, opened 2024-04-28
fix: make the link drop feature actually work

blaggacao:
This was really all it took to make the feature fork.

- Tested this change via live editing in ChromeDebugger
- Confirmed it worked: dragging a (discovered) doc over a TX now added the `document` meta as expected

~~EDIT: although maybe, it shouldn't be an absolute path (only unique on the local machine)?~~ It works: _just needed to take care to drop on the **correct** (same account) posting leg_.

**Recalling the Use Case:**

- User knows how to put PDFs into account-wise folders
- Accountant knows how to book bank statements
- Accountant needs an easy way to link neighboring documents to the transactions [this feature fixed]
- Uploading has disadvantages:
  - Uploading from an input folder doesn't erase the document ("what did I already assign?" is harder to track)
  - A cognizant user can do a best-effort pre-classification for the accountant which is in any case conductive to efficiency

### DIFF
--- a/frontend/src/document-upload.ts
+++ b/frontend/src/document-upload.ts
@@ -20,7 +20,7 @@ import { notify, notify_err } from "./notifications";
  */
 function dragover(event: DragEvent, closestTarget: HTMLElement): void {
   if (
-    event.dataTransfer?.types.includes("Files") ??
+    event.dataTransfer?.types.includes("Files") || // eslint-disable-line @typescript-eslint/prefer-nullish-coalescing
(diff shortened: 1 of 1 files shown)


yagebu:
Thanks for the PR with this fix :)

However, I've tried to address this without the lint-disable (see b03e585cd133c2488537e592a14c0db2e72831e0) a bit more comprehensively in #1798 by also enabling a lint rule that makes this sort of mistake in boolean expressions less likely.

PR #1799, opened 2024-04-29
add toggle to include zero in line charts

Grulfen:
In line charts I sometimes want to include zero in the graph to see the "total" state, and so that small changes dont look like large changes.

To help with this I added an "Include zero" toggle to line charts, it looks like this:

Include zero not toggled:
image

Include zero toggled:
image

If there are better ways toggle this in the UI I'm happy to update this PR.

### DIFF
--- a/frontend/src/charts/Chart.svelte
+++ b/frontend/src/charts/Chart.svelte
@@ -5,6 +5,7 @@
     barChartMode,
     chartToggledCurrencies,
     hierarchyChartMode,
+    includeZeroToggle,
     lineChartMode,
     showCharts,
     treemapCurrency,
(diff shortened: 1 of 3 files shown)


yagebu:
Hi, thanks for the PR :)

This sounds like a useful option to include - however, I'm not sure on whether we should even have an option or just make "Include zero" the default. In particular for the area chart, it seems odd that we don't always include zero.

If we have this as an option, I'd like to include it maybe behind a dropdown menu, this area above the chart is getting a bit crowded.

Grulfen:
It sounds like we have four ways to go forward:
- Make "Include zero" the default.        <--- This is my favourite, given we are ok with changing current behaviour
  - This will make the current visualization no longer available
- Make "Include zero" the default for area charts, but not line charts
  - Both "Include zero" and not possible. But not very intuitive that area charts would have this difference.
- Add a dropdown menu
- Keep current behaviour

What do you think?

Grulfen:
After some trial I have changed my mind and prefer having only area chart have include zero and line charts have the current behaviour.

I will open a new pull request for that

PR #1803, opened 2024-05-09
include zero in line charts for area mode

Grulfen:
This change makes the area mode in line charts always include zero. So that the area is not cropped.
In line mode the old behaviour (show 5% below lowest value) is kept.

This makes it easier to see how big a change in an account is in relation the the account size

Line
bild

Area
bild


yagebu:
Yeah, I think this is also a nice way to address this and have both options :)

Issue #1805, opened 2024-05-13
Unconsistency in week Date filter 


mailhot001:
The week filter is not consistent, there is a 1 week offset when appling a date filter.
For example if you rund the [DEMO](https://fava.pythonanywhere.com/example-beancount-file/income_statement/). 

Show the Expenses, and apply the daily interval, with the changes(weekly) display, the expenses are shown per week, as expected. 

![image](https://github.com/beancount/fava/assets/151871234/5e55a81f-27c8-4483-ac25-6d3e33fcef1f)

If you click on the 2017-W36 week, there will be no data shown, to show the 2017-W36 expenses you need to filter the date to 2017-W35.

I believe the proper week is 2017-W36 so there is a -1 week difference in the week date filter?

mailhot001:
In fava/src/fava/util/date.py
there is the following function:
![image](https://github.com/beancount/fava/assets/151871234/e5bd559b-1827-46c3-919c-2c9aa10ccf42)
There is a +1 in the date defining, so if a date is ISO week no. 1 it will return week 2.
So it seem to be a wanted behavious, but fava on the display use the standard ISO week format, thus the inconsistency?

mailhot001:
Test did not pass, confirming this is a wanted behaviour, so maybe we should change the week number display on weekly expenses display?
![image](https://github.com/beancount/fava/assets/151871234/362ca1da-853b-438d-ba77-3c83ee7a1313)

yagebu:
> Test did not pass, confirming this is a wanted behaviour

That's not necessarily the case - tests are like code, they have bugs too ;) The wanted behaviour here would be for frontend and backend to match up of course. IMHO both should just match what strftime does for `%W` (here's how the Python docs describe that: "Week number of the year (Monday as the first day of the week) as a zero-padded decimal number. All days in a new year preceding the first Monday are considered to be in week 0.".) - the frontend does that using d3-time-format, we should change the Python side to match.

mailhot001:
Thanks for the info, I will double check the d3 to confirm it does the same as python fromisocalendar regarding week handling and summit an updated pull request afterwards.

yagebu:
Thanks for the fix, I've merged your PR.

[...]

PR #1807, opened 2024-05-16
Update date.py 

mailhot001:
remove the + 1 on the week RE match in date parser so that the week number match the front end displayed week number.

related to: #1805


src/fava/util/date.py
```diff
@@ -382,7 +382,7 @@ def parse_date(  # noqa: PLR0911
     match = WEEK_RE.match(string)
     if match:
         year, week = map(int, match.group(1, 2))
-        start = datetime.date.fromisocalendar(year, week + 1, 1)
```
yagebu, line 385: this seems to have been my poor attempt at converting from "iso week number" to "%W-like week number".

I think we should switch to using `datetime.datetime.strptime(string, "%Y-W%W").date()`

mailhot001 (commented):
@yagebu Here is my second try, I believe it matches the D3.js week parsing, I've tested it roughly (without edge case) and it is working the same as frontend.

src/fava/util/date.py
mailhot001: For your info, the tzinfo was suggested by tox, but is useless as Tz is not present in the date we extract afterwards.

yagebu (commented):
Thanks :) Can you add some test cases that show that this is now working correctly in cases where it didn't before?

mailhot001:
I added a test case to show an error if ISO week format is used. 
Also added a note in the help to explain the week numbering used, a month is a month, but week can be taken differently. The note should help users know how it work without checking the code and then checking the %W definition of strptime. 
Or the equivalent definition used in D3.js.

Issue #708, opened 2018-03-27
Date sometimes takes two lines


tbm:
Sometimes (not often) the date is split across two lines.  Maybe you can increase the size of the date box a bit?

![date1](https://user-images.githubusercontent.com/1556037/37964690-ac8c8400-31c2-11e8-98a0-94ca468b57c0.png)
![date2](https://user-images.githubusercontent.com/1556037/37964692-accd5f70-31c2-11e8-878b-0288a6844fea.png)

tbm:
FWIW, this is in Chromium.  I don't see it in Firefox.

yagebu:
#705

tbm:
For the record, this was fixed in 95cd532

xentac:
Just wanted to point out my comment on https://github.com/beancount/fava/commit/95cd532e2a3f619c952543e733751a75b7847a86. It looks like we added nowrap to tree-table, which breaks "at Market Value" rendering.

tbm:
@xentac probably best to open a new issue for this.

PR #1809, opened 2024-05-24
Add a dark mode

dtrai2:
Inspired by the works of #1238 I thought I give it a try. I was also reading in #1084 that some suggested to just use a browser add-on, but I have to admit that I personally never was a fan of third party solutions for that.

@yagebu I tried to also consider your last remarks on #1084 that it would be nice to keep some of the original colors. I hope I was able to do that to your liking. 

Similar to the work of @arkn98 I could achieve good results by just setting some colors. At the moment I only see one big issue though and that are some texts in the editor, like comments, the currency or the price. I couldn't find where I have to set them, so maybe someone can point me into the right direction.

Here is an example of how it currently looks like:
image

And here is a screenshot of the issue I still have with the currency and the values. They do still seem kinda dark in the editor.
image

Looking forward to feedback and remarks!

### DIFF
--- a/frontend/css/editor.css
+++ b/frontend/css/editor.css
@@ -12,3 +12,28 @@
 .cm-editor .cm-placeholder {
   color: var(--placeholder-color);
 }
+
+.cm-editor .cm-tooltip {
+  background-color: var(--sidebar-border);
+}
+
+/* stylelint-disable selector-class-pattern */
+.cm-editor .cm-activeLine {
+  background-color: var(--editor-activeline);
+}
+
+.cm-editor .cm-selectionBackground {
+  background-color: var(--editor-activeline);
+}
+
+.cm-editor.cm-focused
+  > .cm-scroller
+  > .cm-selectionLayer
+  .cm-selectionBackground {
+  background-color: var(--editor-activeline);
+}
+
+.cm-editor .cm-selectionMatch {
+  background-color: var(--editor-selectionmatch);
+}
+/* stylelint-enable selector-class-pattern */
--- a/frontend/css/style.css
+++ b/frontend/css/style.css
@@ -68,16 +68,29 @@
   --budget-negative: #af3d3d;
   --budget-positive: #3daf46;
 
-  /* Editor */
-  --editor-comment: #998;
-  --editor-directive: #333;
-  --editor-class: #b84;
-  --editor-date: #099;
-  --editor-constant: #008080;
+  /* Editor elements. */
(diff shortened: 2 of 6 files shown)


yagebu:
Thanks for the PR :)

> At the moment I only see one big issue though and that are some texts in the editor, like comments, the currency or the price

- currencies: seems we just hard-code that to `#708` right now, see https://github.com/beancount/fava/blob/c349ee4776adac6a9d9122271473e681cd3cf3dd/frontend/src/codemirror/beancount-highlight.ts#L32 - so adding a CSS variable for that just like the other editor colors and then overriding that should do the trick.
- number, ...: similar, but we're not setting that color in Fava explicitly AFAICT - so you would need to add an entry to beancount-highlight.ts for that with a CSS variable as well

dtrai2:
Thanks for your feedback, I incorporated it. :)

How do you feel about the current look and feel of the dark mode? Any suggestions or ideas I should still consider?

yagebu:
Thanks for adding these colors. Here's some things that I noticed that don't look quite right yet:

Some of the contrasts in the journal are very low (maybe darken the colors?):

image

Active / inactive buttons are hard to tell apart (is it intentional that the active buttons aren't blue as they are in the light color scheme?):

image

The query report, very low contrasts:

image

dtrai2:
So I did manage to fix the low contrast for the query editor as well. In the meantime I did notice another thing though. When selecting some lines in the regular editor the background becomes really bright such that it is not possible to read some words:

I did try to find out where to configure this, but apparently my front-end skills don't seem to suffice. @yagebu maybe you have a hint for me here?!

yagebu:
> I struggled to find out how to define the que entry you have in your screenshot though. Maybe you can give me a snippet or a hint on how I can add it to my local test file.

That's a screenshot from one of the test files (example.beancount) - you can startup Fava with those files with `make run-example`.

> When selecting some lines in the regular editor the background becomes really bright such that it is not possible to read some words:

That can be styled as `.cm-selectionBackground`, I've added classes for that.

I've also adjusted the colors for the journal a bit, to simplify the CSS I've darkened the colors there some more and also switched them to the `hsl()` syntax which makes it easier to just tweak the "lightness" for example while leaving hue (and sometimes saturatation) the same as in the light theme.

I've also squashed all the commits and will merge this, thanks for the work on this :)

dtrai2:
Ah thanks for helping out and merging the changes! 👍

Issue #1145, opened 2020-08-29
Faster transaction inserting: Allow looking up past transactions by Narration field


pmarciniak:
Currently, fava allows easily looking up previously added Payees with just a few keystrokes, and populating the transaction template based on the last transaction of the chosen Payee. This is awesome and a great help when manually entering transactions.

It would be even more awesome if the same feature was supported in the Narration field. 

I often have a few recurring transaction "classes" for each payee. In order to populate the transaction template quickly for the correct transaction class, I ended up adding fake payee suffixes per transaction class. Supporting lookup by Narration field would allow eliminating this hack. Furthermore, it would bring this time-saving feature to people who don't use payees.

yagebu:
Sounds good. A PR would be welcome.

PR #1816, opened 2024-05-28
add auto-completion for narration

### LINKED ISSUE #1535: Auto-completion in the "Narration" field in the Add popup
Hi,
is it possible to display suggestions when typing in the "Narration" field in the Add popup?
Like start typing `#sal` and it'd suggest `#salary` tag.
Auto-completion works on any other field but that one. 
It'd especially useful on mobile, for now I'm forced to use the editor which can be problematic.

dtrai2:
Hi, I just gave it a shoot at the auto-completion feature for the narration text field. Please have a look at it. 

Disclaimer: I am not a frontend developer so any feedback is appreciated. Especially when it comes to the `valueExtractor` and `valueSelector`. I simply copy-pasted those functions, but I am sure that there is a smarter way to prevent code-duplication.

closes #1535, #1145

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -124,6 +124,7 @@ disable = [
     "too-many-instance-attributes",
     "too-many-locals",                # is checked by ruff (PLR0914)
     "duplicate-code",
+    "unused-argument",                # is checked by ruff (ARG001)
     "stop-iteration-return",
     "invalid-unary-operand-type",    # type-checking like, had false-positives
     "not-an-iterable",               # type-checking like, had false-positives
--- a/src/fava/application.py
+++ b/src/fava/application.py
@@ -95,7 +95,9 @@ CLIENT_SIDE_REPORTS = [
 ]
 
 
-if not mimetypes.types_map.get(".js", "").endswith("/javascript"):
+if not mimetypes.types_map.get(".js", "").endswith(
+    "/javascript"
+):  # pragma: no cover
     # This is sometimes broken on windows, see
     # https://github.com/beancount/fava/issues/1446
     logging.error("Invalid mimetype set for '.js', overriding")
@@ -246,12 +248,10 @@ def _setup_filters(
             g.ledger = fava_app.config["LEDGERS"][g.beancount_file_slug]
 
     @fava_app.errorhandler(FavaAPIError)
-    def fava_api_exception(error: FavaAPIError) -> str:
+    def fava_api_exception(error: FavaAPIError) -> str:  # pragma: no cover
         """Handle API errors."""
         return render_template(
-            "_layout.html",
-            page_title="Error",
-            content=error.message,
+            "_layout.html", page_title="Error", content=error.message
         )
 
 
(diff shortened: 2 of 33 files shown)


frontend/src/entry-forms/Transaction.svelte
```diff
@@ -1,16 +1,11 @@
-<script lang="ts" context="module">
```
yagebu, line 1: Why was this code removed? We still want to have this logic to be able to input tags and links

frontend/src/api/validators.ts
```diff
@@ -192,6 +193,7 @@ export const getAPIValidators = {
   move: string,
   payee_accounts: array(string),
   payee_transaction: Transaction.validator,
+  narration_transaction: Transaction.validator,
```
yagebu, line 196: This also needs to be added to the GetAPIParams interface, that's currently causing the svelte-check error

yagebu (changes requested):
Thanks for the PR, this looks quite good but I've left two comments :)

frontend/src/entry-forms/Transaction.svelte
dtrai2: I removed it because the linter told me that it wasn't used anymore. I will try to double check it again in the next few days. Same as for your other comment. Thanks for the hint with the svelte-check error. I probably wouldn't have found that one.

frontend/src/api/validators.ts
dtrai2: Thanks I have updated it.

frontend/src/entry-forms/Transaction.svelte
dtrai2: So I did had a look at it again and I inserted it again. Tags and links will now be parsed properly before adding the new entry to the beancount file.

frontend/src/entry-forms/Transaction.svelte
```diff
@@ -41,34 +42,14 @@
...
-  /// Extract tags and links that can be provided in the narration <input>.
-  function onNarrationChange({
-    currentTarget,
-  }: {
-    currentTarget: HTMLInputElement;
-  }) {
-    const { value } = currentTarget;
-    entry.tags = [...value.matchAll(TAGS_RE)].map((a) => a[1] ?? "");
-    entry.links = [...value.matchAll(LINKS_RE)].map((a) => a[1] ?? "");
-    entry.narration = value
```
yagebu, line 53: this needs to be re-added. we need to remove the extracted tags and links here

frontend/src/entry-forms/Transaction.svelte
```diff
@@ -41,34 +42,14 @@
...
+  /// Extract tags and links that can be provided in the narration.
+  function onNarrationChange() {
+    entry.tags = [...entry.narration.matchAll(TAGS_RE)].map((a) => a[1] ?? "");
+    entry.links = [...entry.narration.matchAll(LINKS_RE)].map(
+      (a) => a[1] ?? "",
+    );
   }
 
-  /// Output tags and links in the narration <input>
-  function combineNarrationTagsLinks(e: Transaction): string {
```
yagebu, line 60: this needs to be re-added. we need to add the tags and links here

yagebu (changes requested):

dtrai2:
Ah I don't know what happened last time, but some of my wanted changes weren't committed/pushed. I did reconstruct them now as best as I could. So the tags and links should actually be parsed again. The only weird thing I have noticed is that the existing tags and links are not shown inside the drop-down selection menu. Struggling in reconstructing it I noticed also that I start feel that I am a bit in over my head with my limited frontend skills :smile: So maybe you have a hint or solution for that.

frontend/src/entry-forms/Transaction.svelte
dtrai2: I added it again, sorry that you had to check it when the actual changes I wanted to share weren't present.

frontend/src/entry-forms/Transaction.svelte
dtrai2: It's also present again, but it is only executed once a selection is made.

PR #1817, opened 2024-05-28
add css improvements for journal on mobile devices

dtrai2:
Hi I just tried to improve the mobile view of journals. You can consider this PR as a proposal to improve the look and feel, so any feedback and remarks are welcome!

My issue with the current look and feel is that it is difficult to read what is actually going on. For example like this:

The narrations and the payees are often shortened with `...`.

My suggestion would be to give the rows a second line. The first line has the date, flag, indicators and the prices, while in the second line it has the narrations or the accounts:

As seen in the screenshot: budgets are not perfect yet, but it's a start. 
What do you think about that?

### DIFF
--- a/frontend/css/journal-table.css
+++ b/frontend/css/journal-table.css
@@ -231,3 +231,99 @@
   text-transform: lowercase;
   border-radius: 20px;
 }
+
+@media (width <= 767px) {
+  article {
+    padding: 1rem;
+  }
+
+  .flex-table.journal p {
+    flex-wrap: wrap;
+  }
+
+  .journal .head .flag {
+    flex-grow: 1;
+    text-align: left;
+  }
+
+  .journal .datecell {
+    order: 1;
+    width: 6rem;
+    text-align: left;
+  }
+
+  .journal .flag {
+    order: 2;
+  }
+
+  .journal .description {
+    flex: inherit;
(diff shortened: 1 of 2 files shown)


pwalkr:
The expanded postings look much better without the left-side margin from the date column. What about margins in the account view? The icons to expand postings shouldn't need more than 1-2 dots to get the idea, it seems like the change and balance fields could stay on the same line:
account

dtrai2:
Yes thanks for the hints. It does still need some work, and I'll continue with it. I just wanted to wait a bit for feedback before I start working in a direction that is not favored.

yagebu:
> My suggestion would be to give the rows a second line. The first line has the date, flag, indicators and the prices, while in the second line it has the narrations or the accounts:

Yeah, I guess something like expanding it to two lines is needed on very small screens. The full color version is a bit too colorful in my opinion - could you try reducing that to maybe just having one of the cells colored (or some colored border maybe?)?

I definitely like the changed postings display, removing that left indent is a very good idea.

dtrai2:
thanks for the feedback! I had a try at the postings and the alignments now, and I also removed again the colors highlighting the full row. let me know what you think and if there is something else I should have a look at.

yagebu:
Merged with some modifications

Issue #1297, opened 2021-06-19
Filter by amount?


andreasgerstmayr:
Is there a way to filter transactions by amount? For example, I'd like to view all transactions of the expenses account over 1000 EUR (i.e. where the "changes" column is over 1000 EUR).

I think it's tricky for the other views (e.g. for the journal view, which posting amount should be selected for filtering?), but for the account detail view it would be straight forward to filter (filter by the "changes" column).

I know I can use custom BQL for it, but then I don't have the nice editing functionality and the other UI elements/options of the accounts page.

nzbart:
This would be great as a Fava feature.

For those of you (like I did) who have trouble figuring out how to write BQL to search by amount, the magic column you need is `Number`:

```
select * where Number = 25000
```

xinluh:
Implemented in #1307

LucaBernstein:
Btw, for anyone curious how to perform at least exact matches, just found out that this works in fava:

`any(units: "2.00 USD")`

You can leave out the currency, but then the regex could include unwanted results (e.g. extreme example: `any(units: "2")` would match transactions with units of `2.00 EUR` as well as `0.02 EUR` (i.e. filtering for `.*2.*`))

PR #1819, opened 2024-06-01
support numerical comparison operators for amounts in filter

andreasgerstmayr:
Support numerical comparison operators for amounts in filter
Resolves: #1297

This adds the following features to the filter DSL:
```
>= 100
any(units > 100)
```

I think this addresses your comments from a previous PR https://github.com/beancount/fava/pull/1307#issuecomment-986008505. In a previous iteration I used `any(units > "100 USD")`, however I think it is more straightforward and user friendly to leave the currency out when using comparison operators.

### DIFF
--- a/src/fava/core/filters.py
+++ b/src/fava/core/filters.py
@@ -13,7 +13,6 @@ from typing import TYPE_CHECKING
 
 import ply.yacc  # type: ignore[import-untyped]
 from beancount.core import account
-from beancount.core.amount import Amount
 from beancount.ops.summarize import clamp_opt  # type: ignore[import-untyped]
 
 from fava.beans.account import get_entry_accounts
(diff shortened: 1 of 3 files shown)


yagebu:
Thanks for the PR - I'm still not sure whether I like the bare `= 20`, but meh, let's try it. I think comparing to absolute amounts makes more sense than completely ignoring negative numbers, I've changed it to that.

andreasgerstmayr:
> Thanks for the PR - I'm still not sure whether I like the bare `= 20`, but meh, let's try it. I think comparing to absolute amounts makes more sense than completely ignoring negative numbers, I've changed it to that.

Thanks for merging!
I agree, `= 20` looks a bit weird and probably isn't very useful, but supporting `> 500` in the filter is a nice feature in my opinion (e.g. checking what the one expensive purchase/outlier was in a given month/journal page).

Issue #882, opened 2019-03-08
default-file could also set the insertion file


ljani:
I expected [`default-file`](https://github.com/beancount/fava/blob/77b5436c842f9be27a5dc368a4b058b17d13919e/fava/help/options.md#default-file) option to also set the append file for the "add transaction" dialog of the UI, instead of only changing the behavior which file is opened when editor is opened. Currently the new entries are inserted in the main file regardless of `default-file` setting.

The buttons for the editor and add transaction are next to each others in the UI, thus I found this very confusing.

According the docs, the [`insert-entry`](https://github.com/beancount/fava/blob/77b5436c842f9be27a5dc368a4b058b17d13919e/fava/help/options.md#insert-entry) option seems to be the way to control this, but I think that `insert-entry` should fallback to looking for `default-file` before "the entry will be inserted at the end of the main file."

Any thoughts?

yagebu:
Sure. I don't care much for `default-file` are you willing to provide a PR?

ljani:
I could take a look when I have some spare time, but do not expect it soon.

yagebu:
Closing this stale issue, a PR is still welcome :)

abourget:
Would love this.. how do we change the behavior of the New entry UI?!

Issue #1698, opened 2023-10-19
Fava option 'default file' has no effect (Fava 选项“default-file” 没有效果)


czcn:
Fava 选项“default-file” 没有效果，新交易记录，总是写到Beancount的"filename"中
Fava option 'default file' has no effect. New transaction records are always written to the 'filename' of Beancount option.

czcn:
It works. Used for Fava editor to open by default. Not the file where the transaction record was added.

yagebu:
Yes, exactly, this option only sets to the default file to open for the editor.

You can use the `insert-entry` option to influence where the entries get inserted.

PR #1838, opened 2024-07-18
default-file improvements

Minion3665:
1 of 2:
It's a common point of confusion that default-file does not set the
default file for insertions. This patch brings the behavior into line
with people's expectations.

BREAKING-CHANGE: If you are using default-file and expecting entries to be inserted to your main file, this will no longer happen
Fixes: #882
Refs: #1698, #822

---

2 of 2:
I autogenerate some fava config, and one option I'd like to specify is
the default file.

Unfortunately, this previously had to be specified in the file which
would become the default file, leading to a messy and fragile setup in
which my (readonly) configuration file would become the default if the
magical "default-file" line was ever removed from the user-editable
config.

This feature allows existing semantics around default-file to continue,
while also providing the ability to set it in some other file if the
existing options are unsuitable.

### DIFF
--- a/src/fava/core/fava_options.py
+++ b/src/fava/core/fava_options.py
@@ -119,12 +119,17 @@ def parse_option_custom_entry(  # noqa: PLR0912
     if key not in All_OPTS:
         raise UnknownOptionError(key)
 
+    value = entry.values[1].value if len(entry.values) > 1 else None
+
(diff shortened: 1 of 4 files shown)


yagebu (approved):
Thanks for the PR, I agree that using the default_file as a default for inserts seems more sensible :)

Issue #1849, opened 2024-08-28
The use of `watchfiles` breaks changes detection on NFS


aronsky:
I'm not sure yet what's the culprit of the problem, but in my case, `fava` exposes a ledger mounted into a pod via NFS, and another pod runs periodically and imports new transactions into the ledger. Unfortunately, more often than not, these changes aren't picked up by Fava. This is not a matter of multiple copies of NFS mounts - if I go to edit the files in Fava, the changes are there. They simply aren't reflected in the journal and reports - so it must be Fava not reloading the ledger's files.

I think it's a problem with `watchfiles`, since up until recently everything worked perfectly. I can't be 100% sure though, as I also made a switch from Docker to k8s, so that might have something to do with the issue (though in both cases, I mount the same NFS share with the same options, so I doubt that's the issue).

aronsky:
I verified that the old watcher works as expected, and created a new flag (`--poll-watcher`) that can be used to tell `fava` to use the old watcher instead of the new one (PR #1850)

aronsky:
More info: seems like it is, indeed, an issue with the `Notify` library that's used by `watchfiles`: see here: https://github.com/notify-rs/notify/issues/384#issuecomment-1030934574

The reason it used to work for my Docker setup, but now doesn't work as expected on my k8s setup is that with Docker, there was a single NFS mount shared by all the containers. Whereas in my k8s setup, the containers are separated into different pods, and the NFS volume is potentially mounted several times by separate nodes.

yagebu:
Thanks for adding the option to switch to the polling watcher to workaround this issue, closing this one.

PR #1850, opened 2024-08-29
Add `--poll-watcher` flag.

aronsky:
This enables the use of the old, poll-based watcher. For cases where the`watchfiles` library (more likely, the underlying `Notify` library) isn't working correctly. See relevant issue #1849.

This small change adds support for using the old watcher based on polling the filesystem, instead of the new, `watchfiles`-based watcher. This is relevant for people (like me), for whom the new watcher doesn't work as expected. The support is via a simple command-line flag (`--poll-watcher`), which, when present, is propagated to the constructor of `FavaLedger`, where the decision to use either the old `Watcher` or the new `WatchfilesWatcher` is made based on that flag's value.

### DIFF
--- a/src/fava/cli.py
+++ b/src/fava/cli.py
@@ -97,6 +97,10 @@ def _add_env_filenames(filenames: tuple[str, ...]) -> set[str]:
     type=click.Path(),
     help="Output directory for profiling data.",
 )
+@click.option(
+    "--poll-watcher",
+    is_flag=True, help="Use old polling-based watcher."
+)
 @click.version_option(version=__version__, prog_name="fava")
 def main(
     *,
@@ -109,6 +113,7 @@ def main(
     debug: bool = False,
     profile: bool = False,
     profile_dir: str | None = None,
+    poll_watcher: bool = False,
 ) -> None:  # pragma: no cover
     """Start Fava for FILENAMES on http://<host>:<port>.
 
@@ -129,6 +134,7 @@ def main(
         all_filenames,
         incognito=incognito,
         read_only=read_only,
+        poll_watcher=poll_watcher,
     )
 
     if prefix:
--- a/src/fava/cli.py
+++ b/src/fava/cli.py
@@ -101,7 +101,7 @@ def _add_env_filenames(filenames: tuple[str, ...]) -> set[str]:
     "--poll-watcher", is_flag=True, help="Use old polling-based watcher."
 )
 @click.version_option(version=__version__, prog_name="fava")
-def main( # noqa: PLR0913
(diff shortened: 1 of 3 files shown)


aronsky:
More info: seems like it is, indeed, an issue with the `Notify` library that's used by `watchfiles`: see here: https://github.com/notify-rs/notify/issues/384#issuecomment-1030934574

yagebu:
I guess this is not something we could automatically determine and switch to the polling watcher if we know watchfiles won't work on that filesystem?

There's some linter complaints the pre-commit reports, could you address those?

aronsky:
Sure. The linter complains are mainly due to my attempt to make my changes as limited as possible, without delving into redesigning certain interfaces. Namely:

* The `main` function now receives an additional argument regarding the watcher type, and this puts it over the edge of acceptable argument count (10 instead of 9)
* To make sure that the `watchfiles`-based watcher is used by default, the boolean argument that enables the poll watcher is set to `False` by default, which seems to upset the linter

Is it OK to simply silence those warnings at those spots, or do you want me to refactor the code to avoid them?

And regarding automatic detection - unfortunately, I don't see such an option. The thing is, even if we detect the filesystem type on which the data resides (and let's assume we have an exhaustive list of filesystems that are problematic, even though that's not the case right now) - it doesn't necessarily mean there will be a problem. My case is a perfect example. I'm using NFS - but as long as it's a single NFS mount by the host, and multiple containers access it simultaneously, `watchfiles` works fine. The problem only appears when the containers are on different hosts, each of them mounting the filesystem separately. I don't see any way to detect this logically, let alone programmatically, without explicit user configuration (which this flag essentially is).

src/fava/core/__init__.py
```diff
@@ -222,7 +223,7 @@ class FavaLedger:
     #: Dict of list of all (unfiltered) entries by type.
     all_entries_by_type: EntriesByType
 
-    def __init__(self, path: str) -> None:
+    def __init__(self, path: str, poll_watcher: bool = False) -> None:
```
yagebu, line 226: If you make this a keyword-only argument, the linter won't complain (since it forces the caller to not accidentally pass a boolean arg without telling which one it is)

src/fava/cli.py
```diff
@@ -109,6 +112,7 @@ def main(
     debug: bool = False,
     profile: bool = False,
     profile_dir: str | None = None,
+    poll_watcher: bool = False,
```
yagebu, line 115: For this one, silencing the lint seems fine

yagebu (changes requested):

yagebu (approved):
Thanks :)

Issue #1384, opened 2022-02-01
Feature suggestion - add unrealized profit and loss column in Holdings


jenicek:
One more minor suggestion :) It may be worth considering whether not to add a column for unrealized profit and loss (either as an absolute value or as a percentage of cost) to those Holdings tables that are aggregated (by Account, etc.).

In `src/fava/templates/holdings.html`, either one of these two would be added to the select clause

- `sum(number(value(position))) - sum(number(cost(position))) as unrealized_profit`
- `(abs(sum(number(value(position)))) - abs(sum(number(cost(position))))) / sum(number(cost(position))) * 100 as unrealized_profit_pct`

yagebu:
Sounds good. I have a slight preference for the percentage value as it's the one I usually look at (in my brokerage online account) and it seems like the absolute value could be more easily computed in the head (or in a separate BQL query) if one needs it

jenicek:
I agree - to me, percentages are easier to interpret than absolute values. Should I create a patch for that change?

yagebu:
> Should I create a patch for that change?

Yes, that would be great :)

jenicek:
No worries :) I've tested it for commodities with a negative cost and commodities with no cost (in my original query posted here, a division-by-zero error would have been raised). I've added it in all the tables except the first "holdings_all" table, as that one seems already a bit polluted, but it should be easy to update that one as well.

```
diff --git a/src/fava/templates/holdings.html b/src/fava/templates/holdings.html
index f30e935d..3434544c 100644
--- a/src/fava/templates/holdings.html
+++ b/src/fava/templates/holdings.html
@@ -14,7 +14,8 @@
 'account': 'SELECT account,
     units(sum(position)) as units,
     cost(sum(position)) as book_value,
-    value(sum(position)) as market_value
+    value(sum(position)) as market_value,
+    safediv((abs(sum(number(value(position)))) - abs(sum(number(cost(position))))), sum(number(cost(position)))) * 100 as unrealized_profit_pct
   WHERE account_sortkey(account) ~ "^[01]"
   GROUP BY account, cost_currency, account_sortkey(account), currency
   ORDER BY account_sortkey(account), currency',
@@ -22,14 +23,16 @@
     units(sum(position)) as units,
     SAFEDIV(number(only(first(cost_currency), cost(sum(position)))), number(only(first(currency), units(sum(position))))) as average_cost,
     cost(sum(position)) as book_value,
-    value(sum(position)) as market_value
[...]

PR #1863, opened 2024-09-18
Show unrealized profit percentage in Holdings tab

1ma:
The commit that introduced the unrealized profit/loss percentage column in the holdings tabs left out the primary Holdings tab for some unstated reason.

Issue: https://github.com/beancount/fava/issues/1384
Commit: https://github.com/beancount/fava/commit/971ae3757daeae8b8b407c8a28072adcc40055cd

I believe that seeing the unrealized profit or loss of each individual lot can be very useful:

holdings


federicociro:
Nice

yagebu (approved):
Thanks for the PR :) Yeah, this was not left out for any specific reason and I agree that it's helpful to add it as well.

PR #1864, opened 2024-09-20
frontend: AutocompleteInput: escape * in fuzzyfinding (#1846)

### LINKED ISSUE #1846: Search completion issue when searching payee: field with unescaped `*` character
Hi,

I've met an issue with the search completion where it suggests an unescaped string (I'm searching for "leroy"):

image

leading to the account being empty

image

If I escape the string correctly, I do get the correct payee subacount:

image

Tested on fava-1.29.dev3+g365cd5dc

alexandre-janniaux:
When inserting a suggestion, the value can have invalid characters, and in particular `*` which can easily be present on the payee: field or narration field.

Fixes #1846

### DIFF
--- a/frontend/src/AutocompleteInput.svelte
+++ b/frontend/src/AutocompleteInput.svelte
@@ -72,6 +72,7 @@
       input != null && valueSelector != null
         ? valueSelector(suggestion, input)
         : suggestion;
+    value = value.replace("*", "\\*");
     dispatch("select", input);
     hidden = true;
   }


yagebu (changes requested):
Hi, thanks for the PR :) 

Since this is not the only character that might cause issues when searching as a regular expression, would be good to solve that while we're at it. Could you create a function that covers all of them (and some tests)? (see e.g. this stackoverflow thread for possible solutions https://stackoverflow.com/questions/3561493/is-there-a-regexp-escape-function-in-javascript)

yagebu:
Closing, see #1907

Issue #1874, opened 2024-10-18
Light/Dark mode switch


Evernight:
It's great that the dark-mode has been added, thanks! I personally usually use it with the most websites when available. However, I'm not sure I prefer it with Fava and it also doesn't work well with fava-dashboards yet (which I use extensively). While the latter would probably be a nice fix for fava-dashboards repository, in the meantime it would be great just to have an ability to switch between the light and dark mode (which is typical for the websites that support it).

I checked whether it's possible to configure per-website config in the browser and it seems it's not common or easily available (without plugins). So the switch in Fava either via button or via config option would probably be the best solution in the meantime. Probably the button would be even better as it also would work with publicly hosted https://fava.pythonanywhere.com/.

PR #1883, opened 2024-10-27
feat: add light/dark mode switch (#1874)

nerd4me:
add light/dark mode switch (#1874)

image
image

### DIFF
--- a/src/fava/core/ingest.py
+++ b/src/fava/core/ingest.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import datetime
 import os
 import sys
 import traceback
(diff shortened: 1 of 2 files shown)


yagebu:
Hi, thanks for creating this PR. There's some problems I see with this current approach:

- I think the screen space in the header is too valuable for this sort of infrequently used toggle. I think putting it on the options page would be better.
- The default shouldn't be dark but whatever the system default is. So there would need to be three toggle options. AFAICT there's no good way to implement this without having to duplicate all the styles... Generally, I guess it's annoying that this is not simply a browser feature...

One other point: Did you create those images yourself?

PR #1887, opened 2024-11-01
fix color of cursor in editor when using dark mode

dtrai2:
In dark mode the cursor was black which made it difficult to see, this PR sets the color such that it changes according to the theme.


yagebu (approved):

yagebu:
Thanks :)

PR #1888, opened 2024-11-01
add auto-completion for narration

### LINKED ISSUE #1535: Auto-completion in the "Narration" field in the Add popup

dtrai2:
**third time's the charm**

I accidentally deleted my changes again and closed the pull request #1816, luckily we have something like `git reflog` with which I could reconstruct it. 

Anyways this is still my attempt to add the auto-completion feature for the narration text field. It is still not perfect, but I didn't want to let the changes go to waste. Maybe it's good enough like this or someone might help out to finish it (see my last comment on #1816). 

closes #1535, #1145


tests/__snapshots__/test_application-test_client_side_reports
```diff
@@ -941,6 +941,43 @@
   "links": [
     "test-link"
   ],
+  "narrations": [
```
yagebu, line 944: This part is problematic. In my case, all these narrations would e.g. double the size of the balance sheet report. I think we should switch to lazily loading them only when the dialog is used.

dtrai2: okay yeah, I can understand the problem. Can you give me a hint for a possible implementation though? Or do you have something similar inside the repo where I could get some inspiration? 

I just looked at it and with my current understanding I would have to remove the narrations from the `get_ledger_data()` method inside `internal_api.py` and instead create another endpoint in `json_api.py` which is then somehow called once the dialog is opened? Is this remotely a good way to solve this or is there maybe an easier way?

yagebu: Sorry for the late reply but I wanted to first address the migration to Svelte 5 which has some impact on how such an implementation might look like. With that out of the way, this could now be implemented.

> create another endpoint in json_api.py which is then somehow called once the dialog is opened? Is this remotely a good way to solve this or is there maybe an easier way?

That sounds about right, yes. A `get_narrations` endpoint that would be fetched from the `Transaction.svelte` component.

dtrai2: Ah no worries, I know sometimes these things take time! 👍 I am not sure yet when I actually find time for it again myself, but maybe I can have a look at it in June. I'll let you know when I gave it a try or when I need some help.

dtrai2: Hi @yagebu, I took my shoot at it and I hope it works now like you'd expect. Please have a look at it again. Any feedback is much appreciated! As discussed the narrations aren't part of the `get_ledger_data` anymore and are instead only queried once the `Transaction.svelte` component is loaded.

yagebu (approved):
Thanks, LGTM :)

I've only modified it slightly so that attributes.narrations is a `@property` and used a snapshot test to have some fewer "hand-written" tests.

PR #1923, opened 2024-12-22
Re-execute query on filter param change

nbdd0121:
I discovered that since Fava 1.29, changing time or account filters on top-right corner doesn't update the current query anymore, despite the params in URL being changed.

This PR fixes the issue by re-submitting the query when filter_params are changed.

### DIFF
--- a/frontend/src/reports/query/Query.svelte
+++ b/frontend/src/reports/query/Query.svelte
@@ -29,6 +29,12 @@
     }),
   );
 
+  onMount(() =>
+    filter_params.subscribe(() => {
+      rerun_all();
+    }),
+  );
+
   /** Submit the current query and load the result for it. */
   function submit() {
     const query = query_string;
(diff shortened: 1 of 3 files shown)


frontend/src/reports/query/Query.svelte
```diff
@@ -18,16 +18,20 @@
   const results: Record<string, Result<QueryResult, string>> = {};
   const is_open: Record<string, boolean> = {};
 
-  onMount(() =>
+  onMount(() => {
```
yagebu, line 21: The return value is important so that unsubscribe is called on unmount. You can simply add a second onMount block

frontend/src/reports/query/Query.svelte
```diff
@@ -18,16 +18,20 @@
...
       if (search_query_string !== query_string) {
         query_string = search_query_string;
         submit();
       }
-    }),
-  );
+    });
+
+    filter_params.subscribe(() => {
+      submit();
```
yagebu, line 32: we also need to handle the results cache and the other results that are displayed

yagebu (changes requested):

yagebu:
Thanks for the PR, left some comments

frontend/src/reports/query/Query.svelte
nbdd0121: Sorry, I don't follow. Can you elaborate?

yagebu: When you run multiple queries in sequence, all those results are kept around (in the object `results`). Those results also need to be updated / invalidated when the filter changes

nbdd0121: Before Fava 1.29, when I update the filter only the latest result is updated. Can you confirm the intended behaviour is to re-run all history queries?

yagebu: Re-running all of them is probably a bit excessive. I'll change this to re-run all for which the result is currently opened.

yagebu:
Merged in [4da3342b318b94510647ee32c99f7d16c5c3a7ae](https://github.com/beancount/fava/commit/4da3342b318b94510647ee32c99f7d16c5c3a7ae)

The URL updating I fixed in #1928

PR #1937, opened 2025-01-03
Fix beancount3 duplicate logic

tarioch:
With beancount3 the __duplicate__ field can now contain the original entry instead of "True", mapping this to the old "True" value for the rest of the import logic to work

### DIFF
--- a/src/fava/core/ingest.py
+++ b/src/fava/core/ingest.py
@@ -307,6 +307,11 @@ def extract_from_file(
         entries.sort(key=_incomplete_sortkey)
     if isinstance(importer, Importer):
         importer.deduplicate(entries, existing=existing_entries)
+        for entry in entries:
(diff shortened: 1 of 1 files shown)


tarioch:
@yagebu  lint doesn't pass but I'm not getting what I should adjust. I'm more or less following the code here https://github.com/beancount/beangulp/blob/f8d2bc5da40e74461772dcb8e8394fa3c1b2a65d/beangulp/extract.py#L221

and here https://github.com/beancount/beangulp/blob/f8d2bc5da40e74461772dcb8e8394fa3c1b2a65d/beangulp/extract.py#L198

yagebu:
In the types in this repo, the metadata is typed as immutable, so you're not allowed to modify that (without creating a new entry and replacing the existing one). I don't think there's any reason to mutate it, see #1938

PR #1944, opened 2025-01-07
Bulk importer for Fava

umangv:
This is my first pass at making a bulk importer for Fava.

I find that one of the slowest parts of importing multiple transactions is that there's a lot of visual jumping. You have to look at the payee/narration, then jump to the postings, see if those make sense, sometimes you may need other hints (like the date or the amounts) before you're sure. Once you click next, you start over.

By grouping transactions by their "target account" (i.e., the "interesting" posting, where the "uninteresting" posting is the one that involves the account you are importing into), it's much easier to visually scan through the importer, catch problems, and make changes.

Known issues that I will address before converting from "draft" to "ready":
 * [ ] lint/test issues

Known issues that I hope to address (either in this PR or in a follow up):
 * [ ] Check boxes don't honor shift-clicks to select a range
 * [ ] Doesn't use Svelte v5 runes. I tried migrating the current code to Svelte runes but it got nasty because the imports grouped by target account objects had references to items in the original `entries` array and the runes magic broke those references. This might be fixable with `$state.raw()` but I haven't experimented. 

Nice to haves that I also hope to work on in the future:
 * [ ] A drag and drop interface for moving transactions (either individually, or all selected transactions) to different sections.

### DIFF
--- a/frontend/css/bulk-importer.css
+++ b/frontend/css/bulk-importer.css
@@ -1,0 +1,41 @@
+.bulk-importer span.select,
+.bulk-importer span.edit {
+    width: 2rem;
+}
+
+.bulk-importer span.flag {
+    width: 3rem;
+}
+
+.bulk-importer span.description {
+    flex: 1;
+}
+
+.bulk-importer span.datecell {
+    width: 6rem;
+}
+
+.bulk-importer span.datecell,
+.bulk-importer span.flag {
+    text-align: center;
+    background-color: var(--entry-background);
+}
+
+.bulk-importer span.edit button {
+    padding: 2px 4px;
+    margin: 0;
+    font-weight: bold;
+    background: var(--background);
+}
+
+.bulk-importer span.edit button:hover {
+    background: var(--background-darker);
+}
+
+.bulk-importer .postings {
+    font-size: 0.9em;
+    background-color: var(--journal-postings);
+}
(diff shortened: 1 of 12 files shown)


yagebu:
> I've been using my branch for the past couple of months and my personal experience has been that bulk import as a concept makes importing entries much smoother but I don't love the fact that I made this is a modal (mostly because vertical scrolling has been a little janky inside a modal) and maybe it deserves a whole page. With my limited experience with Svelte, I'll probably have to learn more about how state is maintained between pages to get that working.

Making it full-page (or just a full-screen modal?) doesn't need to have any impact on the state, it can still be rendered from the same `Import.svelte` component.

> Great. I had not used Svelte before hacking at this, so it was especially confusing to be working with two paradigms while getting up to speed on the basics. Given that it's been a few months since I touched this, I'll probably have to do that again but I'll try to get everything working with runes mode.

Yeah, that was a confusing mix of paradigms - let me know if you run into any problems with getting it working with runes, I also still run into weird things there, e.g. https://github.com/beancount/fava/commit/d16a84e7fa7e22f0aec9ad785e1c97fdcb181063

umangv:
I haven't had the time to devote to getting this to work: life has been busy. If anyone is interesting in taking a stab at this, please ahead! Otherwise, I will try to get around to this eventually.

PR #1945, opened 2025-01-07
filters: Match metadata keys for postings as well

iostapyshyn:
This PR extends the metadata filtering to match against postings' metadata as well.

For instance, currently, when filtering for `key:""` or `key:"foo"`, the following transaction would not match as the filter only considers the metadata of the entry:

```beancount
2025-01-01 * "Transaction"
  Assets:A -10 CUR
    key: "foo"
  Expenses:B
```

### DIFF
--- a/src/fava/core/filters.py
+++ b/src/fava/core/filters.py
@@ -361,9 +361,15 @@ class FilterSyntaxParser:
         def _key(entry: Directive) -> bool:
             if hasattr(entry, key):
                 return match(getattr(entry, key) or "")
-            if entry.meta is not None and key in entry.meta:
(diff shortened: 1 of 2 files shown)


iostapyshyn:
Hi @yagebu,

I didn't realize this was possible with `any`/`all` and the PR does seem superfluous now. Thanks for the clarification!

yagebu:
> I didn't realize this was possible with any/all and the PR does seem superfluous now. Thanks for the clarification!

If you can think of an improvement to the help pages which would have made that clearer, a suggestion or PR would be welcome as well :)

Issue #1939, opened 2025-01-05
Add a 'fortnightly' frequency for budgets


JP-Ellis:
## Summary

I would love to see a `fortnightly` time interval for use within budgets.

## Motivation

In Australia, it is quite common to have income and debits charged at a fortnightly (every 2 weeks) frequency. It would be really convenient if the budgets in Fava could support a fortnightly interval:

```beancount
2024-07-01 custom "budget" Expenses:Health:Gym    "fortnightly" 29.00 AUD
```

## Alternatives

Currently, I just do the conversion explicitly into a weekly amount, but it isn't quite as nice:

```beancount
2024-07-01 custom "budget" Expenses:Health:Gym    "weekly" (29.00 / 2) AUD
```

## Implementation

I would be happy to help contribute a PR if that is a change that would be welcome.

PR #1955, opened 2025-02-04
feat: add fortnightly interval

JP-Ellis:
In some countries (such as Australia), it is very common for wages to be paid on a fortnightly basis, that is, every 2 weeks. As a result, a number of other expenses are also paid on a fortnightly basis.

This commit extends the `Interval` enum by adding a `FORTNIGHT` option. I have added a test for that, and have updated the translations where possible (not that other than for English and French, the other translations may be suboptimal).

~As part of this commit, I have also taken the opportunity to fix a confusion between ISO weeks and calendar weeks as the two do not always match, and `2025-W01` is the ISO format indicating the week starting on December 30 2024. Let me know if you want me to split out this particular change into a separate PR.~ EDIT: I have moved this into #1956.

Resolves: #1939

### DIFF
--- a/frontend/src/lib/interval.ts
+++ b/frontend/src/lib/interval.ts
@@ -1,6 +1,12 @@
 import { _ } from "../i18n";
 
-export type Interval = "year" | "quarter" | "month" | "week" | "day";
+export type Interval =
+  | "year"
+  | "quarter"
+  | "month"
+  | "fortnight"
+  | "week"
+  | "day";
 
 export const DEFAULT_INTERVAL: Interval = "month";
 
@@ -8,6 +14,7 @@ export const INTERVALS: Interval[] = [
   "year",
   "quarter",
   "month",
+  "fortnight",
   "week",
   "day",
 ];
@@ -22,6 +29,7 @@ export function intervalLabel(s: Interval): string {
     year: _("Yearly"),
     quarter: _("Quarterly"),
     month: _("Monthly"),
+    fortnight: _("Fortnightly"),
     week: _("Weekly"),
     day: _("Daily"),
   }[s];
--- a/src/fava/util/date.py
+++ b/src/fava/util/date.py
@@ -35,6 +35,9 @@ DAY_RE = re.compile(r"^(\d{4})-(\d{2})-(\d{2})$")
 # this matches a week like 2016-W02 for the second week of 2016
 WEEK_RE = re.compile(r"^(\d{4})-w(\d{2})$")
 
+# this matches a fortnight like 2025-W01/02
+FORTNIGHT_RE = re.compile(r"^(\d{4})-w(\d{2})/(\d{2})$")
+
 # this matches a quarter like 2016-Q1 for the first quarter of 2016
 QUARTER_RE = re.compile(r"^(\d{4})-q([1234])$")
 
@@ -46,7 +49,7 @@ FY_QUARTER_RE = re.compile(r"^fy(\d{4})-q([1234])$")
 
 VARIABLE_RE = re.compile(
     r"\(?(fiscal_year|year|fiscal_quarter|quarter"
-    r"|month|week|day)(?:([-+])(\d+))?\)?",
+    r"|month|fortnight|week|day)(?:([-+])(\d+))?\)?",
 )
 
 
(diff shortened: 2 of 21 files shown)


src/fava/util/date.py
```diff
@@ -93,6 +96,31 @@ class Interval(Enum):
     YEAR = "year"
     QUARTER = "quarter"
     MONTH = "month"
+    FORTNIGHT = "fortnight"
+    # Note: While a fortnight is a well-defined time interval, there are no
```
yagebu, line 100: Can you move this (maybe in a slightly shortened form) to the help pages this seems relevant for users to understand how this works

src/fava/util/date.py
```diff
@@ -163,9 +204,23 @@ def get_prev_interval(
...
         return datetime.date(date.year, date.month, 1)
+    if interval is Interval.FORTNIGHT:
+        year, week, _ = date.isocalendar()
+        w1 = week - 1 if week % 2 == 0 else week
+        if w1 > 2:
+            return datetime.date.fromisocalendar(year, w1, 1)
+        # If we are in the first fortnight of the year, finding the last
+        # fortnight of the previous year is a bit tricky due to being either
+        # starting on week 51 or 53 (if the year has 53 weeks).
+        try:
```
yagebu, line 215: This file had complete code coverage in tests (except for the annotated lines), with your changes some cases don't seem to be fully covered - e.g. this try-except-block. Can you add some more tests for those? (`make test` will show the lines or you can take a look at the HTML coverage in `htmlcov`)

frontend/src/lib/interval.ts
```diff
@@ -1,28 +1,36 @@
 import { _ } from "../i18n";
 
-export type Interval = "year" | "quarter" | "month" | "week" | "day";
+export type Interval =
+	| "year"
```
yagebu, line 4: seems these lines are indented with tabs and not spaces - I noticed `ts` was missing from the pre-commit config for prettier, I just fixed that on main - can you rebase and re-run pre-commit to switch to spaces here as well?

yagebu (changes requested):
Thanks for the PR, looks good - I just left some minor comments.

frontend/src/lib/interval.ts
JP-Ellis: I noticed the formatting changes and just assumed it was covered pre-commit (maybe incrementally changes the style or something)

More than happy to rebase :+1:

JP-Ellis:
I have updated the tests to ensure 100% test coverage. I did have to add to lines with `# pragma nocover` because they should be unreachable.

For the coverage, would you be open to a PR adding integration with Codecov? This will help see coverage of PRs automatically, and it is free for OSS projects. You can see an example of how Codecov integrates within GitHub at:

https://github.com/pact-foundation/pact-python/pull/935

And you can follow links therein to see the more detailed coverage reports.

JP-Ellis:
Any further changes needed before this can be merged?

yagebu (changes requested):
Took a while, but since the "overlapping intervals"  seemed problematic, I needed to take some time to review this.

I think we need to settle and clarify how to correctly handle 53-week years. Overlapping intervals are not an option  so we either need accept that some "fortnights" are only one week long or we'll need global start date (week 1 of year 1 for example) to have them non-overlapping over the years. IMHO the former is probably preferable, as it's otherwise impossible to tell when a fortnight starts.
The `get_next_interval` logic already works that way, however some other places still assume all fortnights are 14 days long or consist of two weeks when printing them, so those places need some adjusting. Other than that, this looks good, could you rebase it?

yagebu:
> For the coverage, would you be open to a PR adding integration with Codecov? This will help see coverage of PRs automatically, and it is free for OSS projects.

Thanks, but no thanks for now. In another project (lektor/lektor) there's been a lot of churn to get codecov to work again after some breaking changes to their actions and it's still not working. I don't see that codecov adds enough value to warrant these efforts to keep it working (in particular if it's a cloud-only proprietary tool, I'd rather focus on open source tools that can be run locally). In #2001 I've added a tox build in CI to test for full coverage.

JP-Ellis:
I'm on holidays at the moment, so I'll get back to this PR when I'm back 🙂

JP-Ellis:
Sorry, this has dropped from my list of priorities. Closing.

PR #1956, opened 2025-02-05
fix: iso week usage

### LINKED ISSUE #1805: Unconsistency in week Date filter

JP-Ellis:
Fava appears to use ISO 8601 weeks (e.g. `2025-W01`), but unfortunately does so incorrectly due to a mismatch between ISO 8601 and Gregorian calendar week definitions.

ISO 8601 defines precisely how to handle weeks, namely, that the first week of a year is the week that contains 4 January. This definition results in some occasional mismatches with the Gregorian calendar such as:

- The first week of 2025 starts on 30 December 2024
- 3 January 2021 belongs to Week 53 of 2020

Fortunately, the fix is straightforward as `strftime` provide[^posix]:

- `%G` for the week-based year
- `%V` for the week number of the year (with no 'week 0' as `%W` can return).

This change will ensure that there is no ambiguity on weeks that straddle the new year, though at the cost of introducing occasional off-by-one errors compared to previously generated data.

[^posix]: See [POSIX.1-2024](https://pubs.opengroup.org/onlinepubs/9799919799/functions/strftime.html)

Fixes #1805


yagebu (changes requested):
Thanks for the PR - this came up in #1805 and I agree with switching to ISO week numbers. The current behaviour is as documented  in the help page on the filters and that will have to be updated as well, could you take a look at that? Other than that, the PR looks good

JP-Ellis:
Amazing! Sure happy to look at the docs too. I'll probably do this tomorrow or Monday.

JP-Ellis:
Docs have been updated 👍

yagebu (approved):
LGTM, thanks :)

PR #1957, opened 2025-02-05
chore: add ruff to dev deps

JP-Ellis:
The `pyproject.toml` file contains a `[tool.ruff]` section, but Ruff is missing from the dev dependencies. I assume this was an oversight and this commits adds it.

I have also taken the opportunity to sort the dev (and `old_deps_pins`) dependencies in alphabetical order.


yagebu:
The pre-commit hooks include ruff so pre-commit sets up a separate virtual environment up for it. I don't think it needs to be added to the `.venv` as well (increased the size by 24MB on my machine).

JP-Ellis:
The reason for including it in the virtual environment is to ensure that the IDE (VSCode in my case) finds the correct version during development, instead of waiting until I create a commit to realise I made some minor linting mistakes. It also ensures that the IDE uses the same version as the pre-commit, as opposed to a perhaps incompatible version it might find elsewhere. Overall, it just makes for a much nicer developer experience.

As to your comment about the size of the virtual environment; this is purely a dev dependency, and therefore should not impact downstream users of Fava on its own (unless they install `favs[dev]`...)

JP-Ellis:
I noticed there's now a merge conflict. Would you like me to update the PR to resolve that? Or is the PR a no-go?

As mentioned above, this would _only_ be a dev dependency and therefore should not impact downstream users of the package. It will only ensure that IDEs can use the same version of ruff as in the pre-commit hook, thereby avoiding linting issues during development (as opposed to waiting for the pre-commit hook to fail).

yagebu:
I'm aware this would only be a dev dependency and so it wouldn't affect end users. Still don't think it's worth it. I've never run into any issues due to a version mismatch the system installed and the pre-commit version. In #1966 I've disabled the preview mode, so there shouldn't be any differences between minor versions. I'm closing this PR, if you run into actual issues due to this, feel free to re-open

JP-Ellis:
Thanks for getting back to me! I appreciate your explanation, but I'm having a little trouble understanding it. It seems a bit inconsistent to me because the dev dependencies already include other CLI tools like `mypy`, `tox`, `twine`, and `pre-commit`. Based on what you mentioned, it sounds like those should also be system-installed instead of being part of the dev dependencies. Ultimately, I'll leave it up to you how you'd like to manage that 🙂 

In the meantime, I'll just manually `uv pip install ruff` into the venv whenever I need to make changes to Fava. Thanks again for your input!

JP-Ellis:
I use (unfortunately) different versions of Ruff in different projects. As a result, I don't use the bundled version of Ruff, nor do I typically use the globally installed version. Instead, I have my IDE configured to search the Python venv, and same goes for my CLI interactions. 

This means that the IDE will occasionally have slight mismatches in what it detects compared to what is used by precommit. It's not a huge annoyance, but still slightly so nevertheless. It was rather trivial to add the dependency, hence why I created the PR; but as a new contributor, I do not want to overstep.

In particular, I absolutely do get the concern around having out of sync Ruff versions! Personally, I have my updates managed by [Renovate](https://github.com/apps/renovate), and have it [configured](https://github.com/pact-foundation/pact-python/blob/297b7881419ab31565c3696e22e7658e9e86ba3f/.github/renovate.json#L15) to update them at the same time. And honestly, I find Renovate to be excellent at managing dependencies, both in its configurability and it's exhaustiveness (much more coverage than Dependabot, and includes pre-commit).

If that is something you would consider, I'd be happy to help onboard Renovate[^1], or feel free to copy the config from that link.

[^1]: I am on no way affiliated to Renovate, just an honest recommendation based on my experience professionally and in my own side projects.

yagebu:
Thanks for elaborating! In #2015 I've now added ruff to the dev environment (and removed other ones like sphinx that are only relevant for specific make targets, with the switch to uv that's now doable).

> In particular, I absolutely do get the concern around having out of sync Ruff versions! Personally, I have my updates managed by [Renovate](https://github.com/apps/renovate), and have it [configured](https://github.com/pact-foundation/pact-python/blob/297b7881419ab31565c3696e22e7658e9e86ba3f/.github/renovate.json#L15) to update them at the same time. And honestly, I find Renovate to be excellent at managing dependencies, both in its configurability and it's exhaustiveness (much more coverage than Dependabot, and includes pre-commit).

Thinking about this some more, the separate versions are not actually that much of a concern on the locking side (and now with uv, it's also pretty likely that the venv will be synced quickly): I do the version bumps via the `update` make target which would bump ruff in both places at the same time.

Haven't used renovate so far but I prefer easily locally runnable tools that don't want to tie into some proprietary tool. So far that's working quite well with the described make target. This also allows me to batch these updates when I have time to look into any breakage they might produce and fix that.

JP-Ellis:
Amazing! The uv lock file looks like it would do everything needed! 

PS: Renovate is open source, you can self host/run it if need be 🙂

PR #1959, opened 2025-02-08
lint: drop pylint

yagebu:
pylint doesn't add much since we have mypy and ruff but is quite slow
and this shaves about 10s off of `make lint` on my machine


yagebu:
also seems to saves round-about 16s on the lint-python CI job (42s on main -> 26s on this branch)

PR #1965, opened 2025-02-18
click indicators on non-transactions to toggle metadata

marvhen:
Hello fava team,

I tried my hand at hacking a bit and got some desired functionality working pretty good I'd say.  Let me know what you think.  I ran `make test` and `make lint` and I think both passed but `make lint` ended with his and I'm not sure what the `Error 1` refers to:

```
...

stylelint................................................................Passed
eslint...................................................................Passed
make: *** [Makefile:45: lint] Error 1
(fava) PS C:\git\fava>
```

Anyhow, this change simply makes it possible to expand the non-transaction rows in the journal that have meta-data, as I understand that is supported for almost all directives I believe.  I'm handling it for balance, note, open, and custom entries.  For custom, it only worked in the case of budget...I think because metadata elements aren't created otherwise.  Also, I think I had trouble getting it to work for pad.  Anyway, I'm not necessarily looking for this to be pulled into main as much as I'm trying to get a little experience with the code base and ask for some feedback.

Thanks!

### DIFF
--- a/frontend/css/journal-table.css
+++ b/frontend/css/journal-table.css
@@ -77,7 +77,12 @@
 .journal.show-metadata .metadata,
 .journal.show-postings .postings,
 .transaction.show-postings .postings,
-.transaction.show-postings .metadata {
+.transaction.show-postings .metadata,
+.balance.show-metadata .metadata,
+.note.show-metadata .metadata,
+.open.show-metadata .metadata,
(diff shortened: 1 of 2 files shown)


marvhen:
Also, one of the things I am wondering is if this type of functionality could possibly be done with a plugin instead?  Any tips on that would be much appreciated!

PR #1976, opened 2025-03-20
Fix #1975

### LINKED ISSUE #1975: `/statement` does not normalize file paths
The `fava.plugins.link_documents` plugin properly normalizes paths so that documents can be referenced with relative paths: https://github.com/beancount/fava/blob/bb37f983bcda81acaa04c607a342fad7a85eedd1/src/fava/plugins/link_documents.py#L76

However, the `/statement` endpoint doesn't perform the same normalization: https://github.com/beancount/fava/blob/bb37f983bcda81acaa04c607a342fad7a85eedd1/src/fava/core/__init__.py#L612-L618

This results in navigating to a linked document from a journal entry failing.

sushain97:
Fixes https://github.com/beancount/fava/issues/1975.

Alternatively, we could check each document against both the normalized and non-normalized path. I'm happy to make that change if it would be preferable.

### DIFF
--- a/src/fava/core/__init__.py
+++ b/src/fava/core/__init__.py
@@ -7,6 +7,7 @@ from datetime import timedelta
 from functools import cached_property
 from functools import lru_cache
 from itertools import takewhile
+from os.path import normpath
 from pathlib import Path
 from typing import TYPE_CHECKING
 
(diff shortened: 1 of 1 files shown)


yagebu (approved):
Thanks :)

PR #1983, opened 2025-04-11
Adding account and importer to match with beangulp

tarioch:
Beangulp also adds the account and the importer to the result (which can e.g. used by hooks). See e.g. https://github.com/beancount/beangulp/blob/master/examples/import.py

### DIFF
--- a/src/fava/core/ingest.py
+++ b/src/fava/core/ingest.py
@@ -7,6 +7,7 @@ import os
 import sys
 import traceback
 from dataclasses import dataclass
+from inspect import get_annotations
 from inspect import signature
 from os import altsep
 from os import sep
@@ -46,7 +47,10 @@ if TYPE_CHECKING:  # pragma: no cover
     from fava.beans.ingest import FileMemo
     from fava.core import FavaLedger
 
-    HookOutput = list[tuple[str, list[Directive]]]
+    HookOutput = (
+        list[tuple[str, list[Directive], str, BeanImporterProtocol | Importer]]
+        | list[tuple[str, list[Directive]]]
+    )
     Hooks = Sequence[Callable[[HookOutput, Sequence[Directive]], HookOutput]]
 
     P = ParamSpec("P")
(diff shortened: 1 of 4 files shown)


tarioch:
@yagebu if you can please review this (found this while working on switching smart-importer to beangulp hooks).

src/fava/core/ingest.py
```diff
@@ -443,15 +446,23 @@ def extract(self, filename: str, importer_name: str) -> list[Directive]:
...
+                importer,
                 path,
                 existing_entries=self.ledger.all_entries,
             )
         except Exception as exc:
             raise ImporterExtractError from exc
 
-        new_entries_list = [(filename, new_entries)]
+        new_entries_list: HookOutput = []
+        if isinstance(importer, Importer):
```
yagebu, line 459: I don't think switching on the Importer type here makes too much sense - the hooks apply to all importers, so we need to treat them consistently.

We could either try to have some try-catch-logic here to try both (or maybe there's some way to inspect the function signature to determine it?) alternatively, we could have this be opt-in by providing the hooks under some different variable name in the import config.

yagebu: I guess my preference would be to check on the typing annotations to allow hooks to opt-in to being called with a list of 4-tuples instead of 2-tuples

tarioch: thanks for your feedback, can you give me a hint how to do that?

yagebu: You can use get_annotations (https://docs.python.org/3.13/library/inspect.html#inspect.get_annotations) to get the resolved annotations (added in 3.10, I guess we'd just keep the old logic on 3.9)

tarioch: Adjusted

tarioch: @yagebu  any chance you can have a look whether the adjustments look good?

src/fava/core/ingest.py
```diff
@@ -443,16 +447,37 @@ def extract(self, filename: str, importer_name: str) -> list[Directive]:
...
             )
         except Exception as exc:
             raise ImporterExtractError from exc
 
-        new_entries_list = [(filename, new_entries)]
         for hook_fn in self.hooks:
+            new_entries_list: HookOutput = []
+            try:
+                annotations = get_annotations(hook_fn)
+            except NameError:
```
yagebu, line 463: why would this throw a NameError?

src/fava/core/ingest.py
```diff
@@ -443,16 +447,37 @@ def extract(self, filename: str, importer_name: str) -> list[Directive]:
...
                 existing_entries=self.ledger.all_entries,
             )
         except Exception as exc:
             raise ImporterExtractError from exc
 
-        new_entries_list = [(filename, new_entries)]
         for hook_fn in self.hooks:
+            new_entries_list: HookOutput = []
+            try:
+                annotations = get_annotations(hook_fn)
```
yagebu, line 462: until today, we actually only had a minimum Python version 3.9, but since beanquery was broken there, I've bumped it to 3.10 so we can use this here as well

yagebu (changes requested):
Can you add a test covering this logic?

src/fava/core/ingest.py
```diff
@@ -443,16 +447,37 @@ def extract(self, filename: str, importer_name: str) -> list[Directive]:
...
+                        filename,
+                        new_entries,
+                        importer_info.account,
+                        importer.importer,
+                    )
+                ]
+            else:
+                new_entries_list = [(filename, new_entries)]
+
             new_entries_list = hook_fn(
```
yagebu, line 481: this logic currently does not work correctly for multiple hooks - for each hook_fn, we'll have to have this check

src/fava/core/ingest.py
tarioch: Dropped

src/fava/core/ingest.py
tarioch: Removed

src/fava/core/ingest.py
tarioch: You're absolutely right, I adjusted the logic further

yagebu (approved):

PR #1988, opened 2025-04-19
Add missing deps in the docker image; tell pip to ignore the fact we are running as root

patrixl:
I was trying to build the Docker container image and ran into various issues with building, namely:

- missing dependency: bison
- missing dependency: flex
- pip install erroring out because it's running as root

Fixed the first two by adding the deps to the `apk install` command, and fixed the latter with the `--root-user-action ignore` flag to the `pip install`.

### DIFF
--- a/contrib/docker/Dockerfile
+++ b/contrib/docker/Dockerfile
@@ -1,7 +1,7 @@
 FROM python:alpine as builder
 
-RUN apk add --update libxml2-dev libxslt-dev gcc musl-dev g++
-RUN pip install --prefix="/install" fava
+RUN apk add --update libxml2-dev libxslt-dev gcc musl-dev g++ bison flex
+RUN pip install --root-user-action ignore --prefix="/install" fava
 
 FROM python:alpine
 


contrib/docker/Dockerfile
```diff
@@ -1,7 +1,7 @@
 FROM python:alpine as builder
 
-RUN apk add --update libxml2-dev libxslt-dev gcc musl-dev g++
-RUN pip install --prefix="/install" fava
+RUN apk add --update libxml2-dev libxslt-dev gcc musl-dev g++ bison flex
```
yagebu, line 3: Thanks for the PR. I'd be for taking this chance and switching away from the alpine image. For this simple docker image I don't think it's a great base image (see e.g. reasons listed in https://pythonspeed.com/articles/alpine-docker-python/). If we instead use `python:slim`, we can drop this line altogether as then pre-compiled wheels will available for all deps.

patrixl: Although the article does say alpine has improved since it was first written, there's still advantages to python:slim. Makes sense to me, I'll look into it!

yagebu: Made the changes in 37fa6a09da279a1e6e6c2c35899fa5163ef4dfab, so closing this one

yagebu:
Made the changes in https://github.com/beancount/fava/commit/37fa6a09da279a1e6e6c2c35899fa5163ef4dfab, so closing this one

PR #1989, opened 2025-04-19
Add Japanese translation

jokester:
Add `ja` translation


jokester:
I'm guessing eventually the translation should be synced to and managed in poeditor?

yagebu:
Thanks for creating the translations - I've imported them on POeditor and will sync them in #2003

PR #1994, opened 2025-04-29
Feature: filter presets

Evernight:
### Motivation
I really like the filters functionality and use it extensively. However, in most cases you have to type some text in the filter fields and that text is often repeated. There are values that I use more frequently than others (e.g. month-1 or month-11 to month-1 for time, or some specific tag or posting filter for the advanced filter). It would be nice to be able to use them without having to type anything.

### Current workarounds
Creating and storing links in a sidebar, bookmarks, dashboards or elsewhere. But it doesn't allow to mix and combine filters together.
Also having to keep complicated Advanced filter strings somewhere in the notes.

### Proposed implementation
Define filter presets in Fava config, similar to sidebar links.
If any presets are defined, display them in a strip that extends the header.
There are two types of presets: for time and advanced filters.
Time is pretty self-explanatory.
Advanced filters "toggle" on click so it's possible to combine them together separated by a space for AND logic. Uses simple substring match to avoid exposing lexer from Python (that overcomplicates things).
Actual value of the filter and its label are separate, especially relevant for Advanced filter that may look rather complicated (that's one of the reasons for this feature)

### Screenshot
Screenshot 2025-04-29 at 14 16 01

### Notes
Currently doesn't include documentation or tests, will add if the feature is generally approved.

### DIFF
--- a/frontend/src/api/validators.ts
+++ b/frontend/src/api/validators.ts
@@ -107,6 +107,7 @@ export const ledgerDataValidator = object({
   payees: array(string),
   precisions: record(number),
   sidebar_links: array(tuple(string, string)),
+  filter_presets: array(tuple(string, string, string)),
   tags: array(string),
   upcoming_events_count: number,
   user_queries: array(object({ name: string, query_string: string })),
--- a/frontend/src/sidebar/AdvancedFilterPresetLink.svelte
+++ b/frontend/src/sidebar/AdvancedFilterPresetLink.svelte
@@ -1,0 +1,51 @@
+<script lang="ts">
+  import { filter_params, fql_filter } from "../stores/filters";
+
+  interface Props {
+    value: string;
+    label: string;
+  }
+
+  let { value, label }: Props = $props();
+
+  // Track if the filter is active (present in the current filter as a substring)
+  let isActive = $derived($filter_params.filter.includes(value));
+
+  function setAdvancedFilter() {
+    const currentFilter = $filter_params.filter;
+
+    // Check if the value is already a substring of the current filter
+    if (currentFilter.includes(value)) {
+      // If it's already in the filter, remove it
+      const newFilter = currentFilter.replace(value, "").trim();
+
+      fql_filter.set(newFilter);
+    } else {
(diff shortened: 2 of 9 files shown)


yagebu:
Hi, thanks for the PR!

I see the ability to toggle filters more quickly as a useful feature but I don't think we should add a new configuration mechanism for this. To enable this via the existing sidebar links configuration (or bookmarks or whatever), I'd instead suggest creating a new URL route similar to jump that applies these filter transformations.

Or even better, additionally, to enable (some parts of this feature) without having any specific configuration, how about storing recent filters and adding them to the dropdowns? For the time filter examples in your screenshot I'd expect that to work quite well.

Evernight:
Hi, thanks for the reply!

> To enable this via the existing sidebar links configuration (or bookmarks or whatever), I'd instead suggest creating a new URL route similar to jump that applies these filter transformations.

What you imagine these urls to look like? Do I understand correctly that it would be something that just toggles specific value (e.g. tag) in the filter, if clicked, but via routing?
Also if you have code pointers for reference at hand, that'd be useful. I'll also need to learn a bit more about Svelte routing.

> storing recent filters 

Do you mean browser's local storage?

I've actually considered adding these to the dropdowns initially but:
1) It adds one additional click to access functinality
2) Needs some design so that current suggestions, recents, etc don't get all mixed up and there's some visual separation
3) While for the time filter it's possible, for the given examples for the advanced filter it's more complicated. That's where aliases are more useful as sometimes filter values (especially when filtering by postings) may get long and complicated. For example, the current filter value field gets not so helpful relatively quickly as it's short, so I need to copy the text elsewhere to read.

As for sidebar links as an implementation, one reason I didn't go with this is that I already use those extensively. I have about dozen of presets for various pages, some with default time or advanced filter applied. Having more just makes the list less comprehensible (or asks for an ability to group links in the sidebar, for example). Also these "filter presets" are meant to be applied on top of already chosen path and to be combined together in various configurations.

In any case, happy to keep in running in the fork on a side as it works well for me. But was wondering if there's anything code- or idea- wise that would be useful for the mainstream.

PR #2006, opened 2025-05-08
docs: edited the badge and its link

operagxoksana:


yagebu:
Hi, can you comment on why you made this change? The current badge seems to show the license just fine.

operagxoksana:
Since the link to the license icon was incorrect and the icon itself was not up to standard, the word license was doubled. If it is significant I can revert it back to its original state

Issue #1964, opened 2025-02-10
Add icicle charts


gorlins:
[Icicle](https://observablehq.com/@d3/icicle/2) charts serve the same purpose as Sunburst charts, but are much more readable, and do not suffer many of the [issues](https://www.data-to-viz.com/caveat/pie.html) with [pie charts](https://scc.ms.unimelb.edu.au/resources/data-visualisation-and-exploration/no_pie-charts).  

They are natively supported by the [d3 library](https://d3js.org/d3-hierarchy/partition) currently used by Fava's sunburst and treemaps, so I presume this would be an easy addition, but I am not a svelte developer.

I would suggest adding this as an option (or replacement!) wherever a sunburst chart is offered, eg in any of the breakdowns on the balance sheet or income statement.

I have included the expenses sunburst from the demo site and an icicle chart for the same (mostly?) data (using the [plotly](https://plotly.com/python/icicle-charts/) library) to illustrate the difference.  Icicle charts are closer to treemaps in terms of readability, but preserve the hierarchical structure from the sunburst which is lost in the treemap.

![Image](https://github.com/user-attachments/assets/b0aa9928-1a26-4da0-ac91-08f65938eb59)

![Image](https://github.com/user-attachments/assets/1ae4e74f-fc45-4509-92b9-2d01abb33dd0)

yagebu:
Implemented in #2019 :)

PR #2019, opened 2025-05-21
Feature/icycle charts

gorlins:
I took a stab at implementing #1964.  Most of the code is copied either from the sunburst or treemap, and merged together to get a working icycle chart.  It appears to work with Income and Expenses on the income statement, as well as Assets, Liabilities, and Equity on the balance statement.

Example image from the example.beancount:

image

I am not a svelte/typescript dev so please modify or suggest anything...

(aside - I kept the colormap used by the sunburst chart, but I feel like there are much better color map options for this...)

### DIFF
--- a/frontend/src/charts/Icycle.svelte
+++ b/frontend/src/charts/Icycle.svelte
@@ -1,0 +1,114 @@
+<script lang="ts">
+  import type { HierarchyRectangularNode } from "d3-hierarchy";
+  import { partition } from "d3-hierarchy";
+  import { scaleLinear, scaleSqrt } from "d3-scale";
+  import { arc } from "d3-shape";
+  import { untrack } from "svelte";
+
+  import { formatPercentage } from "../format";
+  import { urlForAccount } from "../helpers";
+  import { ctx } from "../stores/format";
+  import { sunburstScale } from "./helpers";
+  import type {
+    AccountHierarchyDatum,
+    AccountHierarchyNode,
+  } from "./hierarchy";
+
+  interface Props {
+    data: AccountHierarchyNode;
+    currency: string;
+    width: number;
+    height: number;
+  }
+
+  let { data, currency, width, height }: Props = $props();
+
+  let radius = $derived(Math.min(width, height) / 2);
+
+  let root = $derived(partition<AccountHierarchyDatum>()(data));
+  let nodes = $derived(
+    root.descendants().filter((d) => !d.data.dummy && d.depth > 0),
+  );
+
+  let current: AccountHierarchyNode | null = $state(null);
+
+  $effect.pre(() => {
+    // if-expression to run on each change of chart
+    // eslint-disable-next-line @typescript-eslint/no-unused-expressions
+    data;
+    untrack(() => {
+      current = null;
+    });
+  });
+
+  function balanceText(d: AccountHierarchyNode): string {
+    const val = d.value ?? 0;
+    const total = root.value ?? 0;
+    return total
+      ? `${$ctx.amount(val, currency)} (${formatPercentage(val / total)})`
+      : $ctx.amount(val, currency);
+  }
+
+  const x = scaleLinear([0, 2 * Math.PI]);
+  let y = $derived(scaleSqrt([0, radius]));
+  let arcShape = $derived(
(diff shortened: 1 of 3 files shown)


frontend/src/charts/Icycle.svelte
```diff
@@ -0,0 +1,134 @@
...
+
+<g
+  {width}
+  {height}
+  onmouseleave={() => {
+    current = null;
+  }}
+  role="img"
+>
+  <g
```
yagebu, line 77: I think this part looks weird and should be removed IMHO, we also don't have it in the treemap

frontend/src/charts/Icycle.svelte
```diff
@@ -0,0 +1,134 @@
...
+  }
+
+  let { data, currency, width, height }: Props = $props();
+
+  let root = $derived(partition<AccountHierarchyDatum>()(data));
+  let nodes = $derived(
+    root.descendants().filter((d) => !d.data.dummy && d.depth > 0),
+  );
+
+  let current: AccountHierarchyNode | null = $state(null);
```
yagebu, line 30: For the sunburst, I think this tracking and highlighting of the current account subtree makes it easier to spot as they might not line up.

In the icycle chart here, I don't like it, it seems to more of a flickering and since they're all lined up from left to right it's trivial to see the parent accounts. So maybe we can drop this whole "current" logic and the mouseovers?

yagebu (commented):
Thanks for the PR :)

Left some comments, looks mostly good! It should be "icicle", not "icycle", right?

frontend/src/charts/Icycle.svelte
yagebu: no, the `<g>` element and its contents starting on this line

gorlins: Actually I'm not clear on what you are saying.  Are you just talking about code formatting?  Or are you suggesting removing the block for the root node with the total balance?  I think it's necessary to include it somewhere - it's not in treemap because the treemap doesn't show any of the ancestors, but it's in sunburst as the inner circle.  I think it would be cleaner to remove the block but then change the loop below `each nodes as d` to include the root node in the iteration, which i can try to figure out.  If you are suggesting removing the block entirely, I don't think I agree - I think the root node with top level account and balance should be shown.

gorlins:
> It should be "icicle", not "icycle", right?

Oops!

frontend/src/charts/Icycle.svelte
gorlins: How strong is your opinion?  I kind of liked it myself... i thought certainly it helps for people who haven't seen icicle before, and I thought it helped when there are a lot of small accounts in the hierarchy - but my opinion isn't very strong

gorlins:
Thoughts on interactivity? In this example https://observablehq.com/@d3/zoomable-icicle the zoomability is really nice for deeper hierarchies, but would need to do something like adding an explicit link either for the account breakout or for the zoom.  I could take a stab at implementing if this is desireable

frontend/src/charts/Icicle.svelte
```diff
@@ -0,0 +1,104 @@
...
+    currency: string;
+    width: number;
+    height: number;
+  }
+
+  let { data, currency, width, height }: Props = $props();
+
+  let root = $derived(partition<AccountHierarchyDatum>()(data));
+  let nodes = $derived(
+    root.descendants().filter((d) => !d.data.dummy && d.depth >= 0),
```
yagebu, line 27: Does the `&& d.depth >= 0` do anything?

frontend/src/charts/Icycle.svelte
yagebu: > How strong is your opinion?

Not sure yet :D Maybe we can increase the opacity to reduce the flickering?

yagebu:
> Thoughts on interactivity?

Don't think we need/want it. I think we had this zoomable behaviour for the treemap at some point but removed due to it clashing with clicks / scrolling.

frontend/src/charts/Icycle.svelte
gorlins: I tried with a subtle drop shadow and blur effect - i think it helps with readability, but it's easy to remove

yagebu:
Thanks @gorlins :)

Issue #1230, opened 2021-02-12
Fold single child with parent account in the account table


upsuper:
For certain account settings, it may be very common for an account to only have a single child account without anything on its own, like
```
+ Assets
  + Amazon
  | + Points
  + BCinemas
  | + Points
  + DigitalOcean
    + Credits
```
I think it would be more compact and look nicer if we fold them together, something like
```
+ Assets
  + Amazon:Points
  + BCinemas:Points
  + DigitalOcean:Credits
```

Similar to how GitHub folds directories with a single subdirectory (see `.github/workflows`, `src/fava`, and `stubs/beancount`):
![107746970-cf7d7a00-6d6a-11eb-863e-8fb20450527c](https://user-images.githubusercontent.com/333750/107747107-0f446180-6d6b-11eb-9df7-fe24b8fb4c1d.png)

WDYT?

yagebu:
In flat layouts like the one on GitHub, I find this sort of folding useful as it saves some clicking.

In tree layouts, I am not a great fan of this folding of directories, mainly because it distorts the visual depth of an account in the hierarchy. In Fava, the folding would also be complicated by the fact that intermediate accounts might also have transactions, so the folding would be somewhat inconsistent.

I think that this folding can already be achieved by just naming the accounts appropriately (`Assets:AmazonPoints` instead of `Assets:Amazon:Points` for example).

upsuper:
> In Fava, the folding would also be complicated by the fact that intermediate accounts might also have transactions, so the folding would be somewhat inconsistent.

When there are transactions on the intermediate accounts, or at least when there is non-zero balance in the intermediate accounts, they shouldn't be folded. I don't think it's very inconsistent, as far as you can see that the intermediate account has different balance than the leaf one.

> I think that this folding can already be achieved by just naming the accounts appropriately (`Assets:AmazonPoints` instead of `Assets:Amazon:Points` for example).

It's true, but there is value not to do that. For example, you may want the level of merchant to always exist for accounting consistency or future extension. The intermediate account may also contain closed leaf account that you no longer care about.

The problem I'd like to solve is that when you use this kind of convention for account name, the account table becomes very long and low in information density. I think it would be good to at least allow user to choose (as an option) whether to do so.

PR #2057, opened 2025-07-04
fold empty intermediate accounts

upsuper:
This implements my idea in #1230. While I know it was closed, but I do think it would be nice to have this feature.

This is how it looks like:
image

If you want, I can try adding an option for it, but personally I don't think it's too problematic to always enable.


upsuper:
@yagebu Any thought about this?

PR #2065, opened 2025-07-09
Allow account completion to be case insensitive

Jack-Works:
In the Fava editor, if an intended line starts with a lowercase letter, there will be no completion

```beancount
2020-01-01 *
  account-name
  ;          ^ no completion here
```

This PR fixes this and provides a better edit experience.


yagebu:
Thanks for the PR :) I guess this was initially to not complete accounts when one is typing metadata. But I guess it doesn't hurt a lot to have the account completions when typing metadata but it helps a lot in the more common case of typing accounts.

PR #2067, opened 2025-07-12
fix: prevent double negative sign display on balance sheet tree-table

y0ngb1n:
### What is the problem?

When `diff` is negative, the component displays a double negative sign like `--10823.44`. This happens because the numeric formatting already includes the negative sign, and the template adds an additional `-` prefix.

> prevent-double-negative-sign-display-bug

### What changes were made?

- Use `Math.abs(diff)` to show the absolute value without the sign.
- Add a manual `+` or `-` prefix based on the sign of `diff`.


yagebu:
Thanks :)

PR #2075, opened 2025-07-19
Exclude modal dialogs from browser history

andreasgerstmayr:
Currently, when I close a modal dialog (e.g. the beancount file context in the journal view when pressing on the transaction date) and then press the back button in my browser, the modal dialog opens again. If I open a few modal dialogs after another, all of them open again when I click the back button a few times.

With this change, `history.back()` is called upon closing the modal dialog, therefore removing the modal URL from the browser history. Now, pressing the back button brings me to the previous page, and not the previous modal dialog.

### DIFF
--- a/frontend/src/stores/url.ts
+++ b/frontend/src/stores/url.ts
@@ -43,7 +43,7 @@ export const syncedSearchParams: Readable<URLSearchParams> = derived(
 
 export function closeOverlay(): void {
   if (window.location.hash) {
-    window.history.back();
+    window.history.replaceState(null, "", "#");
   }
   urlHash.set("");
 }
--- a/frontend/src/router.ts
+++ b/frontend/src/router.ts
@@ -303,6 +303,7 @@ export class Router extends Events<"page-loaded"> {
         return;
       }
       if (link.getAttribute("href")?.charAt(0) === "#") {
+        window.history.replaceState(null, "", link.getAttribute("href"));
         return;
       }
       if (is_external_link(link)) {


frontend/src/stores/url.ts
```diff
@@ -43,7 +43,7 @@ export const syncedSearchParams: Readable<URLSearchParams> = derived(
 
 export function closeOverlay(): void {
   if (window.location.hash) {
-    window.history.pushState(null, "", "#");
+    window.history.back();
```
yagebu, line 46: I don't think this is quite right if e.g. the user navigated to the modal from a different page. The "close" button should first and foremost close the overlay and not just act like the back browser back button. I don't think there's an easy way to implement that, since we don't have access to the history...

andreasgerstmayr: You're right, if e.g. Google contains a link to `http://127.0.0.1:3000/beancount/journal/?time=2025#context-6db7912cb27c36ac197dd5fbf9a9bebc` (directly opening the modal), then closing the modal would bring the user back to Google. But in any other case, if the modal is opened from Fava, imho it should work as expected.

Alternatively, what we could do is to not add the modal link to the browser history at all.
E.g. here: https://github.com/beancount/fava/blob/bf18b83232078241296b166d35390035a08f29af/frontend/src/router.ts#L305-L306 we can use `window.history.replaceState()` (replace the current URL with the modal URL) and when closing the modal again in https://github.com/beancount/fava/blob/bf18b83232078241296b166d35390035a08f29af/frontend/src/stores/url.ts#L46 we can use `window.history.replaceState()` instead of `window.history.pushState()` (to replace the current modal URL with the non-modal URL).

PR #2078, opened 2025-07-24
tests/test_json_api: update snapshot for test_api_query_result

Prince213:
This fixes a failed test.

Related-to: https://github.com/NixOS/nixpkgs/pull/427977


Prince213:
Looks like this is a downstream issue caused by outdated version of beanquery in nixpkgs.

PR #2094, opened 2025-08-20
Add commodity name tooltips to more tables

sim642:
This relates to #975, but some tables are still missing the tooltips after the issue was closed.


yagebu (approved):
Thanks, I made some minor edits :)

PR #2097, opened 2025-09-02
Add ctrl-p and ctrl-n for up/down navigation in drop-downs

paulsc:
Hi! I added ctrl+p and ctrl+n shortcuts to move the selection up & down in dropdown menus, e.g. in the slice editor auto-complete, or in the account quick-open. These are common shortcuts that I use everywhere else. I find it quite useful since they let me keep my hands on the home row (and avoid the horrible macbook arrow keys!).

### DIFF
--- a/frontend/src/AutocompleteInput.svelte
+++ b/frontend/src/AutocompleteInput.svelte
@@ -138,10 +138,16 @@
       } else {
         value = "";
       }
-    } else if (event.key === "ArrowUp") {
+    } else if (
+      event.key === "ArrowUp" ||
+      (event.ctrlKey && event.key.toLowerCase() === "p")
+    ) {
       event.preventDefault();
       index = index === 0 ? filteredSuggestions.length - 1 : index - 1;
-    } else if (event.key === "ArrowDown") {
+    } else if (
+      event.key === "ArrowDown" ||
+      (event.ctrlKey && event.key.toLowerCase() === "n")
+    ) {
       event.preventDefault();
       index = index === filteredSuggestions.length - 1 ? 0 : index + 1;
     }
--- a/frontend/src/codemirror/setup.ts
+++ b/frontend/src/codemirror/setup.ts
@@ -3,6 +3,7 @@ import {
   closeBrackets,
   closeBracketsKeymap,
   completionKeymap,
+  moveCompletionSelection,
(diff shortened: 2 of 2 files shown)


upsuper (commented):
My 2c.

I tend to think that this kind of non-obvious shortcuts are better done in a way that doesn't affect other users by default. Each person has their preference on keyboard usage, and unintuitive shortcuts definitely don't help on that. As an example, I removed ctrl-c shortcut for toggling charts in https://github.com/beancount/fava/pull/1603 which was very annoying for me.

Some potential solutions including:
1. Don't do it at all here. It should be fairly easy to implement such shortcuts with a browser user script that captures the keyboard event, and maybe dispatches a synthetic event with arrow key if needed. It would affect just yourself, and doesn't introduce any complexity to the project.
2. Add an option that enables different key scheme. More shortcuts may be added under those key schemes to allow user opt-in something they want to use. One such scheme could be Vim scheme.
3. Add some hook point that allows a plugin to provide shortcuts. Beancount and Fava have been moderately extensible. It might be useful to make it more so with more hook points provided.

I think it makes more sense to start with the first. User scripts can also be shared outside this project, and if such script sees more widespread adoption, it may start making more sense to integrate with the project itself with other solutions above.

upsuper:
> I looked it up, I guess it originated in Emacs, but it's now pretty widespread. In macOS it works on pretty much all search drop-downs, text editors. It's also in Firefox/Chrome search bars, VS Code quick finder etc. I use it a lot with the Alfred launcher, or in terminal, built into zsh/bash, the python REPL, I'm sure others.

I'd be very cautious using macOS as guidance on shortcuts. Majority of major UI shortcuts on macOS uses Command modifier key, but on other systems this role is taken by Ctrl instead. This difference makes using Ctrl shortcuts much less problematic on macOS than on other systems.

Ctrl+P are widely used as shortcut for Print, and Ctrl+N is widely used as creating new window on Windows and Linux distros. On macOS they are instead Cmd-P and Cmd-N so you don't see this problem there. Similarly, I often use Control-A on macOS to move the cursor to the beginning of line, but you need to be aware this is for Select All on all other systems.

That says, it may make more sense if these shortcuts are restricted to macOS rather than applied universally given that's a convention of the system.

Issue #2142, opened 2025-11-20
Cannot build fava: TypeError [ERR_UNKNOWN_FILE_EXTENSION]: Unknown file extension ".ts" for /opt/fava/fava/frontend/build.ts


mlell:
I cannot build fava as stated in the [documentation](https://beancount.github.io/fava/development.html):

Tried with commit: [0237d7](https://github.com/beancount/fava/commit/0237d700eb0e4872fc5e90cef132789efee9d5f5)

* `make dev` does not work anymore
* `make` yields `TypeError: Unknown file extension ".ts"` since commit `0564fd` (`v1.30.7-8-g0564fdcd`)
* `make` yielded another error (`module 'node:module' does not provide an export named 'registerHooks'`) before that, since commit `084fdf` (`v1.30.3-28-g084fdf6d`)

```
git clone https://github.com/beancount/fava
cd fava
python -m venv .venv
source .venv/bin/activate
## PROBLEM 1: make dev does not work anymore
make dev
## make: Nothing to be done for 'dev'.
make
## [...]
## TypeError [ERR_UNKNOWN_FILE_EXTENSION]: Unknown file extension ".ts" for /tmp/fava/frontend/build.ts
## [...]
```
Full Output below (Output 1).

I have bisected the problem to the following commit: [`0564fd frontend: use explicit extensions; use only erasable syntax`](https://github.com/beancount/fava/commit/0564fdcd328cf36533f76639d0b872afac3c7366)

However, before that commit, there is another problem (`SyntaxError: The requested module 'node:module' does not provide an export named 'registerHooks'`) which I bisected to [`084fdf test-js: adjust to allow testing of Svelte components`](https://github.com/beancount/fava/commit/084fdf6d9d01e1bf915fd30a1ba0b316993c575c). See "Output 2"

## Output 1

```
18:41 $ make
cd frontend; npm install --no-progress
npm warn EBADENGINE Unsupported engine {
npm warn EBADENGINE   package: 'fava@1.0.0',
npm warn EBADENGINE   required: { node: '>=22.18' },
npm warn EBADENGINE   current: { node: 'v20.19.5', npm: '11.6.2' }
npm warn EBADENGINE }

added 341 packages, and audited 342 packages in 5s

83 packages are looking for funding
  run `npm fund` for details

found 0 vulnerabilities
touch -m frontend/node_modules
cd frontend; npm run build

> fava@1.0.0 build
> node build.ts

node:internal/modules/esm/get_format:189
[...]

PR #2144, opened 2025-11-21
Enforce minimum node version to prevent Error 'TypeError [ERR_UNKNOWN_FILE_EXTENSION]: Unknown file extension ".ts"'

mlell:
To prevent others from making the same rookie mistake like me in #2142, this simple patch makes the check for the minimal node version mandatory. 

There is a failing test, but given that I have edited no code and just added a 1-line file `frontend/.npmrc`, it seems unrelated:

```
test at test/store.test.ts:32:1
✖ localStorage-synced stores (4.460862ms)
  SecurityError: Cannot initialize local storage without a `--localstorage-file` path
      at Object.get [as localStorage] (node:internal/webstorage:28:17)
      at get localStorage (node:internal/util:660:20)
      at TestContext.<anonymous> (file:///opt/beancount/fava/frontend/test/store.test.ts:38:3)
      at Test.runInAsyncScope (node:async_hooks:213:14)
      at Test.run (node:internal/test_runner/test:1106:25)
      at async Test.processPendingSubtests (node:internal/test_runner/test:788:7)
make: *** [Makefile:58: test-js] Fehler 1
```

### DIFF
--- a/frontend/.npmrc
+++ b/frontend/.npmrc
@@ -1,0 +1,1 @@
+engine-strict=true


yagebu:
> There is a failing test, but given that I have edited no code and just added a 1-line file frontend/.npmrc, it seems unrelated:

Yeah, this seems to happen on the recent Node versions - our script to patch globals from jsdom does not overwrite existing globals, so the native NodeJs localStorage gets used there and doesn't work. Guess we should also add latest node to CI.

yagebu (approved):
Thanks :)

PR #2155, opened 2025-12-13
Extend import help

mlell:
I would like to suggest this expansion of the import help. It was too terse and technical to understand for me when I just discovered beancount. I figured that a GUI like fava might be a good way to learn my ways around beancount, but the import docs were too confusing for me so I ended up working around the whole import process by just generating the beancount ledgers by a self-written stand-alone script.

This is a pitty, given that fava has a nice GUI support and I find the Importer declarations of beangulp very elegant. So therefore I have extended the help page of fava in the hope that other newbie users will have an easier time to start with beancount/fava.

The import help text is less dense now and presupposes less knowledge about importing in beancount and the detailed differences of the different generations of import mechanisms (beancount.ingest vs beangulp). There are clearer step-by-step instructions and an example that should get a user floating for a basic import-CSV case. Also, the full signature of hook function is now documented, which wasn't done anywhere in the documentations of beancount, beangulp, or fava.

Additionally, this adds to the syntax overview the simple information that commoditys = currencies from the perspective of beancount. This would have helped me as a newcomer to find correct BQL functions for my needs.

There are two failing tests, but I think they are not related to this PR, as I have only edited markdown files:

```
FAILED tests/test_core_fava_options.py::test_fava_options_language - fava.core.fava_options.UnsupportedLanguageOptionError: Fava has no translations for: 'zh'.
FAILED tests/test_util.py::test_get_translations - assert None
```


yagebu (changes requested):
Thanks for the PR, I've left some comments.

mlell:
Unfortunately, the `pre-commit` hook fails because the CI/CD system might have an older version of `mdformat`. I get

```
mdformat: error: unrecognized arguments: --exclude
```

The hook runs on my machine. I need the exclude statment, because mdformat wants to destroy the jinja2 syntax `{{ ... }}` in `import.md` which is required to create the GitHub URL for the user example.

@yagebu, can you update `mdformat`, or is there another way to exclude this file from `mdformat` treatment?

wuqs-net:
@yagebu Could you please take a look at this at your convenience? I agree it's a good idea to expand the import help.

Issue #1947, opened 2025-01-14
Add ability to select treemap (Income/Expense/...) in trial balance using url query parameter


christianglodt:
I'd like to be able to link to a treemap of a certain Expenses account. Now I can create a url like this:

http://localhost:5000/beancount/trial_balance/?account=Expenses%3AFood

However, this defaults to the "Income" treemap, which is empty in this case.

Would it be possible to add a url query parameter to select which treemap to show? For example:

http://localhost:5000/beancount/trial_balance/?account=Expenses%3AFood&treemap=Expenses

This would solve my immediate problem. I am using fava 1.29.

yagebu:
I guess we could track the currently selected chart in the URL hash which would enable linking to it. Not a priority item for me, a PR would be welcome.

wincent:
This would be useful for sidebar links that use `jump`. I have a few links defined with this shape:

```
2018-01-01 custom "fava-sidebar-link" "-Taxes" "/jump?filter=all(-account:%22Taxes%22)"
```

Use-case for links like this:

1. Start by looking at a view like "Trial Balance", which starts off showing the "Assets" chart.
2. Select the "Expenses" chart.
3. Decide I want to see expenses excluding taxes, so hit the "-Taxes" sidebar link.
4. That reloads the page, _but_ loses the state that I was viewing an "Expenses" chart and goes back to the default chart ("Assets").
5. Click "Expenses" again.

PR #2191, opened 2026-01-19
Track currently-selected chart name and chart mode in URL parameters

theaprilhare:
Addresses #1947 by adding URL parameters for the chart name and chart mode. 

For example, the Balance Sheet with the Assets chart and the Icicle chart mode uses the balance_sheet/?chart_mode=icicle&chart=Assets parameters.

The chart mode is stored in a per-report localStorage (fava-last-active-chart-modes), so that switching to another page and back preserves the chart mode and associated parameters.

The priority order for determining the chart mode is:

1. URL parameter (chart_mode=...)
2. The aforementioned localStorage
3. Global localStorage store as a fallback

### DIFF
--- a/frontend/src/charts/BarChart.svelte
+++ b/frontend/src/charts/BarChart.svelte
@@ -6,6 +6,7 @@
   import { urlForAccount } from "../helpers.ts";
   import { barChartMode, chartToggledCurrencies } from "../stores/chart.ts";
   import { ctx, currentTimeFilterDateFormat, short } from "../stores/format.ts";
+  import { url_chart_mode } from "../stores/url.ts";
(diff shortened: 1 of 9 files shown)


yagebu:
Hi, thanks for the PR. Did you use AI tools to create this?

Overall, this feels like too big of an implementation, not keen to merge this as is with the multiple layers of configuration.

theaprilhare:
I did for the code to derive the effective mode. Would it be better to just add chart and chart_mode to synced_search_param_names in frontend/src/stores/url.ts so they work the same way as conversion and interval currently do? The downside is obviously that you can't remember a different chart type per report (e.g., icicle for Balance Sheet and stacked bar for Income Statement).

PR #2196, opened 2026-01-21
fix: update current router URL on external location changes

andreasgerstmayr:
When extension modules bring their own router (e.g. `react-router` or `@tanstack/router`), Fava's router is not aware of URL changes, and the `router.current` property will get stale.

This is a problem when the global filters are changed (i.e. time, account or filter), because then Fava will redirect to that stale URL.

This PR subscribes to all navigation changes, and updates the `router.current` param of the router.


yagebu:
I don't like this (overriding prototypes of builtins). What's the use case that you need a router for the extension? We could consider exposing enough of the router to the extension to then enable this.

PR #2209, opened 2026-01-25
fix: update current router URL on external location changes (reworked)

andreasgerstmayr:
Continuation of #2196, which does not break the back button functionality.


yagebu (approved):
This still feels like a band-aid fix and we should probably implement some way for extensions to opt-in into frontend rendering, but well, let's try this out as a patch for now.

Issue #2180, opened 2026-01-08
Metadata indicators hide payee and narration


sclu1034:
My importers heavily utilize metadata to add useful information to a transaction, like https://github.com/jbms/beancount-import/blob/6b5421356f53bc71a3fa09c9438ef92aff56bb10/beancount_import/source/amazon.py#L85.

I have metadata hidden by default in Fava (via the "Toggle metadata" button), but the indicator pills still show, frequently leading to cases like this:

<img width="1041" height="44" alt="Image" src="https://github.com/user-attachments/assets/a9e8db00-a467-4ece-9c17-bb35fdc095bf" />

I would consider payee and narration much more important fields than the indicators, but the latter is given priority in taking the available space.

yagebu:
Option to not show them implemented in #2214, so closing

PR #2214, opened 2026-02-08
feat: add option to control metadata indicator visibility

slimslickner:
Add a new Fava option 'show-metadata-indicators' (default: true) that allows users to disable the display of metadata indicator badges in the journal.

This addresses issue #2180 where metadata indicators were consuming valuable screen space and hiding important information like payee and narration text.

Users can now disable indicators by adding to their Beancount file:
  2024-01-01 custom "fava-option" "show-metadata-indicators" "false"

The option is backward-compatible and defaults to showing indicators, maintaining existing behavior for current users.

### DIFF
--- a/src/fava/core/fava_options.py
+++ b/src/fava/core/fava_options.py
@@ -108,6 +108,7 @@ class FavaOptions:
     show_accounts_with_zero_balance: bool = True
     show_accounts_with_zero_transactions: bool = True
     show_closed_accounts: bool = False
+    show_metadata_indicators: bool = True
     sidebar_show_queries: int = 5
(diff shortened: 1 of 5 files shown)


yagebu:
Thanks for the work on this, I'm still a bit unsure whether this deserves to be an option, but let's go with it for now :)

yagebu (changes requested):
This option needs to be documented, then this would be ready to be merged

slimslickner:
Oops good catch. I added to `src/fava/help/options.md`

yagebu (approved):

PR #2216, opened 2026-02-08
frontend: use darker text color on charts

### LINKED ISSUE #2215: Bad contrast on trial balance chart/text
In light mode, the contrast is ok but it feels like the text could be a bit darker.
Image

In dark mode, the text is really unreadable.
Image

The styling comes from
https://github.com/beancount/fava/blob/e1ea34518be1278160a98fafa01a36e74d122801/frontend/css/charts.css#L3
which is defined here
https://github.com/beancount/fava/blob/e1ea34518be1278160a98fafa01a36e74d122801/frontend/css/style.css#L38

jakobhellermann:
fixes #2215 

Dark:
image
image
Light:
image
image

### DIFF
--- a/frontend/css/charts.css
+++ b/frontend/css/charts.css
@@ -1,6 +1,6 @@
 svg text {
   font-family: var(--font-family);
-  fill: var(--text-color-lightest);
+  fill: var(--text-color-chart);
 }
 
 .tooltip {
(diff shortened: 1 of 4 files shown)


jakobhellermann:
Okay I changed the PR to only apply the new text-on-chart color for the icicle and treemap charts, and keep the rest as is.

yagebu (approved):
Thanks :)

yagebu:
Merged in d2fa9fa9d41edd854bd6e2733dc00eb1bdea6e3a

PR #2220, opened 2026-02-11
ci: add concurrency settings to GitHub Actions workflow

trim21:
there are some jobs never exit when esbuild failed to compiling svelet.

we should also fix it in esbuild side, but add this is also a good idea.

image

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,6 +12,10 @@ on:
 permissions:
   contents: read
 
+concurrency:
+  group: "${{ github.workflow }}-${{ github.ref }}"
+  cancel-in-progress: true
+
 jobs:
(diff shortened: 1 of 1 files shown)


trim21:
> I don't think I like this as this would lead to not getting two failures if two builds fail for different reasons, right?

Jobs won't get cancelled by same commit, it will only be cancelled by new commit on same branch

yagebu:
> Jobs won't get cancelled by other job of same commit, it will only be cancelled by new commit on same branch.

With most of the CI jobs taking under a minute, I don't really see the point then. Pushing more than a commit a minute should really only be necessary in the scenario where one has to debug some CI-only issue where it's then valuable to have all the jobs run.

PR #2225, opened 2026-02-14
Update build instructions

mocsa:
Added instructions for using fdfind on Debian-based distros.

### DIFF
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -34,8 +34,9 @@ You can run the tests with ``make test`` and the linters by running ``make
 lint``. There are further make targets defined, see the `Makefile` for details.
 After any changes to the Javascript code, you will need to re-build the
 frontend, which you can do by running ``make``. If you are working on the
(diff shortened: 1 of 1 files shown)


docs/development.rst
```diff
@@ -34,8 +34,9 @@ You can run the tests with ``make test`` and the linters by running ``make
 lint``. There are further make targets defined, see the `Makefile` for details.
 After any changes to the Javascript code, you will need to re-build the
 frontend, which you can do by running ``make``. If you are working on the
-frontend code, you can e.g. use `entr` to rerun the build on file changes: ``fd |
-entr -s make``.
+frontend code, you can e.g. use `entr` to rerun the build on file changes:
+``fd | entr -s make``, or on Debian based distros (Ubuntu, Mint, etc.)
```
yagebu: I don't really want to start an exhaustive list for all OSes - this is intended to be an example (entr and fd are both probably not available by default on most distros).

yagebu:
See #2227 which does this without any extra dependencies or CLIs that the user might not have installed.

PR #2226, opened 2026-02-15
Add duplicate button

mocsa:
When a Transaction, a Balance, or a Note is opened in the Context editor, a Duplicate button becomes visible. If clicked, it closes the Context editor and opens the AddEntry dialog pre-filled with the entry and the date set to today. I find it a huge quality of life improvement when manually entering reoccuring transactions.

Please note that I'm a typescript beginner (30 years in other languages) and it took me days(!) to put together this relatively simple change because of the extremely complicated matrix of interdependent types in different files. So if something is not according to your liking, please be gentle. This is the best way I could come up with to make the smallest possible change in the program. If you want me to correct something, please also give some hints on how to do it. Thanks.


frontend/src/api/index.ts
```diff
@@ -237,9 +240,48 @@ export const get_commodities = define_endpoint(
   commodities_validator,
   filters,
 );
+/**
+ * The JSON returned by the endpoint contains full details of
+ * the beancount directive with this hash in "entry" but
+ * the validator discards most details, except a few
```
yagebu: Since we want to use the full entries now, I think we can simply swap out `entryBaseValidator` for `entryValidator` in `context_validator`. Then we don't need these duplicate methods below (and we don't need to fetch the same entry twice.

frontend/src/editor/SliceEditor.svelte
```diff
@@ -70,6 +84,50 @@
...
+        default:
+          throw new Error(
+            `Duplication of entry type ${entry.t} not supported.`,
+          );
+      }
+
+      // Open the "Add Entry" dialog pre-filled with this entry's data.
+      // AddEntry will detect the type (Transaction/Note/Balance)
+      // and switch tabs.
+      const url = new SvelteURL(router.current);
```
yagebu:
```suggestion
      const url = new URL(router.current);
```

yagebu (changes requested):
Thanks for the PR, I like this feature. I left some comments.

> Please note that I'm a typescript beginner (30 years in other languages) and it took me days(!) to put together this relatively simple change because of the extremely complicated matrix of interdependent types in different files.

I'd say Fava doesn't do any really complicated Typescript tricks so that might just be Typescript taking some getting used to - but if you have some suggestions how to simplify things, feel free to share those.

frontend/src/editor/SliceEditor.svelte
mocsa: First, I used URL, but then the eslint validation forced me to use SvelteURL. The exact error it complained about was this: https://sveltejs.github.io/eslint-plugin-svelte/rules/prefer-svelte-reactivity/

frontend/src/api/index.ts
mocsa: I tried it a few minutes ago, but in SliceEditor.svelte around line 95 when I do `await get_context({ entry_hash })).entry.set` I again get the same error for `set()` that I fought with for hours before I submitted my PR:

```
This expression is not callable.
  Each member of the union type '(<T extends keyof Transaction>(key: T, value: Transaction[T]) => Transaction) | (<T extends keyof Balance>(key: T, value: Balance[T]) => Balance) | ... 8 more ... | (<T extends keyof Query>(key: T, value: Query[T]) => Query)' has signatures, but none of those signatures are compatible with each other.
```

This was one thing I found overly complex. Things start out logical when an EntryBase abstract class is defined in `entries/index.ts`. It has a `set()` method which every entry type will need, therefore all entry type objects inherit from EntryBase, which is all fine. But the problem is, `set()` is generic (I guess because it is not a real set method but a set+clone method), therefore it cannot be called on an arbitrary entry. First you have to determine the entry's type yourself and pass this info to the set method. This is why I created 3 specialized get_context methods. Again, this might be me thinking in Java, but I cannot figure out how to reduce 3 calls (for the 3 different entry types) to only one.

By the way, the reason why I put `await get_context`, `entry.set`, and `initial_entry.set` all in one command because I simply couldn't figure out how to type variables which will store the output of `await get_context_*` and `entry.set` for arbitrary entry types without giving all kind of errors.

mocsa:
Maybe it is not the best place to discuss it here but one thing I couldn't understand is why do we need the complex network of validators? Why do we have to validate data that is coming out of an API which is developed in this same github repository? Do we not trust that the Python API will send the data we programmed it for? So we validate in Javascript whether Python sent us the correct list of fields with the correct data types? Can it ever happen that Python makes a mistake?

frontend/src/api/index.ts
yagebu: Adding a `set("date"...` overload to the method allows one to call it on all the subclasses without error

yagebu:
> Maybe it is not the best place to discuss it here but one thing I couldn't understand is why do we need the complex network of validators? Why do we have to validate data that is coming out of an API which is developed in this same github repository? Do we not trust that the Python API will send the data we programmed it for? So we validate in Javascript whether Python sent us the correct list of fields with the correct data types? Can it ever happen that Python makes a mistake?

Neither Typescript nor Python do runtime type checking from annotations, so the validators take care of that (and check the types at the boundary). Without them, there's no automatism to validate that the Python and Typescript side match.

Overall I think I prefer adding the duplicated entry as part of the context modal. It avoid switching between the modals, is simple to implement and I find it helpful to see the original entry. I have implemented that in #2308, so I'm closing this PR, thank you for working on this feature :)

PR #2244, opened 2026-03-01
Added missing dependencies for docker build

eldur:
tried with
% docker build --no-cache -t fava .


hosiet:
Two issues:

* Please use `apt-get` instead of `apt` for non-interactive operation.
* Please clean up `apt` caches upon completion to reduce final image size.

It should be something like:

```
RUN apt-get update && apt-get install -y \
    package1 \
    package2 \
    && rm -rf /var/lib/apt/lists/*
```

yagebu:
Closing this PR due to inactivity, feel free to reopen / create a new one

PR #2254, opened 2026-03-07
Add chart tab selection to account report URL params

BenjaminDebeerst:
In my Fava setup, I have a couple of fixed links to certain account reports. It's nice to be able to pre-select in the link the report shown via the `r` URL param, but a similar param is so far missing to pre-select the chart shown. This change adds that capability.

Memoize the active chart index via `c` query parameter in URL, allowing to deep link the chart selection. The default chart 'Account Balance' at index 0 is omitted from the URL analogous to the default report selection.

The values for the selected chart are the chart index in order to provide both a language agnostic solution and allowing to deep link into e.g. the hierarchy chart for the last month (because a label would change depending on the date, while the index does not).

Most of the logic is part of the frontend/template code because it needs to run after the addition of the hierarchy charts, which happens only there.

This purely being frontend changes, I did not add or modify any tests. I tested manually with `long-example.beancount` using both the English and the German languages.

-- 

AI usage disclosure: Since I'm not a frontend dev and only occasionally write TS, Claude Sonnet 4.6 was used to explore the code, find the right places to place modifications and suggest code snippets - all edits where made manually though and carefully reviewed.

### DIFF
--- a/frontend/src/reports/accounts/AccountReport.svelte
+++ b/frontend/src/reports/accounts/AccountReport.svelte
@@ -1,10 +1,13 @@
 <script lang="ts">
   import ChartSwitcher from "../../charts/ChartSwitcher.svelte";
   import { ParsedHierarchyChart } from "../../charts/hierarchy.ts";
+  import type { ParsedFavaChart } from "../../charts/index.ts";
   import { urlForAccount } from "../../helpers.ts";
   import { _ } from "../../i18n.ts";
   import { is_non_empty } from "../../lib/array.ts";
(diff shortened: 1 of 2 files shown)


BenjaminDebeerst:
This partially reimplements #2191 in a simpler form - it tracks the selected chart but not the chart mode.

BenjaminDebeerst:
@yagebu is there interest in this change? It has gotten stale over time, but I'd willing to rebase/rework if there is interest in this.

PR #2262, opened 2026-03-15
add time selection by dragging to charts

jakobhellermann:
I've found myself trying to select a time range in the graphs multiple times, but it doesn't work.
With this PR it does:

https://github.com/user-attachments/assets/0e4f68d7-b66e-43dc-abac-b5062c2ac304

Works with mouse and touch.

This was written with the help of AI, but cleaned up and tested by hand, and the code makes sense to me.

### DIFF
--- a/frontend/src/charts/BarChart.svelte
+++ b/frontend/src/charts/BarChart.svelte
@@ -8,6 +8,7 @@
   import { ctx, currentTimeFilterDateFormat, short } from "../stores/format.ts";
   import Axis from "./Axis.svelte";
   import type { BarChart } from "./bar.ts";
+  import { dateRangeBrush } from "./brush.ts";
   import {
     currenciesScale,
     filterTicks,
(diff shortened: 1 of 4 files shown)


yagebu:
Thanks for the PR, this is a nice feature - however I've implemented this as a Svelte component, see #2306

frontend/src/charts/BarChart.svelte
```diff
@@ -84,10 +85,29 @@
   let yAxis = $derived(
     axisLeft(y).tickPadding(6).tickSize(-innerWidth).tickFormat($short),
   );
+
+  /** Invert a pixel x position to the date of the nearest bar group. */
+  function invertX(px: number): Date {
+    const half = x0.bandwidth() / 2;
+    let bestDate = new Date();
```
yagebu, line 92: This can be written much more concisely with d3.least (see my PR), that's one standard AI implementation gotcha to look out for

frontend/src/charts/brush.ts
```diff
@@ -0,0 +1,108 @@
...
+): Attachment<SVGGElement> {
+  return (node) => {
+    let startX = 0;
+    let activePointerId: number | null = null;
+    let rect: SVGRectElement | null = null;
+
+    // Prevent touch scrolling so the brush works on touch devices.
+    node.style.touchAction = "none";
+
+    function localX(event: PointerEvent): number {
```
yagebu, line 29: Also stumbled on this, due to the random computation with fields like `ctm.e` and `ctm.a`. Again, AI likes to reimplement things instead of just using existing functionality (it doesn't know about due to missing context). For this here, d3.pointer does the job

PR #2267, opened 2026-04-16
Add transaction by reference

mocsa:
This PR is preparatory work to enable the duplication of Transactions, Balances, and Notes.

This is a much simpler approach than what I followed in #2226. It takes inspiration from Context.svelte. It reads the entry_hash from the URL and loads it into the Add dialog. The URL should look like https://server/beancountfile/account/Assets:Bank:Checking#add-transaction-from-cf93b24126f19dadf05c5a1f35a8af93

I took your advice @yagebu in #2226 to replace EntryBaseValidator with EntryValidator to keep all data that is returned by get_context().

Currently, this functionality will not be used in fava, but I thought I submit it separately to make review easier. I plan to submit the related GUI modification after this PR is merged.


yagebu:
Closing, as this was intended for entry duplication, which was implemented in #2308

PR #2269, opened 2026-05-03
Navigation error query holdings

### LINKED ISSUE #2268: Navigation error "reading #page-title failed" when clicking "Query" from Holdings page
A navigation error occurs when using the "Query" hyperlink located on the Holdings page (under the sub-tabs). The page fails to render, displaying a blank content area and the error message `ERROR: reading #page-title failed.` in the heading.

A browser refresh or pasting the URL directly into a new tab renders the page correctly.

## Steps to Reproduce
- Navigate to the Holdings page [https://fava.pythonanywhere.com/example-beancount-file/holdings/](https://fava.pythonanywhere.com/example-beancount-file/holdings/)
- Click the "Query" hyperlink located below the report tabs (Holdings | Holdings by account | ...).

## Environment
- Fava Version: 1.30.12
- Beancount Version: 3.2.0
- Also on PythonAnywhere [https://fava.pythonanywhere.com/example-beancount-file/holdings/](https://fava.pythonanywhere.com/example-beancount-file/holdings/)

guiu-rocafort:
This merge request fixes #2268 by ensuring there is a trailing slash after the "query".
I have added a new test case for the issue. 
All tests are passing, linting does not show any errors.

### DIFF
--- a/frontend/test/urls.test.ts
+++ b/frontend/test/urls.test.ts
@@ -37,3 +37,13 @@ test("extract relative path from URL", () => {
     "Ä€/asdf",
   );
 });
+
+test("report name extraction", () => {
+  const getReport = (path: string) => path.slice(0, path.indexOf("/"));
+
+  equal(getReport("holdings/"), "holdings");
+  equal(getReport("query/"), "query");
+  // This demonstrates the current behavior where a missing trailing slash
+  // leads to an incorrect report name being extracted.
+  equal(getReport("query"), "quer");
+});
--- a/frontend/src/reports/holdings/Holdings.svelte
+++ b/frontend/src/reports/holdings/Holdings.svelte
@@ -53,7 +53,7 @@
 </div>
 
 <p>
-  <a href={$urlFor("query", { query_string })}>{_("Query")}</a>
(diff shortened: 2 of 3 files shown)


frontend/test/urls.test.ts
```diff
@@ -37,3 +37,13 @@ test("extract relative path from URL", () => {
     "Ä€/asdf",
   );
 });
+
+test("report name extraction", () => {
```
yagebu, line 41: What's the point of this test? This locally defines a function and tests it - so it doesn't test any actual part of the application

guiu-rocafort: It tests that there is a trailing slash after the "query" url, it's meant to be a regression test in case it breaks in the future...

yagebu: But it doesn't work as a test. This test is completely independent of everything else, so changing the urlFor parameters back to `query` wouldn't cause this test to break.

yagebu: I instead added types to validate it ends with a trailing slash, see my PR #2292

PR #2280, opened 2026-05-17
fix: use field-sizing: content for AutocompleteInput width on wide chars

### LINKED ISSUE #2250: Incorrect (insufficient) UI size calculation of <input> HTML element in AutocompleteInput.svelte
### The Problem

For the following code in git HEAD:

https://github.com/beancount/fava/blob/dc3ffb9ec496555693188b5fa907630869f6ed94/frontend/src/sidebar/FilterForm.svelte#L107-L119

Looks like its size is computed based on text size, after applying translation:

https://github.com/beancount/fava/blob/dc3ffb9ec496555693188b5fa907630869f6ed94/frontend/src/AutocompleteInput.svelte#L75-L77

...which works great in the default scenario:

Image

```python
-> % python3
Python 3.13.12 (main, Feb  4 2026, 15:06:39) [GCC 15.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> len('Filter by tag, payee, ...')+1
26
```

...but fails to handle CJK characters with wide characters:

Image

```python
-> % python3
Python 3.13.12 (main, Feb  4 2026, 15:06:39) [GCC 15.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> len('筛选（标签，收款人等等）')+1
13
```

This is expected since the `size` attribute of `<input>` element does not consider wide characters and is an approximation anyway. In general, wide characters like CJK characters takes approximately the width of 2 narrow characters.

* * *

### Possible solutions

I don't think native JS provides any elegant solution to this issue.

* We either identify wide characters in the string and manually assign 2x size value to them (ugly),
* or compute rendered result and explicitly set the width of `<input>` width.

I don't really like either of the solution and this bug is not that important, so I would like to hear from others first on how to handle it.

m4cd4r4:
## Fix autocomplete input width for CJK and wide characters

Closes #2250

### Root cause

`AutocompleteInput.svelte` sets the HTML `size` attribute from
`Math.max(value.length, placeholder.length) + 1`. The `length`
property counts UTF-16 code units, not display columns, so CJK and
other wide characters are counted as width 1 when they occupy width 2.
The result is an input that renders too narrow when the placeholder or
value contains wide characters.

### Fix

Added `field-sizing: content` to the `input` CSS rule in
`AutocompleteInput.svelte`. Browsers that support it (Chromium-based
and Safari) will auto-size the input to its rendered content, handling
wide characters correctly.

The existing length-based `size` attribute is kept unchanged as the
fallback. Firefox does not yet support `field-sizing`, so Firefox
behaviour is identical to before - no regression.

### Change

One line added to the `input` style block in
`frontend/src/AutocompleteInput.svelte`:

```css
input {
  field-sizing: content;  /* added */
  width: 100%;
}
```

### Checks

`make lint` passes: pre-commit (all files), `tsc`, `svelte-check`, and the `ty` Python type check are all clean.

`make test`: the frontend suite (`npm run test`) passes 86/86. The Python suite reports one snapshot mismatch in `tests/test_json_api.py::test_api[options-/long-example/api/options]`. That same test fails identically on an unmodified `master` checkout on this machine, so it is a pre-existing environment-specific snapshot drift, not a regression from this change. This PR only adds one CSS property in a Svelte file, which cannot affect the Python JSON API output.

### DIFF
--- a/frontend/src/AutocompleteInput.svelte
+++ b/frontend/src/AutocompleteInput.svelte
@@ -221,6 +221,7 @@
   }
 
   input {
+    field-sizing: content;
     width: 100%;
   }
 


yagebu:
Thanks for the PR.

I don't think this quite the right fix (as this applies to all autocomplete inputs, not just the sized ones). I'd also want to wait for Firefox to have this, so I'll close this PR.

PR #2282, opened 2026-05-17
Fix document metadata links on transaction postings

### LINKED ISSUE #1866: link_documents: support linking documents in posting metadata
According to [the beancount documentation](https://beancount.github.io/docs/beancount_language_syntax.html#metadata-1), is it ok to attach a document key to a posting of a transaction:
```
2024-01-09 * "Bank/9"
  Uitgaven:Verteer  3.75 EUR
    document: "/bean/docs/2024/Uitgaven/Verteer/2024-01-09_120.pdf"
  Vermogen:Prive-mutaties  4847.51 EUR
  Uitgaven:Verteer  17.00 EUR
  Activa:Liquide:ING -4868.26 EUR
```
However, I experience an issue with fava (using in docker, Beancount version 2.3.6 and Fava 1.27.3).

Given the example above, If I open the journal view, expand this transaction and mouse over to the document link, that link is missing the entry_hash=.... part of the url, resulting in an error when clicked.

If I move the document key to the second line of the transaction:
```
2024-01-09 * "Bank/9"
  document: "/bean/docs/2024/Uitgaven/Verteer/2024-01-09_120.pdf"
  Uitgaven:Verteer  3.75 EUR  
  Vermogen:Prive-mutaties  4847.51 EUR
  Uitgaven:Verteer  17.00 EUR
  Activa:Liquide:ING -4868.26 EUR
```
open the journal view, expand this transaction and mouse over to the document link, that link has the entry_hash=... part included in the url and clicking the link does indeed show me the document.

This seems like a bug to me. Is it me, or can anyone confirm?

cvaske:
When a document metadata key is attached to a posting rather than the transaction itself, the journal view was generating a broken link missing the entry_hash parameter. This was caused by _render_metadata being called without entry_hash for posting-level metadata in _journal_table.html. A second related bug in statement_path meant that even with a valid entry_hash, the backend only searched entry.meta and not posting-level metadata, so the document could never be resolved. Both are fixed, with a test covering the posting-level case.

Additionally, updating python test snapshots to ensure `make test` passes.

Fixes #1866


yagebu:
Thanks, I'll rebase this and also extend the link_documents plugin accordingly

PR #2283, opened 2026-05-17
Prevent frontend documents/ crash from multiple doc refs

### LINKED ISSUE #2281: Documents page crashes with `each_key_duplicate` on duplicate Document directives
## Summary 

A document that is present directly in the ledger as a `document` statement, and also located inside a directory specified via `option "documents" "directory-for-documents"`, will cause the Documents page to error out with:

```
Loading /beancount/documents/ failed with error:

https://svelte.dev/e/each_key_duplicate
```

This is easy to run into when importing a document into fava to support a balance assertion.

## Expected behavior

Fava should handle duplicate Document entries from beancount gracefully — either by deduplicating them before rendering, or by displaying a warning — rather than crashing the entire page.

## Steps to reproduce:

```bash
cat <<EOF > example.beancount
option "operating_currency" "USD"
option "documents" "documents"

2020-01-01 open Assets:Bank:Checking USD

2020-01-31 balance Assets:Bank:Checking 100.00 USD
2020-01-31 document Assets:Bank:Checking "documents/Assets/Bank/Checking/2020-01-31-statement.pdf"
EOF

mkdir -p documents/Assets/Bank/Checking/
touch documents/Assets/Bank/Checking/2020-01-31-statement.pdf
```

## Versions

```
%fava --version
fava, version 1.30.12
% bean-check --version
Beancount 3.2.0
```

cvaske:
If a document filename is encountered multiple times, from transaction metadata and/or being auto-found via an `options` directive, Svelte code would cause a UI error without displaying any data. Fixes #2281 

Change summary:

- `src/fava/json_api.py` — `get_documents()` now deduplicates by filename, preferring explicit directives over auto-discovered ones
- `frontend/src/reports/documents/Table.svelte` — Svelte `{#each}` key changed from `doc.filename` to `doc.entry_hash`
- `tests/test_json_api.py` — Two new tests: one for explicit + auto-discovered dedup, one for two explicit directives

### DIFF
--- a/frontend/src/reports/documents/Table.svelte
+++ b/frontend/src/reports/documents/Table.svelte
@@ -46,7 +46,7 @@
     </tr>
   </thead>
   <tbody>
-    {#each sorted_documents as doc (doc.filename)}
+    {#each sorted_documents as doc (doc.entry_hash)}
       <tr
         class:selected={selected === doc}
         draggable={true}
--- a/src/fava/json_api.py
+++ b/src/fava/json_api.py
@@ -608,16 +608,17 @@ def get_documents() -> Sequence[Document]:
     g.ledger.changed()
     # Deduplicate by filename: prefer explicit directives (lineno > 0) over
     # auto-discovered ones (lineno == 0).
-    seen: dict[str, Document] = {}
+    seen: dict[Path, Document] = {}
     for entry in g.filtered.entries:
         if not isinstance(entry, Document):
             continue
-        existing = seen.get(entry.filename)
+        key = Path(entry.filename)
+        existing = seen.get(key)
         if existing is None or (
             existing.meta.get("lineno", 0) == 0
             and entry.meta.get("lineno", 0) != 0
         ):
-            seen[entry.filename] = entry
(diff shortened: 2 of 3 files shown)


frontend/src/reports/documents/Table.svelte
```diff
@@ -46,7 +46,7 @@
     </tr>
   </thead>
   <tbody>
-    {#each sorted_documents as doc (doc.filename)}
+    {#each sorted_documents as doc (doc.entry_hash)}
```
yagebu, line 49: Does this work? Wouldn't duplicate documents also not have the same entry hash?

cvaske: This works for the test case in #2281, which is not surprising because the API change below deduplicates the file.

Moreover, if the API level changes below are removed, then the API returns the duplicate files, the entry_hashes are different, and the display continues to work, even though they have the same filename. I haven't looked deep enough into the beancount repo to find out what precisely gets hashed from the objects.

yagebu:
I went with just showing the duplicates in #2290

PR #2285, opened 2026-05-23
Support extensions triggering file import

mescanne:
This functionality enables extensions to trigger a normal file import process. (for reference - https://github.com/mescanne/beancount-blue)

Background:
I'm building an extension to enable API-driven sync and imports. I'm using the standard Beancount and Fava file import pattern (where files contain configuration and do not move) and would like to leverage Fava's normal import process, but not import screen. Another PR will be raised to cover bulk importing.

Implementation:
* Adds a globally hoisted `<GlobalExtract>` modal that triggers on the URL hash pattern `#extract?filename=...&importer=...`.
* The loading state is delegated to Fava's global `loading_state.await()`, keeping the PR focused and inheriting generic UI loading enhancements natively.

Let me know of any questions or feedback!

Thanks

### DIFF
--- a/frontend/src/modals/GlobalExtract.svelte
+++ b/frontend/src/modals/GlobalExtract.svelte
@@ -1,0 +1,73 @@
+<script lang="ts">
+  import { get_extract, save_entries } from "../api/index.ts";
+  import type { Entry } from "../entries/index.ts";
+  import { is_non_empty } from "../lib/array.ts";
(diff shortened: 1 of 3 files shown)


yagebu:
I don't really want to add a modal that is not part of Fava's functionality that is implicitly triggered.

I've improved the import page to keep the modal state in the URL parameters (#2312), so you could link to that now, would that address your use case?

PR #2287, opened 2026-05-27
Improve balance amount input on mobile

mekanics:
## Summary
- Change the balance amount field from a telephone input to a text input with `inputmode="decimal"` so iOS shows a decimal-friendly keyboard.
- Add a frontend regression test covering the amount input type, input mode, and preserved pattern.

### DIFF
--- a/frontend/src/entry-forms/Balance.svelte
+++ b/frontend/src/entry-forms/Balance.svelte
@@ -38,7 +38,8 @@
     --autocomplete-wrapper-flex="1"
   />
   <input
-    type="tel"
+    type="text"
+    inputmode="decimal"
     pattern="-?[0-9.,]*"
     placeholder={_("Number")}
     size={10}
--- a/frontend/test/entry-forms.test.ts
+++ b/frontend/test/entry-forms.test.ts
@@ -1,0 +1,28 @@
+import { equal, ok } from "node:assert/strict";
+import { beforeEach, test } from "node:test";
+
+import { flushSync, mount } from "svelte";
+
+import { Balance } from "../src/entries/index.ts";
+import BalanceForm from "../src/entry-forms/Balance.svelte";
+import { setup_jsdom } from "./dom.ts";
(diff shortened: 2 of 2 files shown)


frontend/src/entry-forms/Balance.svelte
```diff
@@ -38,7 +38,8 @@
     --autocomplete-wrapper-flex="1"
   />
   <input
-    type="tel"
+    type="text"
+    inputmode="decimal"
```
yagebu, line 42: According to MDN, this would show the decimal separator for the users locale - does that always allow entering a `.` as decimal separator as well?

mekanics: good catch. and actually the decimal input doesn't support the `-` either, so as I see it we could:

a) keep inputmode="decimal" for the good keypad and just normalize the separator (a typed locale `,` → `.`) before saving, since beancount/Decimal() only accept `.`. negatives will only work on desktop.
b) add a +/− toggle button next to the field. works on every keyboard, but doesn't really fit into the design. and we still need to normalize the value (option a).
c) change to type="number", which shows the full keyboard with `-`, `.` and all the special chars. downside is we lose the nice compact numeric keypad on mobile.


Here is a small example of the different inputs:
https://htmlpreview.github.io/?https://gist.githubusercontent.com/mekanics/30201fc5c7fb57a9aacb1a4b755aadeb/raw/2a470c4d4fc08ee345cba5ace2ad790ffb4168d3/index.html

I'm leaning towards c), but curious what you prefer.

yagebu: See also https://css-tricks.com/everything-you-ever-wanted-to-know-about-inputmode/

Just doesn't seem that iOS has a numeric keyboard that works for us, I think I'll change it to use inputmode numeric

yagebu: Implemented in 612e576f8073176d7078ecea8388be220787c5d8, so I'm closing this PR, thanks for raising this :)

PR #2297, opened 2026-06-16
fix: repair test suite against rustledger v0.16.1

robcohen:
Pins rustledger **v0.16.1** and repairs the rustfava test suite (was red on `main`).

## Why v0.16.1

v0.16.0's FFI load surface skipped the pre-booking synth pass, so `auto_accounts` never generated `Open` directives through the embedded engine. [rustledger#1404](https://github.com/rustledger/rustledger/pull/1404) fixed this (shipped in v0.16.1): the FFI now routes loading through the canonical `process::load` pipeline, returning auto-generated Opens with the generated marker (`lineno 0`).

## Changes

**Engine-coupled (correct against v0.16.1):**
- **loader**: sort entries by `(date, sort-order, lineno)` — fava assumes date order, the engine returns source order; drop the spurious `plugins=["auto_accounts"]` arg (`auto_accounts` is a synth plugin the engine runs from the ledger declaration, not a requestable regular plugin — it produced an `Unknown plugin` error).
- **test_json_api**: find generated entries by the real marker (filename `<…>` / falsy `lineno`, what `_get_position` keys on) instead of the literal `<auto_accounts>`.
- **snapshots**: regenerated for the corrected engine output — generated Opens in the journal, resolved cost basis on positions, the removed `Unknown plugin` error, and a more complete booking-error set (each diff verified, all *more correct*).

**Engine-independent:**
- **serialisation**: guard `deserialise_posting` against amounts the rustledger parser silently truncates (e.g. `"10 ////"`) — when the posting has no units/cost/price the amount must be a bare number, else raise `InvalidAmountError`.
- **justfile**: `test-js` runs `bun run test` (node `--conditions browser`, matching CI) — plain `bun test` resolves svelte to its server build, breaking `mount()` under jsdom.

## Testing

Against the **published** v0.16.1 wasm (sha-verified): Python **480 passed, 100% coverage**; JS **86 passed**.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


robcohen:
Opened against the wrong repository by mistake (intended rustledger/rustfava, a fork). Closing.

PR #2298, opened 2026-06-19
Fix minor translation error on account in zh_Hant_TW

chuanghs:
Current translation on zh_Hant_TW has minor error. The characters used for account is misused zh version instead of zh_Hant_TW specific conversion.


yagebu:
The translations are managed on Weblate, please fix it there: https://hosted.weblate.org/projects/fava/fava/

PR #2299, opened 2026-06-19
Adaptive price base

upsuper:
When there is a commodity conversion, in the price column it always shows the per unit price of the target commodity, e.g. if I sell 1 VOO for 695 USD, the price would show 695 USD. This is usually okay for shares where the counterpart commodity is the denominated currency, so the price is usually quite recognizable. However, it may not work well when both sides are fiat currencies and they have very different scale of value. As an example, when converting from USD to JPY or AUD to JPY, the price just shows as `0.01 USD` or `0.01 AUD`, which is not very useful to know the actual exchange rate involved.

This PR changes how price is displayed. It chooses an adaptive price base according to the difference between the pair of commodities. For example, in the case of USD to JPY and AUD to JPY, it would show something like `100 JPY = 0.66 USD` and `100 JPY = 0.91 AUD`, which is much more useful than above.

Disclosure: the implementation is mostly done by AI, but I reviewed the code and think it is a reasonable implementation.

### DIFF
--- a/src/fava/template_filters.py
+++ b/src/fava/template_filters.py
@@ -51,6 +51,12 @@ def format_currency(value: Decimal, currency: str | None = None) -> str:
     return g.ledger.format_decimal(value or ZERO, currency)
 
 
+def price_base(value: Decimal) -> Decimal:
+    """Return a scaled base amount for displaying a price."""
+    exponent = max(0, round(-value.log10())) if value > ZERO else 0
+    return Decimal(10) ** exponent
+
+
 FLAGS_TO_TYPES = {"*": "cleared", "!": "pending"}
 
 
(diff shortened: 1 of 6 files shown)


src/fava/template_filters.py
```diff
@@ -51,6 +51,12 @@ def format_currency(value: Decimal, currency: str | None = None) -> str:
     return g.ledger.format_decimal(value or ZERO, currency)
 
 
+def price_base(value: Decimal) -> Decimal:
+    """Return a scaled base amount for displaying a price."""
+    exponent = max(0, round(-value.log10())) if value > ZERO else 0
```
upsuper, line 56: Using `round` here means that the result would be consistent if the number is floating around power of 10, e.g. we can have both `1 EUR = 1.18 USD` and `1 EUR = 0.97 USD`, rather than `10 EUR = 9.70 USD`. The latter is what it would be if it's `floor` here. But in fact it just pushes the boundary towards $\sqrt{10} \approx 3.16$, so if the value floats around that, the base can still switch back and forth.

PR #2303, opened 2026-06-24
fix print viewport

stenius:
Firefox does not fragment a CSS grid container across printed pages reliably, which clipped long reports to a single viewport and stranded the account chart/journal on page 2 with whitespace above it. Block flow releases the `100vh` body height and lays out header -> chart -> journal naturally, paginating row by row. (aside is hidden in print.)

### DIFF
--- a/frontend/css/layout.css
+++ b/frontend/css/layout.css
@@ -85,7 +85,33 @@ article:has(> .fixed-fullsize-container) {
 }
 
 @media print {
+  /*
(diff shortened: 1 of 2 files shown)


yagebu:
Thanks, I'll move the journal adjustment to the journal css file and merge this.

yagebu:
Merged with adjustment in 6c739923c822198cf0ae4c325e342d17dfa025f3

stenius:
thanks @yagebu !

PR #2313, opened 2026-07-17
fix: export empty set columns

### LINKED ISSUE #1951: Exporting report with tags to Excel fails
Attempting to export a report containing a frozenset column (so tags at least) to Excel causes a 500 error.

```
Exception on /beancount/download-query/query_result.xlsx [GET]
  File "/home/hambone/.local/share/uv/tools/fava/lib/python3.13/site-packages/openpyxl/worksheet/_write_only.py", line 133, in _values_to_row
    cell.value = value
    ^^^^^^^^^^
  File "/home/hambone/.local/share/uv/tools/fava/lib/python3.13/site-packages/openpyxl/cell/cell.py", line 218, in value
    self._bind_value(value)
    ~~~~~~~~~~~~~~~~^^^^^^^
  File "/home/hambone/.local/share/uv/tools/fava/lib/python3.13/site-packages/openpyxl/cell/cell.py", line 187, in _bind_value
    raise ValueError("Cannot convert {0!r} to Excel".format(value))
ValueError: Cannot convert frozenset() to Excel
```

Looks like this might be straightforwardish to fix. I'll look into later.

mgalore:
## Summary

- serialize set-valued query columns before handling empty values
- export empty tag sets as blank cells instead of `frozenset()`
- add CSV and XLSX regression coverage for empty and populated tag sets

Fixes #1951

## Verification

- `pytest -q tests/test_util_excel.py` — 3 passed
- `PATH="$PWD/.venv/bin:$PATH" pytest -q` — 470 passed
- `ruff check src/fava/util/excel.py tests/test_util_excel.py`
- `ruff format --check src/fava/util/excel.py tests/test_util_excel.py`
- `git diff --check`

### DIFF
--- a/src/fava/util/excel.py
+++ b/src/fava/util/excel.py
@@ -91,16 +91,15 @@ def _row_to_pyexcel(row: ResultRow, header: list[Column]) -> list[str | float]:
     result: list[str | float] = []
     for idx, column in enumerate(header):
         value = row[idx]
-        if not value:
-            result.append(value)
-            continue
         type_ = column.datatype
-        if type_ is Decimal:
+        if type_ is set:
+            result.append(" ".join(value))
+        elif not value:
+            result.append(value)
+        elif type_ is Decimal:
             result.append(float(value))
         elif type_ is int:
             result.append(value)
-        elif type_ is set:
-            result.append(" ".join(value))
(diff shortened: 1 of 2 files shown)


yagebu (approved):
Thanks for this fix :)

Issue #2148, opened 2025-12-02
Feature Request: Average Value View for Income Statement (per day/week/month)


richtefee:
### Context
In the Income Statement, Fava lets you change chart types, conversions, and filter by time, account, or tag. What I’m missing is a way to show average values per time unit (day/week/month) for the filtered period.

### Example:
For a trip from Oct 10–26, all tagged 2025_Holidays, I’d like to see the average cost per day of that trip.

### Idea:
Add an “Average” dropdown (similar to the “Conversion” menu) that converts all displayed numbers into averages based on a chosen unit: Daily, Weekly, Monthly, etc. The already known sum and  first and last booking slots can be used to calculate this quantities.

### Commitment:
I’m new to contributing to Fava but would like to try. A pointer to where this would fit in the code (maybe around fava.core.tree.SerialisedTreeNode?) would be very helpful.

PR #2314, opened 2026-07-19
income statement: add interval averages to report values

vahidmah:
Implement the average view requested in #2148 by adding an `average` query parameter (day/week/month/quarter/year) and applying the computed divisor across income statement trees and charts.

Add frontend support for selecting and syncing the average interval in the URL, include per-currency bar chart averages in the API payload, and adjust number formatting to preserve useful precision for averaged values.

Update API/frontend tests and snapshots for the new behavior.

Fix issue, #2148


vahidmah:
@yagebu, please let me know if you have any concerns with this implementation.

PR #2320, opened 2026-07-26
Allow emoji/Unicode characters in filter search strings

### LINKED ISSUE #2302: Emoji characters are not searchable in Fava
I use a lot of emoji in my beancount to log the transcation, but I find it doesn't work when I want to filter my wallet with emoji, here is the example, when I search for ⛽️

http://127.0.0.1:8888/wallet/api/account_report?filter=%E2%9B%BD%EF%B8%8F&time=2025&a=Expenses%3ATransport&r=journal

```
{"error":"Illegal character \"\u26fd\" in filter.\u26fd\ufe0f"}
```

## Steps to Reproduce
1. Add a transaction containing an emoji, for example:
   ```beancount
   2026-06-24 * "Cafe ☕" "Coffee"
     Expenses:Food:Coffee   5.00 AUD
     Assets:Bank

2. Open the ledger in Fava.
3. Use the search box to search for ☕.

## Expected Behavior
Fava should return transactions containing the searched emoji character.

## Actual Behavior
Error show up `HTTP 400 - Illegal character "☕" in filter.☕`

## Notes
Searching for regular text in the same transaction works as expected. The issue seems specific to emoji / Unicode symbol search terms.

## Environment
- Fava version: `1.30.14`
- Beancount version: `3.2.3`
- Browser: `Google Chrome 149.0.7827.196`
- OS: `macOS 26.5.1`

pike00:
The `STRING` token regex in `FilterSyntaxLexer` only matched ASCII word characters (`\w`), so emoji like ☕ fell through to the illegal-character error handler.

Added a catch-all pattern `[^\s\-(),#^:="\']+` as the last alternative in the STRING regex, so emoji and other Unicode characters are now valid search terms. This only matches after all other token patterns (TAG, LINK, KEY, quoted strings, numbers, operators) have been tried.

Fixes #2302

### DIFF
--- a/src/fava/core/filters.py
+++ b/src/fava/core/filters.py
@@ -104,7 +104,7 @@ class FilterSyntaxLexer:
         ("EQ_OP", r":"),
         ("CMP_OP", r"(=|>=|<=|<|>)"),
         ("NUMBER", r"\d*\.?\d+"),
-        ("STRING", r"""\w[-\w]*|"[^"]*"|'[^']*'"""),
+        ("STRING", r"""\w[-\w]*|"[^"]*"|'[^']*'|[^\s\-(),#^:="']+"""),
     )
 
     regex = re.compile(
--- a/tests/test_core_filters.py
+++ b/tests/test_core_filters.py
@@ -87,6 +87,24 @@ def test_lexer_emoji() -> None:
     ]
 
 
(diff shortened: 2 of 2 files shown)


yagebu (approved):
Thanks :)

PR #2327, opened 2026-07-30
Fix quarter start boundary dates

### LINKED ISSUE #2318: _IntervalQuarter.get_prev off by one: first month of each quarter maps to the previous quarter
`_IntervalQuarter.get_prev` uses `>` where it needs `>=`, so a date in the first month of a quarter is assigned to the previous quarter.

```python
def get_prev(self, date):
    for i in [10, 7, 4]:
        if date.month > i:
            return datetime.date(date.year, i, 1)
    return datetime.date(date.year, 1, 1)
```

April is not `> 4`, so it falls through to January 1. Same for July and October.

## Repro

fava 1.30.14:

```python
import datetime
from fava.util.date import _IntervalQuarter

q = _IntervalQuarter()
for m in (1, 4, 7, 10):
    d = datetime.date(2023, m, 15)
    print(d, q.format_date(d), q.get_prev(d), q.number_of_days(d))
```

```
2023-01-15 2023-Q1 2023-01-01 90
2023-04-15 2023-Q2 2023-01-01 90   <- should be 2023-04-01, 91
2023-07-15 2023-Q3 2023-04-01 91   <- should be 2023-07-01, 92
2023-10-15 2023-Q4 2023-07-01 92   <- should be 2023-10-01, 92
```

`format_date` computes the quarter correctly with `(date.month - 1) // 3 + 1`, so the label and the period disagree inside the same class: a date labelled Q2 gets Q1's start and length.

## Effect on budgets

`calculate_budget` divides by `number_of_days`, so a quarterly budget accrues at the wrong daily rate for most of the year. April to June divides by 90 instead of 91, July to September by 91 instead of 92.

Ledger:

```beancount
2019-01-01 open Expenses:X
2020-02-28 custom "budget" Expenses:X "quarterly" 4683.56 USD
```

```python
calculate_budget(budgets, "Expenses:X", date(2022, 10, 19), date(2025, 3, 4))
# {'USD': Decimal('44513.98439539204756596060974')}
```

Summing by hand over calendar quarters gives 44462.14374879227053140096618, a difference of about 52. Working:

```
4683.56 * (74/92)   Q4 2022, from Oct 19
4683.56 * 8         Q1 2023 through Q4 2024, full quarters
4683.56 * (62/90)   Q1 2025, to Mar 3
```

## Fix

```python
if date.month >= i:
```

Anything else using `Interval.QUARTERLY` for grouping or labelling is affected too, not just budgets.

Found while building a differential test for a quarterly budget report against fava. The other four intervals (daily, weekly, monthly, yearly) match to full precision.

bm1016bm-svg:
## Summary

- treat April, July, and October as the starts of their current quarters
- cover exact quarter boundaries in both interval and day-count tests

Fixes #2318.

## Testing

- `uv run --no-dev --group test pytest tests/test_util_date.py` — 151 passed
- `ruff check src/fava/util/date.py tests/test_util_date.py` — passed

## AI assistance

Implementation and tests were developed with OpenAI Codex assistance, then reviewed and validated locally.


yagebu:
Thanks for creating the PR, I decided to do this a bit differently and get rid of the loop altogether, see #2328

Issue #2198, opened 2026-01-24
Show error in chinese


crazyzhc98:
The transaction dialog shows error in chinese word which shows red background since 1.30.9.
Such as account is 'Expenses:Service:生活服务' or payee is '水果店'

yagebu:
The tree-sitter grammar used in the editor follows the account regex specified by beancount, by which these are not valid account names (they would need to start with an uppercase letter which isn't possible for uncased scripts...). There's various tickets in the Beancount issue tracker on this, e.g. https://github.com/beancount/beancount/issues/398

The current Parser of Beancount is not so strict, so they might work in many places. I've adjusted the tree-sitter grammar to also allow uncased characters in https://github.com/yagebu/tree-sitter-beancount/pull/6, once we pull that in in Fava, those shouldn't be shown in red anymore.

yagebu:
Fixed in #2352

PR #2352, opened 2026-09-05
update tree-sitter grammar

yagebu:
This includes the change to diverge from Beancounts specified regex and
allow uncased characters where currently on uppercase ones are allowed
to better support e.g. Chinese characters, which have no case.


yagebu:
Fixes #2198

PR #2360, opened 2026-09-17
Switch to lightningcss-wasm

folliehiyuki:
`lightningcss` NPM package only has pre-built binaries for x86_64 and aarch64. By switching to `lightningcss-wasm`, we can ship `fava` to other platforms.


folliehiyuki:
We can also look into switching from `esbuild` to `esbuild-wasm`, but the currently in-use `esbuild` package already ships pre-built binaries for most common platform targets.

yagebu:
On what kind of machine are you developing that this caused an issue?

folliehiyuki:
We're trying to package the latest version of `fava` for Alpine Linux. The distribution is available for a lot of platforms, so we want to build and ship `fava` on all of them if possible. You can see the ongoing effort at https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/101073.

